PR #10, opened 2016-01-22
Modify Churn to use fewer partitions

mreid-moz:
Instead of extending `SimpleDerivedStream`, directly extend `DerivedStream`, and ignore the supplied list of S3 objects, instead fetching the file list for each day in the specified range.

Also repartition into 100 pieces to avoid huge numbers of tiny files on S3 which causes bad performance.

There are a few ugly hacks in here (including the hard-coding of `100` as the number of partitions). Feedback welcome.

### DIFF
--- a/src/main/scala/DerivedStream.scala
+++ b/src/main/scala/DerivedStream.scala
@@ -24,19 +24,19 @@ case class ObjectSummary(key: String, size: Long) // S3ObjectSummary can't be se
 abstract class DerivedStream extends java.io.Serializable{
   private val appConf = ConfigFactory.load()
   private val parquetBucket = Bucket(appConf.getString("app.parquetBucket"))
-  private val metadataBucket = Bucket("net-mozaws-prod-us-west-2-pipeline-metadata")
-  protected val metadataSources = {
-    val Some(sourcesObj) = metadataBucket.get(s"sources.json")
+  private val metaBucket = Bucket("net-mozaws-prod-us-west-2-pipeline-metadata")
+  protected val metaSources = {
+    val Some(sourcesObj) = metaBucket.get(s"sources.json")
     parse(Source.fromInputStream(sourcesObj.getObjectContent()).getLines().mkString("\n"))
   }
   private val metaPrefix = {
-    val JString(metaPrefix) = metadataSources \\ streamName \\ "metadata_prefix"
+    val JString(metaPrefix) = metaSources \\ streamName \\ "metadata_prefix"
     metaPrefix
   }
 
   protected val clsName = DerivedStream.uncamelize(this.getClass.getSimpleName.replace("$", ""))  // Use classname as stream prefix on S3
   protected val partitioning = {
-    val Some(schemaObj) = metadataBucket.get(s"$metaPrefix/schema.json")
+    val Some(schemaObj) = metaBucket.get(s"$metaPrefix/schema.json")
     val schema = Source.fromInputStream(schemaObj.getObjectContent()).getLines().mkString("\n")
     Partitioning(schema)
   }
--- a/src/main/scala/DerivedStream.scala
+++ b/src/main/scala/DerivedStream.scala
@@ -121,7 +121,7 @@ object DerivedStream {
     val sc = new SparkContext(conf)
     println("Spark parallelism level: " + sc.defaultParallelism)
 
-    val summaries = sc.parallelize(0 until daysCount + 1)
+    val summaries = sc.parallelize(0 to daysCount)
       .map(fromDate.plusDays(_).toString("yyyyMMdd"))
       .flatMap(date => {
                  s3.objectSummaries(bucket, s"$prefix/$date/$filterPrefix")
--- a/src/main/scala/DerivedStream.scala
+++ b/src/main/scala/DerivedStream.scala
@@ -18,16 +18,19 @@ import scala.collection.JavaConverters._
 import scala.io.Source
 import telemetry.streams.{E10sExperiment, ExecutiveStream, Churn}
 
+// key is the S3 filename, size is the object size in bytes.
 case class ObjectSummary(key: String, size: Long) // S3ObjectSummary can't be serialized
 
 abstract class DerivedStream extends java.io.Serializable{
   private val appConf = ConfigFactory.load()
   private val parquetBucket = Bucket(appConf.getString("app.parquetBucket"))
   private val metadataBucket = Bucket("net-mozaws-prod-us-west-2-pipeline-metadata")
-  private val metaPrefix = {
+  protected val metadataSources = {
     val Some(sourcesObj) = metadataBucket.get(s"sources.json")
-    val sources = parse(Source.fromInputStream(sourcesObj.getObjectContent()).getLines().mkString("\n"))
-    val JString(metaPrefix) = sources \\ streamName \\ "metadata_prefix"
+    parse(Source.fromInputStream(sourcesObj.getObjectContent()).getLines().mkString("\n"))
+  }
+  private val metaPrefix = {
+    val JString(metaPrefix) = metadataSources \\ streamName \\ "metadata_prefix"
     metaPrefix
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/DerivedStream.scala
```diff
@@ -18,16 +18,19 @@ import scala.collection.JavaConverters._
...
 
+// key is the S3 filename, size is the object size in bytes.
 case class ObjectSummary(key: String, size: Long) // S3ObjectSummary can't be serialized
 
 abstract class DerivedStream extends java.io.Serializable{
   private val appConf = ConfigFactory.load()
   private val parquetBucket = Bucket(appConf.getString("app.parquetBucket"))
   private val metadataBucket = Bucket("net-mozaws-prod-us-west-2-pipeline-metadata")
-  private val metaPrefix = {
+  protected val metadataSources = {
```
vitillo: Could you rename this to metaSources (to be consistent with the other variable named metaPrefix)?

src/main/scala/streams/Churn.scala
```diff
@@ -2,32 +2,124 @@ package telemetry.streams
...
 import org.apache.avro.generic.{GenericRecord, GenericRecordBuilder}
 import org.json4s.jackson.JsonMethods._
 import scala.collection.JavaConverters._
-import telemetry.{SimpleDerivedStream, Partitioning}
+import telemetry.{DerivedStream, ObjectSummary}
+import telemetry.DerivedStream.s3
 import telemetry.heka.{HekaFrame, Message}
-import org.json4s.JsonAST.{JValue, JNothing, JInt, JObject}
+import org.json4s.JsonAST.{JValue, JNothing, JInt, JObject, JString}
+import telemetry.parquet.ParquetFile
```
vitillo: Please order includes alphabetically (your editor should likely have a command for it).

src/main/scala/streams/Churn.scala
```diff
@@ -2,32 +2,124 @@ package telemetry.streams
...
+    val histograms = parse(fields.getOrElse("payload.histograms", "{}").asInstanceOf[String])
+    
+    val weaveConfigured = booleanHistogramToBoolean(histograms \ "WEAVE_CONFIGURED")
+    val weaveDesktop = enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_DESKTOP")
+    val weaveMobile = enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
+    
+    val map = Map[String, Any](
+        "clientId" -> (fields.getOrElse("clientId", None) match {
+          case x: String => x
+          case _ => None
```
vitillo: `clientId` can't be `None` as you haven't made it nullable in the schema definition.

src/main/scala/streams/Churn.scala
```diff
@@ -2,32 +2,124 @@ package telemetry.streams
...
+    // TODO: don't compute any of the expensive stuff if required fields are missing.
+    val profile = parse(fields.getOrElse("environment.profile", "{}").asInstanceOf[String])
+    val histograms = parse(fields.getOrElse("payload.histograms", "{}").asInstanceOf[String])
+    
+    val weaveConfigured = booleanHistogramToBoolean(histograms \ "WEAVE_CONFIGURED")
+    val weaveDesktop = enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_DESKTOP")
+    val weaveMobile = enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
+    
+    val map = Map[String, Any](
+        "clientId" -> (fields.getOrElse("clientId", None) match {
```
vitillo: No need to use `getOrElse`, please extract values in the case statement

src/main/scala/streams/Churn.scala
```diff
@@ -2,32 +2,124 @@ package telemetry.streams
...
+        "version" -> (fields.getOrElse("appVersion", None) match {
+          case x: String => x
+          case _ => ""
+        }),
+        "submissionDate" -> (fields.getOrElse("submissionDate", None) match {
+          case x: String => x
+          case _ => None
+        }),
+        "profileCreationDate" -> ((profile \ "creationDate") match {
+          case JNothing => null
```
vitillo: remove case for `JNothing`

src/main/scala/streams/Churn.scala
```diff
@@ -95,67 +187,36 @@ case class Churn(prefix: String) extends SimpleDerivedStream{
...
-      })
-      .set("syncConfigured", weaveConfigured.getOrElse(null))
-      .set("syncCountDesktop", weaveDesktop.getOrElse(null))
-      .set("syncCountMobile", weaveMobile.getOrElse(null))
-      .set("timestamp", message.timestamp)
-      .build
-    Some(root)
+    try {
+      // Required fields, raise an exception if they're not present.
+      for (field <- Array("clientId", "sampleId", "submissionDate", "timestamp")) {
```
vitillo: You should filter out submissions with missing fields before you repartition the data.

src/main/scala/streams/Churn.scala
```diff
@@ -95,67 +187,36 @@ case class Churn(prefix: String) extends SimpleDerivedStream{
...
+        root.set(field, (fields(field) match {
+          case None => throw new IllegalArgumentException()
+          case x => x
+        }))
+      }
+      
+      // String fields that are not nullable. Default to "".
+      for (field <- Array("channel", "normalizedChannel", "country", "version")) {
```
vitillo: This work can all be done in `messageToMap`.

src/main/scala/DerivedStream.scala
mreid-moz: Should rename `metadataBucket` too?

vitillo: Yes please.

src/main/scala/streams/Churn.scala
mreid-moz: Would it be reasonable to make a mutable map and add these fields incrementally, aborting if required fields are missing?

vitillo: What about returning `None` in case of failure? The return type of the method would change to `Option[Map[String, Any]]`.

src/main/scala/streams/Churn.scala
```diff
@@ -1,33 +1,127 @@
...
+      val summaries = sc.parallelize(s3.objectSummaries(bucket, s"$prefix/$currentDay/$filterPrefix")
+                        .map(summary => ObjectSummary(summary.getKey(), summary.getSize())))
+
+      val groups = DerivedStream.groupBySize(summaries.collect().toIterator)
+      val churnMessages = sc.parallelize(groups, groups.size)
+        .flatMap(x => x)
+        .flatMap{ case obj =>
+          val hekaFile = bucket.getObject(obj.key).getOrElse(throw new Exception("File missing on S3: " + obj.key))
+          for (message <- HekaFrame.parse(hekaFile.getObjectContent(), hekaFile.getKey()))  yield message }
+        .flatMap{ case message => messageToMap(message) }
```
mreid-moz: Is this `flatMap` all that's required to skip `None` values returned by `messageToMap`?

vitillo: Yes

src/main/scala/streams/Churn.scala
```diff
@@ -1,33 +1,127 @@
...
+import org.json4s.jackson.JsonMethods.parse
+import org.json4s.jvalue2monadic
+import org.json4s.string2JsonInput
+
+import awscala.s3.Bucket
+import telemetry.{DerivedStream, ObjectSummary}
+import telemetry.DerivedStream.s3
 import telemetry.heka.{HekaFrame, Message}
-import org.json4s.JsonAST.{JValue, JNothing, JInt, JObject}
+import telemetry.parquet.ParquetFile
```
vitillo: Could you sort those alphabetically?

vitillo:
r+

PR #11, opened 2016-01-26
Lazy-init the expensive payload transforms

mreid-moz:
Save the computation for records where we won't need it.

I'm not totally sure what the performance tradeoffs are for the `lazy` keyword, but this seems like a simple optimization for skipping computation on records without clientids.


vitillo:
r+

PR #14, opened 2016-01-27
Include version in assembly jar

mreid-moz:
This changes the assembly jar filename from `telemetry-batch-view.jar` to `telemetry-batch-view-1.0.jar` (and will change when we bump the version in `build.sbt`).


vitillo:
r+

PR #15, opened 2016-02-02
Add longitudinal view.

vitillo:
This patch adds a longitudinal view for release Telemetry data. See
https://bugzilla.mozilla.org/show_bug.cgi?id=1242039.

### DIFF
--- a/src/main/scala/histograms/Histogram.scala
+++ b/src/main/scala/histograms/Histogram.scala
@@ -1,0 +1,148 @@
+package telemetry.histograms
+
+import org.json4s._
+import org.json4s.jackson.JsonMethods._
+import scala.collection.mutable.{Map => MMap}
+import scala.io.Source
+
+case class RawHistogram(values: Map[String, Long], sum: Long)
+
+sealed abstract class HistogramDefinition
+case class FlagHistogram(keyed: Boolean) extends HistogramDefinition
+case class BooleanHistogram(keyed: Boolean) extends HistogramDefinition
+case class CountHistogram(keyed: Boolean) extends HistogramDefinition
+case class EnumeratedHistogram(keyed: Boolean, nValues: Int) extends HistogramDefinition
+case class LinearHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
+case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
+
+object Histograms {
+  val definitions = {
+    implicit val formats = DefaultFormats
+
+    val uris = Map("release" -> "https://hg.mozilla.org/releases/mozilla-release/raw-file/tip/toolkit/components/telemetry/Histograms.json",
+                   "beta" -> "https://hg.mozilla.org/releases/mozilla-beta/raw-file/tip/toolkit/components/telemetry/Histograms.json",
+                   "aurora" -> "https://hg.mozilla.org/releases/mozilla-aurora/raw-file/tip/toolkit/components/telemetry/Histograms.json",
+                   "nightly" -> "https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/telemetry/Histograms.json")
+
+    val parsed = uris.map{ case (key, value) =>
+      val json = parse(Source.fromURL(value, "UTF8").mkString)
+      val result = MMap[String, MMap[String, Option[Any]]]()
+
+      /* Unfortunately the histogram definition file does not respect a proper schema and
+         as such it's rather unpleasant to parse it in a statically typed langauge... */
+
+      for {
+        JObject(root) <- json
+        JField(name, JObject(histogram)) <- root
+        JField(k, v) <- histogram
+      } yield {
+        val value = try {
+          (k, v) match {
+            case ("low", JString(v)) => Some(v.toInt)
+            case ("low", JInt(v)) => Some(v.toInt)
+            case ("high", JString(v)) => Some(v.toInt)
+            case ("high", JInt(v)) => Some(v.toInt)
+            case ("n_buckets", JString(v)) => Some(v.toInt)
+            case ("n_buckets", JInt(v)) => Some(v.toInt)
+            case ("n_values", JString(v)) => Some(v.toInt)
+            case ("n_values", JInt(v)) => Some(v.toInt)
+            case ("kind", JString(v)) => Some(v)
+            case ("keyed", JBool(v)) => Some(v)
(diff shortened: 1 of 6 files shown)


src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+import telemetry.{DerivedStream, ObjectSummary}
+import telemetry.DerivedStream.s3
+import telemetry.avro
+import telemetry.heka.{HekaFrame, Message}
+import telemetry.histograms._
+import telemetry.parquet.ParquetFile
+
+case class Longitudinal() extends DerivedStream {
+  override def streamName: String = "telemetry-release"
+  override def filterPrefix: String = "telemetry/4/main/Firefox/release/*/*/*/42/"
```
mreid-moz: We should include `beta` as well.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+    if (!isS3PrefixEmpty(prefix)) {
+      println(s"Warning: prefix $prefix already exists on S3!")
+      return
+    }
+
+    val groups = DerivedStream.groupBySize(summaries.collect().toIterator)
+    val clientMessages = sc.parallelize(groups, groups.size)
+      .flatMap(x => x)
+      .flatMap{ case obj =>
+        val hekaFile = bucket.getObject(obj.key).getOrElse(throw new Exception("File missing on S3"))
```
mreid-moz: Add `obj.key` to the Exception message (helpful for debugging)

src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+  private def buildRecord(history: Iterable[Map[String, Any]], schema: Schema): Option[GenericRecord] = {
+    // Sort records by timestamp
+    val sorted = history
+      .toList
+      .sortWith((x, y) => {
+                 (x("creationTimestamp"), y("creationTimestamp")) match {
+                   case (creationX: Double, creationY: Double) =>
+                     creationX < creationY
+                   case _ =>
+                     return None  // Ignore 'unsortable' client
```
mreid-moz: Does this ignore all records for any client who submitted a record without "creationTimestamp"?

src/main/scala/histograms/Histogram.scala
```diff
@@ -0,0 +1,148 @@
...
+        JField(name, JObject(histogram)) <- root
+        JField(k, v) <- histogram
+      } yield {
+        val value = try {
+          (k, v) match {
+            case ("low", JString(v)) => Some(v.toInt)
+            case ("low", JInt(v)) => Some(v.toInt)
+            case ("high", JString(v)) => Some(v.toInt)
+            case ("high", JInt(v)) => Some(v.toInt)
+            case ("n_buckets", JString(v)) => Some(v.toInt)
```
mreid-moz: This will skip histograms that use non-numeric "expression" values in Histograms.json, right?

src/main/scala/streams/Longitudinal.scala
mreid-moz: I encountered this exception when backfilling the churn data, we may want to add a basic retry around the `bucket.getObject` call.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+          uploadLocalFileToS3(localFile, prefix)
+        }
+    }
+  }
+
+  private def buildSchema: Schema = {
+    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json
+    val buildType = SchemaBuilder
+      .record("build").fields()
+      .name("applicationId").`type`().optional().stringType()
```
mreid-moz: Is `applicationName` left out on purpose? Can it just be derived from `applicationId`?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+      .name("architecture").`type`().optional().stringType()
+      .name("architecturesInBinary").`type`().optional().stringType()
+      .name("buildId").`type`().optional().stringType()
```
mreid-moz: Is there a way to avoid repeating `.type().optional()` so much?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -0,0 +1,376 @@
...
+        case Some(histogram) =>
+          buffer += flatten(histogram)
+        case None =>
+          buffer += default
+      }
+    }
+    buffer.toArray
+  }
+
+  private def vectorizeHistogram[T:ClassTag](name: String,
```
mreid-moz: This whole section could use some comments describing how histograms of various types are mapped to different data structures.

mreid-moz:
I would like to instrument and log the number of records / clients that were discarded due to bad data. I'm not sure where best to add it (or even whether it's practical), but it would be very useful in ensuring that the data is relatively reliable and representative.

mreid-moz:
Another general comment - should we put an upper bound on the maximum number of pings we store for a single clientId (or maybe a max per submission day)?

src/main/scala/streams/Longitudinal.scala
vitillo: Yes.

src/main/scala/histograms/Histogram.scala
vitillo: Yes.

src/main/scala/streams/Longitudinal.scala
vitillo: The schema is not yet complete but at some point it will include applicationName as well.

vitillo:
We should have an upper limit but I am not sure what value is sensible. Once we have a better feeling of the distribution of pings per clients we can make that choice.

src/main/scala/histograms/Histogram.scala
mreid-moz: Ok, we should add a comment with that caveat and link to Bug 920169

src/main/scala/streams/Longitudinal.scala
vitillo: This is planned for a future version of the dataset. It's easier to validate having only a single channel.

vitillo:
@mreid-moz r?

mreid-moz:
Looks good.  I'd still like to explicitly mention that entries in Histograms.json that use "expressions" will be skipped, but let's get this landed.

mreid-moz:
r+

PR #17, opened 2016-02-04
Expand the schema to include all fixed ping fields

Uberi:
* Add schema stuff for all the fixed fields of pings.
* The SchemaBuilder code was initially generated by a quick, hacky Python script (included in this PR), and then fixed by hand to get the correct format.
    * For example, there are some invalid structures in the https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json that are ignored by the script.
    * Additional structures and fixes were added by hand from http://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry/environment.html.
* Added tests for these changes.
* Added required ping fields in the test cases.
* Although all tests pass, I don't have AWS credentials yet, so haven't tried it live.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1245489


src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,159 @@ case class Longitudinal() extends DerivedStream {
...
+        .name("buildId").`type`().stringType().noDefault()
+        .name("name").`type`().stringType().noDefault()
+        .name("version").`type`().stringType().noDefault()
+        .name("vendor").`type`().stringType().noDefault()
+        .name("platformVersion").`type`().stringType().noDefault()
+        .name("xpcomAbi").`type`().stringType().noDefault()
+        .name("channel").`type`().optional().stringType()
+      .endRecord()
+
+    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json commit 9993f0c0d3fc2e2dc16113070eb77554a72ce975
```
vitillo: Let's use a single comment to reference Sam's schema.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,159 @@ case class Longitudinal() extends DerivedStream {
...
+        .name("searchCohort").`type`().optional().stringType()
+        .name("e10sEnabled").`type`().optional().booleanType()
+        .name("telemetryEnabled").`type`().optional().booleanType()
+        .name("locale").`type`().optional().stringType()
+        .name("update").`type`().optional().record("update").fields()
+          .name("channel").`type`().optional().stringType()
+          .name("enabled").`type`().optional().booleanType()
+          .name("autoDownload").`type`().optional().booleanType()
+        .endRecord()
+        //.name("userPrefs").`type`().optional().map().values().stringType()
```
vitillo: Why is this commented?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,159 @@ case class Longitudinal() extends DerivedStream {
...
+      .endRecord()
```
vitillo: Doesn't reflect the new indentation scheme.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,159 @@ case class Longitudinal() extends DerivedStream {
...
 
   private def buildSchema: Schema = {
-    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json
+    // The generate-ping-schema.py script can be used to generate these SchemaBuilder definitions, and the output of that script is combined with data
+    // from http://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry/environment.html to get the results below
+
+    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/application.schema.json commit 9993f0c0d3fc2e2dc16113070eb77554a72ce975
+    val applicationType = SchemaBuilder
+      .record("application").fields()
+        .name("architecture").`type`().stringType().noDefault()
```
vitillo: Please make all fields optional as the data isn't validated upstream yet and we don't want to get an error for one missing field in a ping.

src/main/scala/streams/generate-ping-schema.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
```
vitillo: Please move this script under a top level directory called "scripts".

src/main/scala/streams/generate-ping-schema.py
```diff
@@ -0,0 +1,124 @@
...
+    if isinstance(schema, str): # primitive, named type
+        return """.{actual_type}Type(){default}""".format(
+            spacing = "  " * level, type_name = type_name,
+            is_optional = ".optional()" if is_optional else "",
+            default = "" if is_optional else ".noDefault()",
+            actual_type = schema["type"]
+        )
+
+    # skip the things we don't know how to handle
+    if "type" not in schema: return "<INVALID>"
```
vitillo: How did you generate the array sections in the schema if you don't handle the array type?

src/test/scala/Longitudinal.scala
```diff
@@ -17,69 +17,87 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
       // TODO: Use Scala Map and List directly?
       val histograms =
         ("TELEMETRY_TEST_FLAG" ->
-           ("values" -> ("0" -> 1) ~ ("1" -> 0)) ~
-           ("sum" -> 0)) ~
+           ("values" -> ("0" -> 1) ~ ("1" -> 0)) ~ ("sum" -> 0)) ~
```
vitillo: Please restore original indentation, same goes for the lines below.

src/test/scala/Longitudinal.scala
```diff
@@ -17,69 +17,87 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
...
+                ("RAM" -> 1024) ~ ("description" -> "FOO2") ~ ("deviceID" -> "2") ~ ("vendorID" -> "Vendor2") ~ ("GPUActive" -> false)
               )))
 
-      val settings =
-        ("e10sEnabled" -> true)
-
-      val build =
-        ("buildId" -> "20160101001100")
-
       Map("clientId" -> "26c9d181-b95b-4af5-bb35-84ebf0da795d",
```
vitillo: clientId doesn't respect the indentation

vitillo:
Can you open some issues in Sam's repository so that he is aware of the invalid structures?

vitillo:
Could you please add the `environment.addons` and the `info` sections of the pings as well? Note that the [structure](https://github.com/mozilla-services/data-pipeline/blob/master/heka/sandbox/decoders/extract_telemetry_dimensions.lua) of the pings is slightly different than the raw pings you have accessed in the past with `get_pings`.

src/main/scala/streams/generate-ping-schema.py
Uberi: These are done manually, as the JSON schema doesn't have the array items format anyways. I supplemented them with information from http://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry to ensure we're capturing all of the structure.

Uberi:
Changes:

* Implemented all the above feedback.
* Tested on a Spark cluster (see https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=telemetry-test-bucket&prefix=longitudinal/generationDate%3D20160203/ for results).
    * Only taking 5 objects per summary as you suggested.
* Submitted PR to data-pipeline to fix schema.
* All tests pass still.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,230 @@ case class Longitudinal() extends DerivedStream {
   }
 
   private def buildSchema: Schema = {
-    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json
+    // The $PROJECT_ROOT/scripts/generate-ping-schema.py script can be used to generate these SchemaBuilder definitions, and the output of that script is combined with data
+    // from http://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry to get the results below
+    // See the schemas at https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test commit 9993f0c0d3fc2e2dc16113070eb77554a72ce975
+    val applicationType = SchemaBuilder
+      .record("application").fields()
```
vitillo: The application record seems to be missing from the generated dataset.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,230 @@ case class Longitudinal() extends DerivedStream {
...
-      .name("defaultSearchEngine").`type`().optional().booleanType()
-      .name("e10sEnabled").`type`().optional().booleanType()
-      .name("locale").`type`().optional().stringType()
-      .name("telemetryEnabled").`type`().optional().booleanType()
+        .name("addonCompatibilityCheckEnabled").`type`().optional().booleanType()
+        .name("blocklistEnabled").`type`().optional().booleanType()
+        .name("isDefaultBrowser").`type`().optional().booleanType()
+        .name("defaultSearchEngine").`type`().optional().stringType()
+        .name("defaultSearchEngineData").`type`().optional().record("defaultSearchEngineData").fields()
+          .name("name").`type`().optional().stringType()
```
vitillo: defaultSearchEngineData.name doesn't look right in the generated dataset, e.g. `Row(name=u'\u042f\u043d\u0434\u0435\u043a\u0441,...)`

src/main/scala/streams/Longitudinal.scala
```diff
@@ -66,78 +66,230 @@ case class Longitudinal() extends DerivedStream {
...
-      .record("Submission")
-      .fields
-      .name("clientId").`type`().stringType().noDefault()
-      .name("os").`type`().stringType().noDefault()
-      .name("creationTimestamp").`type`().array().items().doubleType().noDefault()
-      .name("system").`type`().optional().array().items(systemType)
-      .name("build").`type`().optional().array().items(buildType)
-      .name("settings").`type`().optional().array().items(settingsType)
+    val addonsType = SchemaBuilder
+      .record("addons").fields()
```
vitillo: This sections is huge and we should split it as Spark can't fetch (yet) a single field of a record without fetching all other fields as well. Let's create a top-level record for:
- activeAddons
- theme
- activePlugins
- activeGMPlugins
- activeExperiment
- persona

vitillo:
With your patch 99% of the clients are missing. Please make sure to test it thoroughly on real data, i.e. you should load the data in Spark and make sure it looks right. Note that all exceptions [are catched](https://github.com/vitillo/telemetry-batch-view/blob/longitudinal/src/main/scala/streams/Longitudinal.scala#L393) to deal with problematic clients; you might want to disable that behaviour for debugging purposes. You should also rebase your patch on the latest `longitudinal` branch, so that you get some statistics about the number of clients ignored at the end of the job.

src/main/scala/streams/Longitudinal.scala
Uberi: This is simply "Yandex" in Cryllic, so I'm guessing it's probably correct, just without non-Latin characters displayed.

Uberi:
Changes:

* Rebased off upstream
* Fix double conversion issue that makes all those clients not show up (tested on a Spark cluster).
* Add threadHangStats.
* Add simpleMeasurements.
* JSON4S had several bugs that there are no workarounds for, so I had to bump it up to 3.2.11.
* Various other small fields.

vitillo:
The reason why we are using an older JSON4S version is that we have to use the same one that Spark 1.5.2 uses or jobs launched with `spark-submit` blow up. 

Have you tried running your job with `spark-submit`? That's what we use to run on multiple machines and I am pretty sure that you are going to see some weird errors.

Is there a way to work around the limitations you encountered and not change the library's version?

vitillo:
The longitudinal branch has been merged into master so you should submit a PR against the latter which has been updated since then.

PR #19, opened 2016-02-10
Longitudinal dataset fields

Uberi:
Continued from #17.

Also, it adds a TravisCI config, which, if enabled at travis-ci.org, will automatically run tests on every push.


vitillo:
When generating the longitudinal dataset for a single day on a 20 machine cluster I noticed that a large fraction of clients are getting ignored now. Do you know why?

```
[hadoop]$ spark-submit —master yarn-client —class telemetry.DerivedStream target/scala-2.10/telemetry-batch-view-1.1.jar —from-date 20160209 —to-date 20160209 Longitudinal
Spark parallelism level: 320
Clients seen: 1035425
Clients ignored: 63246
```

vitillo:
More concerning is that the job didn't output any file at all on S3.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -358,9 +535,83 @@ case class Longitudinal() extends DerivedStream {
...
+        (name -> mapAsJavaMap(stackMapping))
+      }.toMap
+      mapAsJavaMap(threadHangStackMap)
+    } toArray
+
+    root.set("threadHangActivity", threadHangActivityList)
+    root.set("threadHangStacks", threadHangStackMapList)
+  }
+
+  private def simpleMeasurements2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
```
vitillo: Using a map is not the best way to deal with this data as some queries become harder to perform. The simpleMeasurements section should have a schema like the other sections containing the following integer fields:
- 'activeTicks',
- 'profileBeforeChange',
- 'selectProfile',
- 'sessionRestoreInit',
- 'firstLoadURI',
- 'uptime',
- 'totalTime',
- 'savedPings',
- 'start',
- 'startupSessionRestoreReadBytes',
- 'pingsOverdue',
- 'firstPaint',
- 'sessionRestored',
- 'startupWindowVisibleWriteBytes',
- 'startupCrashDetectionEnd',
- 'startupSessionRestoreWriteBytes',
- 'startupCrashDetectionBegin',
- 'startupInterrupted',
- 'afterProfileLocked',
- 'delayedStartupStarted',
- 'main',
- 'createTopLevelWindow',
- 'sessionRestoreInitialized',
- 'maximalNumberOfConcurrentThreads',
- 'startupWindowVisibleReadBytes']

build.sbt
```diff
@@ -1,4 +1,4 @@
-javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint")
+javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint", "-Xss2M")
```
vitillo: Why was this change needed?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -83,77 +83,225 @@ case class Longitudinal() extends DerivedStream {
...
+        .name("sessionLength").`type`().optional().longType()
+        .name("sessionStartDate").`type`().optional().stringType()
+        .name("subsessionCounter").`type`().optional().intType()
+        .name("subsessionId").`type`().optional().stringType()
+        .name("subsessionLength").`type`().optional().longType()
+        .name("subsessionStartDate").`type`().optional().stringType()
+        .name("timezoneOffset").`type`().optional().intType()
+      .endRecord()
     val buildType = SchemaBuilder
       .record("build").fields()
```
vitillo: The [application](https://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry/common-ping.html) section of the ping is missing.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -83,77 +83,225 @@ case class Longitudinal() extends DerivedStream {
   }
 
   private def buildSchema: Schema = {
-    // See https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test/environment.schema.json
+    // The $PROJECT_ROOT/scripts/generate-ping-schema.py script can be used to generate these SchemaBuilder definitions, and the output of that script is combined with data
+    // from http://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry to get the results below
+    // See the schemas at https://github.com/SamPenrose/data-pipeline/blob/v4-baseset-schema/schemas/test commit 9993f0c0d3fc2e2dc16113070eb77554a72ce975
+    val infoType = SchemaBuilder
```
vitillo: We should add a top-level field just for the channel.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -83,77 +83,225 @@ case class Longitudinal() extends DerivedStream {
   }
 
   private def buildSchema: Schema = {
```
vitillo: We should add a top-level field just for the channel.

vitillo:
The rest looks excellent, once we fix the above mentioned issues we can create a new dataset and advertise it on the proper channels.

build.sbt
Uberi: Some of the longer SchemaBuilder statements seem to cause stack overflows in the compiler, and increasing the stack size was the recommended way to solve these issues. It's technically not necessary because users could just set the envvars themselves, but it does make it easier to build out of the box.

Uberi:
Changes:

* Implemented all the feedback above.
* Tested changes on 1.2 million pings from Jan 28, 2016 (all pings successfully parsed): https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=telemetry-test-bucket&prefix=longitudinal/generation_date%3D20160128/

src/main/scala/streams/Longitudinal.scala
```diff
@@ -358,47 +565,139 @@ case class Longitudinal() extends DerivedStream {
...
+    root.set("threadHangActivity", threadHangActivityList)
+    root.set("threadHangStacks", threadHangStackMapList)
+  }
+
+  private def simpleMeasurements2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
+    implicit val formats = DefaultFormats
+
+    // we need to handle simple measures differently because there are simpleMesaurements that have sub-values,
+    // which we'll ignore since they're not used much - only integer simpleMeasurements are included
+    val simpleMeasurements = payloads.map{ case (x) =>
```
vitillo: Do we still need this?

vitillo:
Looks good, thanks!

PR #20, opened 2016-02-11
Include info type again

Uberi:
This got excluded in the previous merge.

A small test to make sure everything's still sane:

    Clients seen: 449
    Clients ignored: 0


vitillo:
Looks good, thanks.

PR #23, opened 2016-02-14
Pack more profiles in a row group.

vitillo:
The longitudinal dataset has only a few hundred profiles per row group, which is detrimental for performance. This patch increases the number of profiles per row group while simultaneously reducing the memory footprint of the job, without which it wouldn't be possible to run on a typical c3.4xlarge cluster.

Furthermore, the number of cores per executor has to be reduced by 2x and the storage memory fraction has to be set to 0 when running the job, e.g.:
```
spark-submit --master yarn-client --executor-cores 8 --class telemetry.DerivedStream \
  --conf spark.storage.memoryFraction=0 ./target/scala-2.10/telemetry-batch-view-1.1.jar \
  --from-date 20151115 --to-date 20160213 Longitudinal
```
Note that even though the number of executors is halved, the time of the job reduces to about 2h over the time period as there is less swapping.

### DIFF
--- a/src/main/scala/parquet/ParquetFile.scala
+++ b/src/main/scala/parquet/ParquetFile.scala
@@ -28,8 +28,8 @@ object ParquetFile {
     return new Path(uri)
   }
 
-  def serialize(data: Iterator[GenericRecord], schema: Schema): Path = {
-    val blockSize = ParquetWriter.DEFAULT_BLOCK_SIZE
+  def serialize(data: Iterator[GenericRecord], schema: Schema, blockSizeMultiplier: Int = 1): Path = {
+    val blockSize = blockSizeMultiplier*ParquetWriter.DEFAULT_BLOCK_SIZE
     val pageSize = ParquetWriter.DEFAULT_PAGE_SIZE
     val enableDict = ParquetWriter.DEFAULT_IS_DICTIONARY_ENABLED
     val parquetFile = temporaryFileName()
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -114,6 +114,11 @@ case class Longitudinal() extends DerivedStream {
       .repartitionAndSortWithinPartitions(new ClientIdPartitioner(320))
       .map{case (key, value) => (key._1, value)}
 
+    /* One file per partition is generated at the end of the job. We want to have
+       few big files but at the same time we need a high enough degree of parallelism
+       to keep all workers busy. Since the cluster typically used for this job has
+       8 workers and 20 executors, 320 partitions provide a good compromise. */
+
     val partitionCounts = clientMessages
       .mapPartitions{ case it =>
         val clientIterator = new ClientIterator(it)
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -4,14 +4,14 @@ import awscala._
 import awscala.s3._
 import org.apache.avro.{Schema, SchemaBuilder}
 import org.apache.avro.generic.{GenericData, GenericRecord, GenericRecordBuilder}
-import org.apache.spark.SparkContext
+import org.apache.spark.{SparkContext, Partitioner}
 import org.apache.spark.rdd.RDD
 import org.json4s._
 import org.json4s.jackson.JsonMethods._
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
-import scala.math.max
+import scala.math.{max, abs}
 import scala.reflect.ClassTag
 import telemetry.{DerivedStream, ObjectSummary}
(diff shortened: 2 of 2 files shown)


src/main/scala/streams/Longitudinal.scala
```diff
@@ -20,6 +20,64 @@ import telemetry.heka.{HekaFrame, Message}
 import telemetry.histograms._
 import telemetry.parquet.ParquetFile
 
+private class ClientIdPartitioner(size: Int) extends Partitioner{
+  def numPartitions: Int = size
+  def getPartition(key: Any): Int = key match {
+    case (clientId: String, startDate: String, counter: Int) => abs(clientId.hashCode) % size
```
mreid-moz: Should we use crc32(clientId) here for consistency with sampleId?

vitillo: I don't see a strong reason why.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -40,19 +100,25 @@ case class Longitudinal() extends DerivedStream {
...
+          json <- fields.get("payload.info").asInstanceOf[Option[String]]
+          info = parse(json)
+          tmp = for {
+            JString(startDate) <- info \ "subsessionStartDate"
+            JInt(counter) <- info \ "profileSubsessionCounter"
+          } yield (startDate, counter)
+          (startDate, counter) <- tmp.headOption
+        } yield ((clientId, startDate, counter.toInt), fields)
+      }
+      .repartitionAndSortWithinPartitions(new ClientIdPartitioner(320))
```
mreid-moz: Please comment the magic value of `320`

src/main/scala/streams/Longitudinal.scala
```diff
@@ -20,6 +20,64 @@ import telemetry.heka.{HekaFrame, Message}
...
 
+private class ClientIdPartitioner(size: Int) extends Partitioner{
+  def numPartitions: Int = size
+  def getPartition(key: Any): Int = key match {
+    case (clientId: String, startDate: String, counter: Int) => abs(clientId.hashCode) % size
+    case _ => throw new Exception("Invalid key")
+  }
+}
+
+class ClientIterator(it: Iterator[Tuple2[String, Map[String, Any]]], maxHistorySize: Int = 1000) extends Iterator[List[Map[String, Any]]]{
```
mreid-moz: Is there a way to detect from the resulting longitudinal data that a client's history has been truncated?  If not, can we add that to the data?

src/main/scala/parquet/ParquetFile.scala
```diff
@@ -28,8 +28,8 @@ object ParquetFile {
     return new Path(uri)
   }
 
-  def serialize(data: Iterator[GenericRecord], schema: Schema): Path = {
-    val blockSize = ParquetWriter.DEFAULT_BLOCK_SIZE
+  def serialize(data: Iterator[GenericRecord], schema: Schema, blockSizeMultiplier: Int = 1): Path = {
+    val blockSize = blockSizeMultiplier*ParquetWriter.DEFAULT_BLOCK_SIZE
```
mreid-moz: nit: add spaces around `*`

src/main/scala/streams/Longitudinal.scala
vitillo: Counting the number of clients with a history length == N yields a good approximation of the number of clients with truncated histories.

src/main/scala/streams/Longitudinal.scala
vitillo: One file per partition is generated at the end of the job. We want to have few big files but at the same time we need a high enough degree of parallelism to keep all workers busy. Since the cluster typically used for this job has 8 workers and 20 executors, 320 tasks provides a good compromise.

src/main/scala/streams/Longitudinal.scala
mreid-moz: I don't think you need the `abs` here - modulo on negative values should work fine.

vitillo: It's not, it yields a negative value in Scala.

src/main/scala/streams/Longitudinal.scala
mreid-moz: Sounds good - can you add that description as a code comment?

src/main/scala/streams/Longitudinal.scala
mreid-moz: oops, my mistake

src/main/scala/streams/Longitudinal.scala
vitillo: Ok

src/main/scala/parquet/ParquetFile.scala
vitillo: I tend to leave spaces only around addition and subtraction to be closer to the mathematical convention of not leaving spaces around multiplications and divisions.

PR #24, opened 2016-02-16
Add geoCountry and geoCity fields

Uberi:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1247668

All the other fields mentioned in the bug are already present.


vitillo:
Looks good, thanks!

PR #25, opened 2016-02-17
Add versioning scheme to datasets.

vitillo:
We don't currently have a homogeneous versioning convention which is
needed by parquet2hive.


mreid-moz:
Looks good to me. r+

PR #27, opened 2016-02-18
Fix geoCountry generally being missing

Uberi:
Before, geoCountry would be omitted whenever geoCity is missing. However, :chutten mentioned that it's still useful to have geoCountry even this is the case.

The geoCountry/geoCity is set to an empty string because using nullables in arrays seems to expand the schema quite a bit. This seems to be a known issue with Avro.

### DIFF
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -636,9 +636,11 @@ case class Longitudinal() extends DerivedStream {
       x.getOrElse("geoCity", "").asInstanceOf[String]
     }
 
-    // only set the keys if there are valid country entries
+    // only set the keys if there are valid entries
     if (countries.exists( country => country != "" )) {
       root.set("geoCountry", countries.toArray)
+    }
+    if (cities.exists( city => city != "" )) {
       root.set("geoCity", cities.toArray)
     }
   }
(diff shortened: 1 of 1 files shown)


src/main/scala/streams/Longitudinal.scala
```diff
@@ -630,10 +630,10 @@ case class Longitudinal() extends DerivedStream {
 
   private def geo2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
     val countries = payloads.map{ case (x) =>
-      x.getOrElse("geoCountry", return).asInstanceOf[String]
+      x.getOrElse("geoCountry", "").asInstanceOf[String]
```
vitillo: To reflect the way we are handling the other fields the following should happen:
- if geoCountry is present in at least one fragment of the client's history, then all the other fragment's geoCountry fields are initialized to "" if they are missing
- if geoCountry is missing in all fragments of a client's history, then the geoCountry field is null (and there is no array).
Same goes for geoCity.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -630,14 +630,17 @@ case class Longitudinal() extends DerivedStream {
...
     }
     val cities = payloads.map{ case (x) =>
-      x.getOrElse("geoCity", return).asInstanceOf[String]
+      x.getOrElse("geoCity", "").asInstanceOf[String]
     }
 
-    root.set("geoCountry", countries.toArray)
-    root.set("geoCity", cities.toArray)
+    // only set the keys if there are valid country entries
+    if (countries.exists( country => country != "" )) {
```
vitillo: Unless geoCity is present if and only if geoCountry is, you should probably treat them separately.

vitillo:
r+

PR #28, opened 2016-02-19
Split out info fields

Uberi:
Based on #27 (which I updated, by the way)

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1249624

Quick test with 500 pings: https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=telemetry-test-bucket&prefix=longitudinal/generation_date%3D20160128/


vitillo:
r+

PR #29, opened 2016-02-23
Use underscore_notation for field names in longitudinal dataset.

vitillo:
Presto can't access camelCased fields within structs therefore
all fields and column names have to be converted to underscore_notation.


Uberi:
Looks good to me :+1:

PR #30, opened 2016-02-24
Add additional fields of interest.

Uberi:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1250889

### DIFF
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -381,6 +381,7 @@ case class Longitudinal() extends DerivedStream {
         .name("creation_timestamp").`type`().optional().array().items().doubleType()
         .name("geo_country").`type`().optional().array().items().stringType()
         .name("geo_city").`type`().optional().array().items().stringType()
+        .name("dnt_header").`type`().optional().array().items().stringType()
         .name("addons").`type`().optional().array().items().stringType()
         .name("async_plugin_init").`type`().optional().array().items().booleanType()
         .name("flash_version").`type`().optional().array().items().stringType()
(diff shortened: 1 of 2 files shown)


src/main/scala/streams/Longitudinal.scala
```diff
@@ -702,20 +707,17 @@ case class Longitudinal() extends DerivedStream {
...
-    val countries = payloads.map{ case (x) =>
-      x.getOrElse("geoCountry", "").asInstanceOf[String]
-    }
-    val cities = payloads.map{ case (x) =>
-      x.getOrElse("geoCity", "").asInstanceOf[String]
+  private def value2Avro[T:ClassTag](field: String, avroField: String, default: T,
+                          payloads: List[Map[String, Any]],
+                          root: GenericRecordBuilder,
+                          schema: Schema) {
+    val values = payloads.map{ case (x) =>
```
vitillo: See comments in #31.

Uberi:
Changes:

* Cherry picked "Implement code review feedback" from `fix-dates`
* #31 is also updated

Uberi:
Changes:

* Cherry-picked "Implement code review feedback" from `fix-dates`
* Update #31

vitillo:
Looks good, I am waiting for @mreid-moz to comment on the bug.

vitillo:
"The web server that receives Telemetry submissions stores the actual DNT header 
value the client sends. This is more comprehensive than the DNT_USAGE 
histogram, because it is included for all document types (even ones without 
histograms), and that histogram is only being collected for clients who opt in 
to extended telemetry. "

Uberi:
The latest commit adds this field. `dnt_header` will be either the string `0`, `1`, or some other unknown value. As a result, it's a string. This field doesn't seem to be present in most telemetry pings from the pipeline.

vitillo:
Looks good, thanks.

PR #31, opened 2016-02-24
Fix dates with weird time zones.

Uberi:
Joda-time seems to have very standards-compliant parsing and output.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1250894

### DIFF
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -718,9 +718,10 @@ case class Longitudinal() extends DerivedStream {
           // certain steps later in the pipeline have a hard time with certain date edge cases,
           // especially time zone offsets that are not between -12 and 14 hours inclusive (see bug 1250894)
           // we're going to use the relatively lenient joda-time parser and output it in a very standard format
-          // note that in the process the timestamp will be converted into a local time, though it will still represent the same instant in time
+          // note that in the process the timestamp will be converted into UTC, though it will still represent the same instant in time
+          val dateFormatter = org.joda.time.format.ISODateTimeFormat.dateTime()
           val date = new DateTime(value);
-          date.toString(org.joda.time.format.ISODateTimeFormat.dateTime());
+          dateFormatter.withZone(org.joda.time.DateTimeZone.UTC).print(date)
         case _ =>
           return
       }
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -822,7 +822,7 @@ case class Longitudinal() extends DerivedStream {
       val formatISO = org.joda.time.format.ISODateTimeFormat.dateTime()
       val formatYYYYMMDD = org.joda.time.format.DateTimeFormat.forPattern("yyyyMMdd")
       val reformatYYYYMMDD = (value: Any) =>
-        formatISO.withZone(org.joda.time.DateTimeZone.UTC).print(formatYYYYMMDD.parseDateTime(value.asInstanceOf[String]))
+        formatISO.withZone(org.joda.time.DateTimeZone.UTC).print(formatYYYYMMDD.withZone(org.joda.time.DateTimeZone.UTC).parseDateTime(value.asInstanceOf[String]))
       val reformatTimestamp = (value: Any) =>
         formatISO.withZone(org.joda.time.DateTimeZone.UTC).print(new DateTime((value.asInstanceOf[Double] / 1e6).toLong))
       subsessionStartDate2Avro(sorted, root, schema)
--- a/src/main/scala/streams/Longitudinal.scala
+++ b/src/main/scala/streams/Longitudinal.scala
@@ -8,6 +8,7 @@ import org.apache.spark.{SparkContext, Partitioner}
 import org.apache.spark.rdd.RDD
 import org.json4s._
 import org.json4s.jackson.JsonMethods._
+import org.joda.time
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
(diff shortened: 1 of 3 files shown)


vitillo:
Tests are failing.

Uberi:
Yep, I think this latest commit should fix the issue. Note that it will make all timestamps UTC, though.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -702,20 +708,39 @@ case class Longitudinal() extends DerivedStream {
...
+
+    val fieldValues = payloads.map{ case (x) =>
+      var record = parse(x.getOrElse("payload.info", return).asInstanceOf[String]) \ "subsessionStartDate"
+      record match {
+        case JString(value) =>
+          // certain steps later in the pipeline have a hard time with certain date edge cases,
+          // especially time zone offsets that are not between -12 and 14 hours inclusive (see bug 1250894)
+          // we're going to use the relatively lenient joda-time parser and output it in a very standard format
+          // note that in the process the timestamp will be converted into UTC, though it will still represent the same instant in time
+          val dateFormatter = org.joda.time.format.ISODateTimeFormat.dateTime()
```
vitillo: There is no need to instantiate a new formatter for each value, could you move this above the `fieldValues` declaration?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -702,20 +708,39 @@ case class Longitudinal() extends DerivedStream {
...
+          return
+      }
     }
-    val cities = payloads.map{ case (x) =>
-      x.getOrElse("geoCity", "").asInstanceOf[String]
+
+    root.set("subsession_start_date", fieldValues.toArray)
+  }
+
+  private def value2Avro[T:ClassTag](field: String, avroField: String, default: T,
```
vitillo: Please indent parameters uniformely.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -702,20 +708,39 @@ case class Longitudinal() extends DerivedStream {
...
+                          payloads: List[Map[String, Any]],
+                          root: GenericRecordBuilder,
+                          schema: Schema) {
+    val values = payloads.map{ case (x) =>
+      x.getOrElse(field, default).asInstanceOf[T]
```
vitillo: Can't you use a return instead of a default value that you later use to decide if to keep the values or not?

src/test/scala/Longitudinal.scala
```diff
@@ -131,7 +131,7 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
             ("branch" -> "B"))
 
       val info =
-        ("subsessionStartDate"      -> "2015-12-09T00:00:00.0-08:00") ~
+        ("subsessionStartDate"      -> "2015-12-09T00:00:00.0-13:00") ~
```
vitillo: Why did you change the offset?

src/main/scala/streams/Longitudinal.scala
```diff
@@ -702,20 +708,39 @@ case class Longitudinal() extends DerivedStream {
     root.set("thread_hang_stacks", threadHangStackMapList)
   }
 
-  private def geo2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
-    val countries = payloads.map{ case (x) =>
-      x.getOrElse("geoCountry", "").asInstanceOf[String]
+  private def subsessionStartDate2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
```
vitillo: @mreid-moz do we have other date fields that need special treatment?

mreid-moz: We have the following types of fields that would be nice to handle as dates:
1. strings of the form `2016-02-25T13:57:34.002Z` (such as `creationDate`)
2. strings of the form `2016-02-23T00:00:00.0-04:00` (such as `subsessionStartDate`)
3. strings of the form `20160223` (such as `meta/submissionDate`)
4. "number of days since epoch" fields (such as `profile.creationDate`)
5. nanos since epoch (such as the system-added `meta/Timestamp` field)
There's a field for `gfx.adapters.driverDate`, but I'm not sure what form it takes (it may not be consistent). There may also be others in parts of the payload I'm not familiar with (Telemetry Experiments, UITelemetry), but the above are definitely variations that exist in the current data.

mreid-moz:
> Yep, I think this latest commit should fix the issue. Note that it will make all timestamps UTC, though.
I believe the metrics team wanted a way to tell what time zone a client is in.

src/main/scala/streams/Longitudinal.scala
Uberi: Using `return` would be equivalent to "don't set if any of them are unset", while using a default value would be equivalent to "don't set if all of them are unset".

Is the first behaviour more desirable? I assume we'd want to keep the data if possible.

Uberi:
Since the timestamp timezones are dependent on the user's system clock  settings, would it make more sense to use `geoCountry` instead (i.e., `geoCountry` is a better indicator of what time zone a client is in)?

Another option is to only convert it to UTC if the timezone is out of the `-12, 14` range. This feels a little hacky, though.

src/main/scala/streams/Longitudinal.scala
vitillo: Right, the current behaviour is the more desirable.

vitillo:
I would prefer to use `geoCountry` instead of the offset in the timestamp.

src/main/scala/streams/Longitudinal.scala
vitillo: We should convert `creationDate`, `subsessionStartDate`, `meta.submissionDate`, `profile.creationDate` to a UTC date.

mreid-moz:
I don't have a compelling use case for that information, so maybe just go ahead with the original plan (use UTC dates) and we'll see if `geoCountry` is enough.

src/main/scala/streams/Longitudinal.scala
```diff
@@ -763,16 +811,23 @@ case class Longitudinal() extends DerivedStream {
...
+
+      val formatISO = org.joda.time.format.ISODateTimeFormat.dateTime()
+      val formatYYYYMMDD = org.joda.time.format.DateTimeFormat.forPattern("yyyyMMdd")
+      val reformatYYYYMMDD = (value: Any) =>
+        formatISO.withZone(org.joda.time.DateTimeZone.UTC).print(formatYYYYMMDD.withZone(org.joda.time.DateTimeZone.UTC).parseDateTime(value.asInstanceOf[String]))
+      val reformatTimestamp = (value: Any) =>
+        formatISO.withZone(org.joda.time.DateTimeZone.UTC).print(new DateTime((value.asInstanceOf[Double] / 1e6).toLong))
+      subsessionStartDate2Avro(sorted, root, schema)
+      value2Avro("submissionDate", "submission_date", "", reformatYYYYMMDD, sorted, root, schema)
+      value2Avro("creationTimestamp", "creation_date", "", reformatTimestamp, sorted, root, schema)
```
vitillo: Can't we use directly [creationDate](https://github.com/mozilla-services/data-pipeline/blob/master/heka/sandbox/decoders/extract_telemetry_dimensions.lua#L159) instead of creationTimestamp?

vitillo:
`profile.creationDate` is still missing.

src/main/scala/streams/Longitudinal.scala
Uberi: creationDate is buried in a JSON object, which makes it a little messier. It's simple to work around it, but we need to handle one less case this way. Would using creationDate be preferable?

vitillo: That makes sense, thanks.

Uberi:
Changes:

* Added fields for `creationDate` and `resetDate`.
* Removed the `creationTimestamp` field since it's just computed from `creationDate` in the data pipeline.
* Update tests to account for these changes.

vitillo:
@Uberi The branch has conflicts that must be resolved.

vitillo:
It seems that Avro has [recently](https://issues.apache.org/jira/browse/AVRO-739) started supporting a date/time type. We should test if using that datatype allows us to read dates from Spark and Presto directly without any form of manual parsing.

vitillo:
Nevermind, Spark 1.6 uses an older Avro version and as such we can't upgrade the one used in our job without the likelihood of entering the JAR hell.

PR #32, opened 2016-03-01
Fix dates with weird time zones.

vitillo:
Joda-time seems to have very standards-compliant parsing and output.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1250894, #31 .


vitillo:
@Uberi I had to revert your changes as the data doesn't look right, e.g. in profile_creation_date I see the following from Spark: "profile_creation_date=u'[Ljava.lang.String;@189c048f', ...". Please test it properly by trying to access to the from Spark.

PR #34, opened 2016-03-01
Fix date parsing

Uberi:
Changes from last PR:

* Rebased off master.
* Fixed errors caused by merge and the rebase.
* Profile creation time is actually days since epoch rather than a unix timestamp.
* Change tests to match.

### DIFF
--- a/src/test/scala/Longitudinal.scala
+++ b/src/test/scala/Longitudinal.scala
@@ -195,12 +195,11 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
 
   "top level fields" must "be converted correctly" in {
     val fieldValues = Array(
-      "submission_date"    -> "20160128",
-      "sample_id"          -> 42.0,
-      "size"               -> 93691.0,
-      "creation_timestamp" -> 1.45393974518300006E18,
-      "geo_country"        -> "US",
-      "geo_city"           -> "New York",
+      "submission_date"       -> "2016-01-28T00:00:00.000Z",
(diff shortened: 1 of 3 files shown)


vitillo:
@Uberi Have you tried reading back a small dataset with Spark and check if the dates are correctly deserialized?

Uberi:
screenshot from 2016-03-02 13-18-57

It seems like all of the new date fields are being correctly parsed and are well formed ISO8601 timestamps.

src/test/scala/Longitudinal.scala
```diff
@@ -195,21 +195,17 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
...
   "top level fields" must "be converted correctly" in {
     val fieldValues = Array(
-      "submission_date"    -> "20160128",
-      "sample_id"          -> 42.0,
-      "size"               -> 93691.0,
-      "creation_timestamp" -> 1.45393974518300006E18,
-      "geo_country"        -> "US",
-      "geo_city"           -> "New York",
-      "dnt_header"         -> "1"
+      "submission_date"       -> "2016-01-28T00:00:00.000Z",
```
vitillo: It might be worthwhile to have a date with an offset that is outside the ISO bounds so that we can check if the conversion is correct.

Uberi:
Changes:

* Added test for invalid timezone.
* #35 should be merged before putting this into production.

vitillo:
Looks good, thanks.

PR #36, opened 2016-03-03
Include the timezone, but normalize it if needed

Uberi:
It's kind of hacky, but it works!

All dates within real timezones are kept as-is. All dates outside are brought inside by adding/subtracting 12 hours.


Uberi:
Changes:

* Move all time handling into Utils.
* Accept all timezones, even those outside of the standard range.

vitillo:
Looks good, thanks!

PR #38, opened 2016-03-08
Increase ThreadStackSize to 4MB.

mreid-moz:
When running `sbt assembly`, I kept getting the error:
`[error] (compile:compileIncremental) java.lang.StackOverflowError`

I tried modifying the `-Xss` setting in `build.sbt`, but it had no
effect (I don't think that's even a valid `javac` argument). I
was able to apply the setting by adding it in the `.sbtopts` file.


vitillo:
Looks good, thanks!

PR #40, opened 2016-03-10
Split system into multiple top-level fields

Uberi:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1255465


src/main/scala/streams/Longitudinal.scala
```diff
@@ -791,6 +801,11 @@ case class Longitudinal() extends DerivedStream {
       JSON2Avro("environment.partner",        List[String](),                   "partner", sorted, root, schema)
       JSON2Avro("environment.settings",       List[String](),                   "settings", sorted, root, schema)
       JSON2Avro("environment.system",         List[String](),                   "system", sorted, root, schema)
+      JSON2Avro("environment.system",         List[String]("cpu"),              "system_cpu", sorted, root, schema)
```
vitillo: There is no need to specify the type here as it is inferred.

vitillo:
Looks good, thanks!

PR #41, opened 2016-03-15
Add KPI dataset.

vitillo:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1253644.

### DIFF
--- a/src/main/scala/views/KPI.scala
+++ b/src/main/scala/views/KPI.scala
@@ -1,0 +1,62 @@
+package telemetry.views
+
+import com.github.nscala_time.time.Imports._
+import com.typesafe.config._
+import org.apache.spark.{SparkConf, SparkContext}
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.hive.HiveContext
+import org.rogach.scallop._
+import telemetry.spark.sql.aggregates._
+
+class Conf(args: Array[String]) extends ScallopConf(args) {
+  val from = opt[String]("from", descr = "From submission date", required = false)
+  val to = opt[String]("to", descr = "To submission date", required = false)
+  verify()
+}
+
+object KPIView {
+  private val hll = new HyperLogLog(10)
+  private val base = List("normalizedChannel", "country", "version", "e10sEnabled", "e10sCohort")
+  private val selection = "clientId" :: "substr(subsessionStartDate, 0, 10) as activityDate" :: base
+
+  val dimensions = "activityDate" :: base
+
+  def aggregate(frame: DataFrame): DataFrame = {
+    frame
+      .selectExpr(selection:_*)
+      .groupBy(col1=dimensions.head, cols=dimensions.tail:_*)
+      .agg(hll(col("clientId")).as("hll"))
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args)
+    val fmt = DateTimeFormat.forPattern("yyyyMMdd")
+
+    val to = conf.to.get match {
+      case Some(t) => fmt.print(fmt.parseDateTime(t))
+      case _ => fmt.print(DateTime.now.minusDays(1))
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/scala/views/KPI.scala
```diff
@@ -0,0 +1,62 @@
...
+import telemetry.spark.sql.aggregates._
+
+class Conf(args: Array[String]) extends ScallopConf(args) {
+  val from = opt[String]("from", descr = "From submission date", required = false)
+  val to = opt[String]("to", descr = "To submission date", required = false)
+  verify()
+}
+
+object KPIView {
+  private val hll = new HyperLogLog(10)
```
mreid-moz: why 10?

vitillo: I have been using a low value just for testing purposes.

src/main/scala/views/KPI.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.rogach.scallop._
+import telemetry.spark.sql.aggregates._
+
+class Conf(args: Array[String]) extends ScallopConf(args) {
+  val from = opt[String]("from", descr = "From submission date", required = false)
+  val to = opt[String]("to", descr = "To submission date", required = false)
+  verify()
+}
+
+object KPIView {
```
mreid-moz: I don't think `KPI` is a very informative name, I would prefer something like "ClientCountView"

src/main/scala/views/KPI.scala
```diff
@@ -0,0 +1,62 @@
...
+    val hadoopConf = sc.hadoopConfiguration
+    hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
+
+    val df = sqlContext.read.load("s3://telemetry-parquet/churn/v1")
+    val subset = df.where(s"submission_date_s3 >= $from and submission_date_s3 <= $to")
+    val aggregates = aggregate(subset)
+
+    val appConf = ConfigFactory.load()
+    val parquetBucket = appConf.getString("app.parquetBucket")
+    aggregates.write.parquet(s"s3://$parquetBucket/KPI/v$from$to")
```
mreid-moz: Is Submission Date one of the dimensions we'll be using? If so, maybe we should use a path like `s3://bucket/viewname/vX/submission_date_s3=YYYYmmdd/<files>`

Then we can incrementally update the aggregates each day.

src/main/scala/views/KPI.scala
mreid-moz: We should add a comment in the source that either links to the table [here](https://github.com/twitter/algebird/blob/0.10.0/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala#L194-L210) or just notes that the expected error rate using 10 bits is ~3%.

src/test/scala/KPITest.scala
```diff
@@ -0,0 +1,70 @@
...
+    import sqlContext.implicits._
+
+    val dataset = sc.parallelize(Submission.randomList).toDF()
+    val aggregates = KPIView.aggregate(dataset)
+
+    val dimensions = Set(KPIView.dimensions:_*) -- Set("clientId", "hll")
+    (Set(aggregates.columns:_*) -- Set("clientId", "hll")) should be (dimensions)
+
+    val estimates = aggregates.select(hll_cardinality(col("hll"))).collect()
+    estimates.foreach(x => x(0) should be (Submission.dimensions("clientId").size))
```
mreid-moz: If I understand correctly, this is checking that each combination of dimensions ends up containing the expected count.

Could we also add a test that combines across several dimensions and see if that also contains the expected count? For example, roll together all the `hll`s where `normalizedChannel` is `release` and ensure that it still matches `Submission.dimensions("clientId").size`

src/main/scala/views/KPI.scala
vitillo: I am planning to regenerate the whole report periodically.

PR #42, opened 2016-03-21
Add beta46-noapz experiment

Uberi:
Also some bug numbers to make cross-referencing easier.


mreid-moz:
Looks good to me, but r? @vitillo since he's familiar with the e10s experiments.

vitillo:
Looks good, thanks!

PR #43, opened 2016-03-22
Fix timezone edge case that was incorrect due to integer division

Uberi:
Utils was using integer division when it should use float division. As a result, the timezone 14:30, which is invalid, was truncated to timezone 14:00, which is valid.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1258683

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # telemetry-batch-view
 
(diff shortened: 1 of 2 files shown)


vitillo:
Tests are failing.

Uberi:
Seems like it's failing due to dependency issues. It doesn't seem to be related, so I'll send a separate PR for that later.

vitillo:
r+

PR #47, opened 2016-04-01
Add a "Main Summary" data set per Bug 1260847

mreid-moz:
This is intended to be the go-to data set for analysis of 100% of main pings.


vitillo:
@mreid-moz could you rebase this PR so that we have a cleaner history once we merge it?

mreid-moz:
Bah, git rebase fail. See #48 instead.

PR #48, opened 2016-04-01
Clean History: Add a "Main Summary" data set per Bug 1260847

mreid-moz:
This is intended to be the go-to data set for analysis of 100% of main pings. Same as #47 but with a cleaned up commit history.


src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+    lazy val system = parse(fields.getOrElse("environment.system", "{}").asInstanceOf[String])
+    lazy val info = parse(fields.getOrElse("payload.info", "{}").asInstanceOf[String])
+    lazy val histograms = parse(fields.getOrElse("payload.histograms", "{}").asInstanceOf[String])
+    lazy val keyedHistograms = parse(fields.getOrElse("payload.keyedHistograms", "{}").asInstanceOf[String])
+
+    lazy val weaveConfigured = Utils.booleanHistogramToBoolean(histograms \ "WEAVE_CONFIGURED")
+    lazy val weaveDesktop = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_DESKTOP")
+    lazy val weaveMobile = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
+
+    // TODO: confirm that it is safe to consider a wonky histogram as zero.
```
vitillo: Has this been addressed?

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+      }),
+      "is_default_browser" -> ((settings \ "isDefaultBrowser") match {
+        case JBool(x) => x
+        case _ => null
+      }),
+      "default_search_engine_data_name" -> ((settings \ "defaultSearchEngineData" \ "name") match {
+        case JString(x) => x
+        case _ => null
+      }),
+      "search_counts" -> Utils.getSearchCounts(keyedHistograms \ "SEARCH_COUNTS")
```
vitillo: Could we [LOOP_ACTIVITY_COUNTER](https://bugzilla.mozilla.org/show_bug.cgi?id=1261829) as well?

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+      "crashes_detected_content" -> hsum(keyedHistograms \ "SUBPROCESS_CRASHES_WITH_DUMP" \ "content"),
+      "crashes_detected_gmplugin" -> hsum(keyedHistograms \ "SUBPROCESS_CRASHES_WITH_DUMP" \ "gmplugin"),
+      "crash_submit_attempt_main" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_ATTEMPT" \ "main-crash"),
+      "crash_submit_attempt_content" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_ATTEMPT" \ "content-crash"),
+      "crash_submit_attempt_plugin" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_ATTEMPT" \ "plugin-crash"),
+      "crash_submit_success_main" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_SUCCESS" \ "main-crash"),
+      "crash_submit_success_content" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_SUCCESS" \ "content-crash"),
+      "crash_submit_success_plugin" -> hsum(keyedHistograms \ "PROCESS_CRASH_SUBMIT_SUCCESS" \ "plugin-crash"),
+      // End crash count fields
+      "active_addons_count" -> (Utils.countKeys(addons \ "activeAddons") match {
```
vitillo: Might be worthwhile to keep the full add-on list as that is something that comes in handy pretty often.

src/main/scala/streams/MainSummary.scala
vitillo: Can we move `getSearchCounts` in Utils.histograms?

src/main/scala/DerivedStream.scala
```diff
@@ -6,8 +6,7 @@ import com.typesafe.config._
 import java.io.File
 import java.util.UUID
 import org.apache.hadoop.fs.Path
-import org.apache.spark.{SparkConf, SparkContext}
```
vitillo: I am trying to move away from DerivedStream. Could you use a similar structure as the one used for `ClientCountView`?

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+
+      val groups = DerivedStream.groupBySize(summaries.collect().toIterator)
+      val summaryMessages = sc.parallelize(groups, groups.size)
+        .flatMap(x => x)
+        .flatMap { case obj =>
+          val hekaFile = bucket.getObject(obj.key).getOrElse(throw new Exception("File missing on S3: " + obj.key))
+          for (message <- HekaFrame.parse(hekaFile.getObjectContent(), hekaFile.getKey())) yield message
+        }
+        .flatMap { case message => messageToMap(message) }
+        .repartition(100) // TODO: partition by sampleId
```
vitillo: Please implement the partitioning by sampleId.

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+        val allRecords = for {
+          record <- partitionIterator
+          saveable = buildRecord(record, schema)
+        } yield saveable
+
+        var ignoredCount = 0
+        var processedCount = 0
+        val records = allRecords.map(r => r match {
+          case Some(_) =>
+            processedCount += 1
```
vitillo: You could also use an accumulator for this.

src/main/scala/streams/MainSummary.scala
mreid-moz: Let's include those later, when we have a specific need for them - they will significantly increase the data size, and may require some design for making them easy/performant to analyze.

vitillo: Does the size really matter with Parquet?

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+    for ((fieldName, fieldValue) <- fields) {
+      fieldValue match {
+        case Some(optValue) => optValue match {
+          case listValue: List[Map[String,Any]] => {
+            // Handle "list" type fields (namely search counts)
+            val items = scala.collection.mutable.ArrayBuffer.empty[GenericRecord]
+            val fieldSchema = schema.getField(fieldName).schema().getTypes().get(1).getElementType()
+            for (r <- listValue) {
+              r match {
+                case m: Map[String,Any] => {
```
vitillo: Why is this matching needed here? The compiler knows that `listValue` is a `List[Map[String,Any]]`.

src/main/scala/streams/MainSummary.scala
```diff
@@ -0,0 +1,395 @@
...
+                  val built = buildRecord(m, fieldSchema)
+                  for (b <- built)
```
vitillo: Could you merge this for loop with the one in a single for comprehension that returns an array?

src/main/scala/streams/MainSummary.scala
mreid-moz: The compiler was complaining about type erasure and being unable to enforce the specific `Map` types - I'll see if I can get it to compile without the redundant matcher.  The warning was:
```
[warn] /Users/mark/mozilla/github/telemetry-batch-view/src/main/scala/streams/MainSummary.scala:369: non-variable type argument Map[String,Any] in type pattern List[Map[String,Any]] is unchecked since it is eliminated by erasure
[warn]           case listValue: List[Map[String,Any]] => {
[warn]                           ^
```

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+    o match {
+      case JObject(x) => Some(x.length)
+      case _ => {
+        None
+      }
+    }
+  }
+
+  // Check if a json value contains a number greater than zero.
+  def gtZero(v: JValue): Boolean = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+
+  def getHistogramSum(h: JValue, default: Int): Int = {
+    (h \ "sum") match {
+      case x: JInt => x.num.toInt
+      case _ => default
+    }
+  }
+
+  // Count the number of keys inside a JSON Object
+  def countKeys(o: JValue): Option[Long] = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
     val millisecondsPerDay = 1000 * 60 * 60 * 24
     dateFormatter.withZone(org.joda.time.DateTimeZone.UTC).print(new DateTime(timestamp.toLong * millisecondsPerDay))
   }
+
+  // Find the largest numeric bucket that contains a value greater than zero.
+  def enumHistogramToCount(h: JValue): Option[Long] = {
```
vitillo: Please move this under `Utils.histogram`.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+      case _ => {
+        None
+      }
+    }
+  }
+
+
+  // Given histogram h, return true if it has a value in the "true" bucket,
+  // or false if it has a value in the "false" bucket, or None otherwise.
+  def booleanHistogramToBoolean(h: JValue): Option[Boolean] = {
```
vitillo: Please move this under `Utils.histogram`.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+    (gtZero(h \ "values" \ "1"), gtZero(h \ "values" \ "0")) match {
+      case (true, _) => Some(true)
+      case (_, true) => Some(false)
+      case _ => None
+    }
+  }
+
+  def getHistogramSum(h: JValue, default: Int): Int = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+
+  def toInt(s: String): Option[Int] = {
+    try {
+      Some(s.toInt)
+    } catch {
+      case e: Exception => None
+    }
+  }
+
+  def compareFlashVersions(a: Option[String], b: Option[String]): Option[Int] = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+        }
+
+        // They're the same.
+        return Some(0)
+      }
+      case _ => None
+    }
+  }
+
+  def maxFlashVersion(a: String, b: String): String = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+    val c = compareFlashVersions(Some(a), Some(b)).getOrElse(1)
+    if (c < 0)
+      b
+    else
+      a
+  }
+
+  // See also:
+  //  https://github.com/mozilla-services/data-pipeline/blob/master/hindsight/modules/fx/ping.lua#L82
+  def getFlashVersion(addons: JValue): Option[String] = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+    })
+
+    if (flashVersions.nonEmpty)
+      Some(flashVersions.reduceLeft(maxFlashVersion(_, _)))
+    else
+      None
+  }
+
+  val searchKeyPattern = "^(.+)\\.(.+)$".r
+  def searchHistogramToMap(name: String, hist: JValue): Option[Map[String, Any]] = {
```
vitillo: Please remove this function from `Utils` as it's not generally useful.

src/main/scala/utils/Utils.scala
```diff
@@ -76,4 +77,198 @@ object Utils{
...
+      }
+    })
+
+    if (flashVersions.nonEmpty)
+      Some(flashVersions.reduceLeft(maxFlashVersion(_, _)))
+    else
+      None
+  }
+
+  val searchKeyPattern = "^(.+)\\.(.+)$".r
```
vitillo: `searchKeyPattern` should not have public visibility.

src/test/scala/ChurnTest.scala
```diff
@@ -80,23 +79,21 @@ class ChurnTest extends FlatSpec with Matchers{
 }
 """
   "A boolean histogram" can "be converted to a boolean" in {
-    val churn = Churn("")
```
vitillo: Can we move this tests under telemetry.test.histogram?

src/test/scala/MainSummaryTest.scala
```diff
@@ -0,0 +1,413 @@
...
+    val schema = ms.buildSchema
+    val built = ms.buildRecord(testMap, schema)
+
+    built.isEmpty should be (false)
+  }
+
+  "MainSummary records" can "be serialized" in {
+    val ms = MainSummary("")
+    val schema = ms.buildSchema
+    val built = ms.buildRecord(testMap, schema)
```
vitillo: `testMap` shouldn't be defined manually but generated from the raw ping using messageToMap, which doesn't appear to be tested.

src/main/scala/streams/MainSummary.scala
mreid-moz: What kind of representation would you like to see for `LOOP_ACTIVITY_COUNTER`?

vitillo: `List[Int]` seems reasonable.

src/main/scala/streams/MainSummary.scala
mreid-moz: Is there a better alternative to this behaviour? It was kind of a toss-up here between zero, null, and some "special" flag-value like -1.  I'm still unsure what's best, though zero is clearly the most convenient for analysis.

src/main/scala/streams/MainSummary.scala
mreid-moz: Not too much, though it impacts the size of each object on S3, and might make us want to reconsider the number of partitions we're using (currently 100).

src/main/scala/streams/MainSummary.scala
mreid-moz: Is it OK to partition using a field that is nullable?

src/main/scala/DerivedStream.scala
mreid-moz: Yes, let's handle that as part of the next version.

src/main/scala/streams/MainSummary.scala
mreid-moz: > Can we move getSearchCounts in Utils.histograms?

Are you OK with me adding two more classes in the `telemetry.utils` package: `HistogramUtils` for `booleanHistogramToBoolean` etc, and another one `MainDocUtils` for things specific to pings of type `main` containing stuff like `getFlashVersion` and `getSearchCounts`?

vitillo: I am not convinced that we are going to need those functions in other derived datasets and if we do I would like to have a complete set of functions to deal with all histogram types. Why don't you create a namespace for your dataset and add those classes to it?

src/main/scala/streams/MainSummary.scala
vitillo: Yes, you can just map `null` to one of the partitions.

src/main/scala/streams/MainSummary.scala
vitillo: Imho `null` would be preferable to signal that no (meaningful) data is available as 0 has a different meaning.

mreid-moz: Ok, I'll change to use `null` for missing/bad sums

src/main/scala/streams/MainSummary.scala
mreid-moz: Per discussion on IRC, I've moved this dataset and its utils into a "main_summary" namespace.
Regarding `LOOP_ACTIVITY_COUNTER`, I will add separate fields for each enum value.

From Histograms.json: `"Number of times an action is performed by an user (0=OPEN_PANEL, 1=OPEN_CONVERSATION, 2=ROOM_OPEN, 3=ROOM_SHARE, 4=ROOM_DELETE)"`
So we will have fields for `loop_activity_open_panel`, `loop_activity_open_conversation`, etc.

src/main/scala/streams/MainSummary.scala
mreid-moz: Let's do this as part of the next round of changes too.

src/main/scala/streams/MainSummary.scala
mreid-moz: I'm going to rework this whole function as part of the move away from Avro, so I'll clean it up then

src/test/scala/MainSummaryTest.scala
mreid-moz: I've added another test for `messageToMap` that reads directly from a heka-framed file.

src/main/scala/streams/main_summary/MainSummary.scala
```diff
@@ -0,0 +1,424 @@
...
+
+    // Consider a wonky histogram (one for which the "sum" field
+    // is not a valid number) as null.
+    val hsum = (h: JValue) => h \ "sum" match {
+      case x: JInt => x.num.toInt
+      case _ => null
+    }
+
+    val hval = (h: JValue, b: String) => h \ "values" \ b match {
+      case x: JInt => x.num.toInt
```
vitillo: Isn't `x` already an Integer?

src/main/scala/streams/main_summary/MainSummary.scala
```diff
@@ -0,0 +1,424 @@
...
+    lazy val keyedHistograms = parse(fields.getOrElse("payload.keyedHistograms", "{}").asInstanceOf[String])
+
+    lazy val weaveConfigured = Utils.booleanHistogramToBoolean(histograms \ "WEAVE_CONFIGURED")
+    lazy val weaveDesktop = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_DESKTOP")
+    lazy val weaveMobile = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
+
+    // Consider a wonky histogram (one for which the "sum" field
+    // is not a valid number) as null.
+    val hsum = (h: JValue) => h \ "sum" match {
+      case x: JInt => x.num.toInt
```
vitillo: Isn't `x` already an Integer?

src/main/scala/streams/main_summary/MainSummary.scala
vitillo: What about `case JInt(x) => x` ?

src/main/scala/streams/main_summary/MainSummary.scala
```diff
@@ -0,0 +1,424 @@
...
+      "country" -> (fields.getOrElse("geoCountry", None) match {
+        case x: String => x
+        case _ => ""
+      }),
+      "city" -> (fields.getOrElse("geoCity", None) match {
+        case x: String => x
+        case _ => ""
+      }),
+      "profile_creation_date" -> ((profile \ "creationDate") match {
+        case x: JInt => x.num.toLong
```
vitillo: What about `case JInt(x) => x` ?

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -0,0 +1,192 @@
...
+    val c = compareFlashVersions(Some(a), Some(b)).getOrElse(1)
+    if (c < 0)
+      b
+    else
+      a
+  }
+
+  // See also:
+  //  https://github.com/mozilla-services/data-pipeline/blob/master/hindsight/modules/fx/ping.lua#L82
+  def getFlashVersion(addons: JValue): Option[String] = {
```
vitillo: This could be implemented with a [for comprehension](https://github.com/json4s/json4s#linq-style).

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -0,0 +1,192 @@
...
+        if (buf.isEmpty) None
+        else Some(buf.toList)
+      }
+      case _ => None
+    }
+  }
+
+
+  // Find the largest numeric bucket that contains a value greater than zero.
+  def enumHistogramToCount(h: JValue): Option[Long] = {
```
vitillo: What's the logic of this? Why are you considering only the largest numeric bucket?

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -0,0 +1,192 @@
...
+      }
+      case _ => {
+        None
+      }
+    }
+  }
+
+  // Given histogram h, return true if it has a value in the "true" bucket,
+  // or false if it has a value in the "false" bucket, or None otherwise.
+  def booleanHistogramToBoolean(h: JValue): Option[Boolean] = {
```
vitillo: A boolean histogram can have non-negative values in both the `true` and `false` bucket. It seems like you are considering it as a `flag` histogram here, which is different.

src/main/scala/streams/main_summary/MainSummary.scala
mreid-moz: that would give a `BigInt`, I think

src/main/scala/streams/main_summary/Utils.scala
mreid-moz: This is used to count the number of active sync devices.  It is based on this histogram (and the `MOBILE` variant):
```
"WEAVE_DEVICE_COUNT_DESKTOP": {
    "alert_emails": ["fx-team@mozilla.com"],
    "bug_numbers": [1232050],
    "expires_in_version": "never",
    "kind": "enumerated",
    "n_values": 10,
    "releaseChannelCollection": "opt-out",
    "description": "Number of desktop devices (including this device) associated with this Sync account. Recorded each time Sync successfully completes the 'clients' engine."
  },
```
The idea is that I don't care how many times this value was measured, nor the sum. I just want to know how many devices were active, so I pick the highest bucket containing a measurement.

So if the number of sync devices goes from zero to one to two and back to one in a subsession (and happened to be counted each time), I would expect to see something like:
```
"WEAVE_DEVICE_COUNT_DESKTOP": {
        "range": [
          1,
          10
        ],
        "bucket_count": 11,
        "histogram_type": 1,
        "values": {
          "0": 1,
          "1": 2,
          "2": 1,
          "3": 0
        },
        "sum": 4
      }
```

an I would want to extract a value of `2`, being the highest number of devices observed.

src/main/scala/streams/main_summary/Utils.scala
mreid-moz: Nice! That's way cleaner.

src/main/scala/streams/main_summary/Utils.scala
mreid-moz: That's fine - if it has positive values in both buckets, I want to return `true` for this field. It is measuring whether the underlying feature (sync) is configured. If it was both configured and not configured in a subsession, I still want to report that the feature was used.

vitillo:
Looks good, thanks. Could you please rebase your commits and a clear commit message?

mreid-moz:
All cleaned up and commit message updated.

vitillo:
Looks good, thanks!

PR #49, opened 2016-04-04
Add necessary fields for beta46 experiment 2

Uberi:
The second beta46 experiment changed the format; there's now a system addon that sets environment/settings/e10sCohort.

This means we need a new flag in E10SExperiment that can handle this correctly, or a new class for it. This PR is an initial implementation, it probably still needs some more tests.

### DIFF
--- a/src/main/scala/DerivedStream.scala
+++ b/src/main/scala/DerivedStream.scala
@@ -186,29 +186,9 @@ object DerivedStream {
           val churn = Churn("telemetry/4/main/Firefox")
           Some(options.getOrElse('fromDate, to), churn)
 
-        case "e10s-enabled-aurora-43" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1193089
-          val from = options.getOrElse('fromDate, "20151022")
-          val exp = E10sExperiment("e10s-enabled-aurora-20151020@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/aurora/43.0a2/")
-          Some(from, exp)
-
-        case "e10s-enabled-beta-44" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1229104
-          val from = options.getOrElse('fromDate, "20151214")
-          val exp = E10sExperiment("e10s-enabled-beta-20151214@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/beta/44.0/")
-          Some(from, exp)
-
-        case "e10s-enabled-beta-45-withaddons" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1241336
-          val from = options.getOrElse('fromDate, "20160129")
-          val exp = E10sExperiment("e10s-beta45-withaddons@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/beta/45.0/")
-          Some(from, exp)
-
-        case "e10s-enabled-beta-45-withoutaddons" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1244187
-          val from = options.getOrElse('fromDate, "20160211")
-          val exp = E10sExperiment("e10s-beta45-withoutaddons@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/beta/45.0/")
-          Some(from, exp)
-
-        case "e10s-beta46-noapz" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1253669
-          val from = options.getOrElse('fromDate, "20160309")
-          val exp = E10sExperiment("e10s-beta46-noapz@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/beta/46.0/")
+        case "E10SExperiment" =>
+          val from = options.getOrElse('fromDate, "20160321")
+          val exp = E10sExperiment("e10s-beta46-cohorts", "telemetry/4/saved_session/Firefox/beta/46.0/")
           Some(from, exp)
 
         case _ =>
--- a/src/main/scala/streams/E10sExperiment.scala
+++ b/src/main/scala/streams/E10sExperiment.scala
@@ -24,6 +24,15 @@ private class SampleIdPartitioner extends Partitioner{
   }
 }
 
+/*
+Useful links re: e10sCohort experiment design and structure:
+
+* An overview of the rationale: https://bugzilla.mozilla.org/show_bug.cgi?id=1251259
+* Original bug for e10sCohort: https://bugzilla.mozilla.org/show_bug.cgi?id=1249845
+* A brief description of e10sCohort's schema: https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/4
+* A later addition to the schema: https://bugzilla.mozilla.org/show_bug.cgi?id=1255013
+*/
+
 case class E10sExperiment(experimentId: String, prefix: String) extends DerivedStream {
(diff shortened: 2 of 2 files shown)


src/main/scala/streams/E10sExperiment.scala
```diff
@@ -24,7 +24,7 @@ private class SampleIdPartitioner extends Partitioner{
   }
 }
 
-case class E10sExperiment(experimentId: String, prefix: String) extends DerivedStream {
+case class E10sExperiment(experimentId: String, prefix: String, newE10SExperimentFormat: Boolean) extends DerivedStream {
```
vitillo: Let's just support the new experiment format and remove the past ones from `DerivedStream`.

src/main/scala/streams/E10sExperiment.scala
```diff
@@ -49,27 +49,38 @@ case class E10sExperiment(experimentId: String, prefix: String) extends DerivedS
...
-        (clientId, sampleId, id, branch) match {
-          case (Some(client: String), Some(sample: Double), JString(id), JString(branch)) if id == experimentId && status.isEmpty =>
-            List(((client, sample.toInt), fields))
-          case _ => Nil
+        if (newE10SExperimentFormat) { // new e10s experiment format, which uses environment/settings/e10sCohort value for e10s experiments
+          val settings = parse(fields.getOrElse("environment.settings", "{}").asInstanceOf[String])
+          val cohort = settings \ "e10sCohort"
+
+          (clientId, sampleId, cohort) match {
+            case (Some(client: String), Some(sample: Double), JString(cohort)) if cohort != "unknown" =>
```
vitillo: When is cohort == unknown?

src/main/scala/streams/E10sExperiment.scala
```diff
@@ -49,27 +49,38 @@ case class E10sExperiment(experimentId: String, prefix: String) extends DerivedS
...
+              List(((client, sample.toInt), fields))
+            case _ => Nil
+          }
+        } else { // old e10s experiment format, which uses environment/addons/activeExperiment values for e10s experiments
+          val addons = parse(fields.getOrElse("environment.addons", "{}").asInstanceOf[String])
+          val id = addons \ "activeExperiment" \ "id"
+          val branch = addons \ "activeExperiment" \ "branch"
+
+          // Ignore the first session of the experiment
+          val log = parse(fields.getOrElse("payload.log", "[]").asInstanceOf[String])
```
vitillo: The reason we were doing this check is that in the first session of the experiment an user might be flagged as having e10s on but effectively still have it disabled as it requires a restart. How is this handled with the new format?

src/main/scala/DerivedStream.scala
```diff
@@ -188,27 +188,32 @@ object DerivedStream {
 
         case "e10s-enabled-aurora-43" => // https://bugzilla.mozilla.org/show_bug.cgi?id=1193089
           val from = options.getOrElse('fromDate, "20151022")
-          val exp = E10sExperiment("e10s-enabled-aurora-20151020@experiments.mozilla.org", "telemetry/4/saved_session/Firefox/aurora/43.0a2/")
```
vitillo: Please remove the older experiments.

src/main/scala/streams/E10sExperiment.scala
Uberi: According to :felipe:

    16:58 <azhang> for experiments, we had a thing where on the first session, the user is in the test group, but since they didn't restart yet, e10s isn't actually enabled for that session
    16:59 <felipe> azhang: there's a possibility, but it's much smaller with the system add-on. We looked at the data and that happened to about 0.5% of users. If that's really important to you you can use the "e10sEnabled" flag to be 100% sure
    17:03 <felipe> azhang: same reason, the code to enable them ran too late in the session to get it enabled in that session.  Since the system add-on startup is a lot earlier than the previous experiment system, this has a smaller chance of happening
    17:03 <felipe> azhang: but later when we start whitelisting some add-on users for e10s that chance might increase again. So I suggest that you take e10sEnabled into account if that's simple enough

So I'll add a check for that in a bit.

src/main/scala/streams/E10sExperiment.scala
Uberi: The unknown cohort shouldn't occur usually, but it means something basically went wrong. Now that I think of it, this would still be useful information to have, so I'll remove that check.

Uberi:
So GitHub still seems to be a bit wonky, but I removed the old experiments and added a check for first session weirdness. Seems like Travis isn't updating either, so push hooks are probably down as well.

src/main/scala/streams/E10sExperiment.scala
vitillo: OK

src/main/scala/streams/E10sExperiment.scala
vitillo: OK

src/main/scala/streams/E10sExperiment.scala
```diff
@@ -49,25 +58,19 @@ case class E10sExperiment(experimentId: String, prefix: String) extends DerivedS
...
-          if status == "ACTIVATED"
-          JString(exp) <- entry(3)
-          if exp == experimentId
-        } yield status
-
-        (clientId, sampleId, id, branch) match {
-          case (Some(client: String), Some(sample: Double), JString(id), JString(branch)) if id == experimentId && status.isEmpty =>
+        val settings = parse(fields.getOrElse("environment.settings", "{}").asInstanceOf[String])
+
+        // the e10s cohort determines which group a user belongs to in the e10s experiment (as of beta46 experiment 2; previously, we used the Telemetry Experiments framework)
```
vitillo: Please limit comments to 80 characters.

src/main/scala/streams/E10sExperiment.scala
```diff
@@ -49,25 +58,19 @@ case class E10sExperiment(experimentId: String, prefix: String) extends DerivedS
...
+        val cohort = settings \ "e10sCohort"
+
+        // the first time the system addon is run, it is possible that the addon determines the user should have e10s enabled for the experiment,
+        // and sets the e10sCohort to "test", yet it is too late in the session to actually enable e10s
+        // the suggested solution is to check the value of environment/settings/e10sEnabled, and exclude those pings from consideration
+        val JBool(e10sEnabled) = settings \ "e10sEnabled"
+        val pingShouldBeExcluded = cohort == "test" && !e10sEnabled // the ping is in the "test" cohort, which means e10s should be enabled, yet e10s wasn't actually enabled
```
vitillo: Could you do the same check for the control group? Might come in handy for the future.

vitillo: `pingShouldBeExcluded` is never used in the code

Uberi:
Changes:

* Add control group check as well.
* Fix comments.

src/main/resources/application.conf
```diff
@@ -1,3 +1,3 @@
 app {
-    parquetBucket="telemetry-parquet"
+    parquetBucket="telemetry-test-bucket"
```
mreid-moz: Eep, please don't change this in "master"

vitillo:
Could you please squash all your commits into a single one that has a clear short title and a description of your changes?

vitillo:
Looks good, thanks!

PR #50, opened 2016-04-07
Fix cohort detection

Uberi:
The E10S experiment notebooks reject pings with mismatched e10s status, but we also want to do this at the derived stream level.

The previous PR had a bug where the test would never pass. This fixes it.


vitillo:
Looks good, thanks!

PR #51, opened 2016-04-11
Use spark-hyperloglog package.

vitillo:
Use the spark-hyperloglog package instead of a local implementation of the same functionality.


build.sbt
```diff
@@ -1,10 +1,16 @@
 javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint")
 
+sparkVersion := "1.6.0"
+
+sparkComponents ++= Seq("yarn", "sql", "hive")
```
mreid-moz: Where is `sparkComponents` used to set dependencies?

src/main/scala/views/ClientCountView.scala
```diff
@@ -19,7 +19,7 @@ class Conf(args: Array[String]) extends ScallopConf(args) {
 object ClientCountView {
   private val hllMerge = new HyperLogLogMerge
   private val base = List("normalizedChannel", "country", "version", "e10sEnabled", "e10sCohort")
-  private val selection = "hll_create(clientId) as clientId" :: "substr(subsessionStartDate, 0, 10) as activityDate" :: base
+  private val selection = "hll_create(clientId, 12) as clientId" :: "substr(subsessionStartDate, 0, 10) as activityDate" :: base
```
mreid-moz: Please add a comment for `12` - doesn't have to be as specific as before, but indicate that it determines the error rate.

vitillo: Done

build.sbt
vitillo: It's defined in `project/plugins.sbt`.

PR #56, opened 2016-04-12
Crash aggregates

Uberi:
Intended to have behaviour equivalent to https://github.com/mozilla/moz-crash-rate-aggregates, but faster.

It's also a little bit simpler, since the Spark Scala API is somewhat richer and more robust than PySpark.


src/main/scala/streams/Crash.scala
```diff
@@ -0,0 +1,276 @@
...
+      val (activityDate, dimensions) = (uniqueKey.head.asInstanceOf[String], uniqueKey.tail.asInstanceOf[List[Option[String]]])
+      val dimensionsMap: Map[String, String] = (dimensionNames, dimensions).zipped.flatMap((key, value) =>
+        (key, value) match { // remove dimensions that don't have values
+          case (key, Some(value)) => Some(key, value)
+          case (key, None) => None
+        }
+      ).toMap
+      val statsMap = (statsNames, stats).zipped.toMap
+
+      val schema = buildSchema()
```
vitillo: We are trying to move away from using Avro schemas in new datasets and use only the one provided by SparkSQL. Basically you would have to port the same approach you used in your Python version.

src/main/scala/streams/Crash.scala
```diff
@@ -0,0 +1,276 @@
...
+import telemetry.{DerivedStream, ObjectSummary}
+import telemetry.DerivedStream.s3
+import telemetry.avro
+import telemetry.heka.{HekaFrame, Message}
+import telemetry.histograms._
+import telemetry.parquet.ParquetFile
+import telemetry.utils.Utils
+import org.joda.time._
+
+case class Crash() extends DerivedStream {
```
vitillo: We want to move away from `DerivedStream` for new datasets. Have a look at `ClientCountView.scala` for an example. The plan is to have a method in Scala with a similar functionality to `get_records` provided in `python_moztelemetry` so that all jobs needing raw data can use that instead of depending on some functionality inherited from their parent class.

Uberi:
After a lot of debugging, I found out that Parquet-Avro 1.8.1 (the most recent version) entirely breaks `DataFrame.write.parquet`:

    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.8.1",

Basically, if we even install parquet-avro, writing Parquet files from dataframes no longer works:

    java.lang.NoSuchMethodError: org.apache.parquet.schema.Types$MessageTypeBuilder.addFields([Lorg/apache/parquet/schema/Type;)Lorg/apache/parquet/schema/Types$GroupBuilder;
    (...16 more lines...)

There's a [resolved issue](https://issues.apache.org/jira/browse/HIVE-12828) in Parquet for this, but it's only for 1.8.2-SNAPSHOT, so it isn't really considered stable yet.

We should (1) move this into a separate project (so we don't need to depend on parquet-avro at all), or (2) move away from the Avro stuff all at once.

src/main/scala/views/CrashAggregateView.scala
```diff
@@ -0,0 +1,335 @@
...
+
+object CrashAggregateView {
+  class Conf(args: Array[String]) extends ScallopConf(args) {
+    val from = opt[String]("from", descr = "From submission date", required = false)
+    val to = opt[String]("to", descr = "To submission date", required = false)
+    verify()
+  }
+
+  def main(args: Array[String]) {
+    // load configuration for the
```
vitillo: Comment is incomplete.

src/main/scala/views/CrashAggregateView.scala
```diff
@@ -0,0 +1,335 @@
...
+    val to = conf.to.get match {
+      case Some(t) => fmt.parseDateTime(t)
+      case _ => DateTime.now.minusDays(1)
+    }
+    val from = conf.from.get match {
+      case Some(f) => fmt.parseDateTime(f)
+      case _ => DateTime.now.minusDays(1)
+    }
+
+    // set up the Spark, Hadoop, and other configurations
```
vitillo: Please rename to `// set up Spark `.

src/main/scala/views/CrashAggregateView.scala
```diff
@@ -0,0 +1,335 @@
...
+      activityDateRaw
+    }
+    val activityDateString = format.DateTimeFormat.forPattern("yyyy-MM-dd").print(activityDate) // format activity date as YYYY-MM-DD
+
+    // obtain the unique key of the aggregate that this ping belongs to
+    val uniqueKey = activityDateString :: (
+      for (path <- comparableDimensions) yield {
+        if (path.head == null) {
+          pingFields.get(path.tail.head) match {
+            case Some(topLevelField: String) => Some(topLevelField)
```
vitillo: Could we use a List with a single element to identify top-level fields?

src/main/scala/views/CrashAggregateView.scala
```diff
@@ -0,0 +1,335 @@
...
+    buildId match {
+      case Some(value: String) if value.matches("\\d{14}") => null
+      case _ => return None
+    }
+
+    // obtain the relevant stats for the ping
+    val usageHours: Double = info \ "subsessionLength" match {
+      case JInt(subsessionLength) =>
+        Math.min(25, Math.max(0, subsessionLength.toDouble / 3600))
+      case JNothing => 0
```
vitillo: Maybe we should ignore the submission in this case as well instead of returning 0?

src/main/scala/views/CrashAggregateView.scala
```diff
@@ -0,0 +1,335 @@
...
+      case _ => return None
+    }
+    val mainCrashes: Double = pingFields.get("docType") match {
+      case Some("main") => 0
+      case Some("crash") => 1
+      case _ => return None
+    }
+    val contentCrashes: Double = try {
+      getCountHistogramValue(keyedHistograms \ "SUBPROCESS_CRASHES_WITH_DUMP" \ "content")
+    } catch { case _: Throwable => 0 }
```
vitillo: Can we move the exception handling in getCountHistogramValue?

src/test/scala/CrashAggregateViewTest.scala
```diff
@@ -0,0 +1,195 @@
...
+        crashProcessedAccumulator, crashIgnoredAccumulator
+      ) = CrashAggregateView.compareCrashes(sc.get, sc.get.parallelize(pings))
+      val schema = CrashAggregateView.buildSchema()
+      val records = sqlContext.get.createDataFrame(rowRDD, schema)
+      records.count() // Spark is pretty lazy; kick it so it'll update our accumulators properly
+    }
+  }
+
+  "Records" must "have the correct lengths" in {
+    assert(fixture.records.count() == fixture.pings.length / 2)
```
vitillo: Could you write a comment that mentions why this assert has to be true?

src/main/scala/views/CrashAggregateView.scala
Uberi: Crash pings don't have this field, so we do need to allow pings without subsession lengths in some cases (we could validate the doc types, though).

Uberi:
Changed:

* Added comment about ping aggregates.
* Clean up code a bit.
* Use list with single element to identify top-level fields.

src/main/scala/views/CrashAggregateView.scala
vitillo: Right, there are two issues here. The first issue is that we should treat main and crash pings differently, i.e. if a main ping doesn't have a subsession length it should be ignored while the subsession length of a crash ping should be set to 0.

The other problem I see here is that we are treating crash pings as main pings even though they don't have a subsession length. By setting their subsession length to 0 it will skew the mean and the variance, unless we avoid incrementing the number of pings seen. We could have a separate ping count that counts just the number of crash pings seen.

vitillo: I am assuming here that each crash ping has an associated main ping with a "aborted-session" reason, which is where we are actually counting the ping to which the crash belongs to. Is that correct?

Uberi: Since aborted-session pings are being overreported, I don't think we can say a crash ping belongs to any particular main ping. Maybe we should exclude aborted-sessions?

I also just pushed changes that do the ping count change you mentioned. Now, `ping_count` measures only the number of main pings. I dunno if this will help with our variance though, since the crash pings make up a relatively tiny fraction of the total ones.

vitillo:
This looks good, did you compare the output generated by this job with the one generated by the Python one?

vitillo:
Could you please also add some documentation about the metrics collected in this job similarly to what you have written down for the Python one?

Uberi:
Changes:

* Add some better docs.
* Test the scala aggregator against the crash aggregator on nightly, aurora, and beta:
    * Scala output: https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=telemetry-test-bucket&prefix=crash_aggregates/v1/submission_date%3D2016-04-10/
    * In the Scala one, there's 10 files since it was set to `coalesce(10)`. In this PR it's set back to 1.
    * Python output: https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=telemetry-test-bucket&prefix=crash_aggregates/vPy/submission_date%3D2016-04-10/
* The differences can be explained by the Scala version being slightly more strict: it doesn't allow crash pings that have subsession lengths or main pings with no subsession lengths. In Python these are still processed.
* Upload path for parquet files updated to match the Python one.

vitillo:
Looks good, thanks!

PR #57, opened 2016-04-12
Add build field

Uberi:
This is needed in order to split populations based on architecture and other build-rated values.


vitillo:
Looks good, thanks!

PR #60, opened 2016-04-26
Update e10s experiment for beta47

Uberi:
e10sCohort is part of Beta now, so the experiment should basically start as soon as it's released. Today's merge day!


vitillo:
Looks good, we should wait a week before generating a dataset for it though.

PR #61, opened 2016-04-27
Add more metrics to client_count dataset.

vitillo:


mreid-moz:
This looks good to me.  Do we need to communicate the modified column names before we land / deploy?

vitillo:
I don't think that's necessary as I wrote all the dashboards that use that dataset.

mreid-moz:
Good :)

PR #62, opened 2016-04-29
Streams to views

Uberi:
Based on #56:

* Convert the Churn and Executive streams to views rather than the old `DerivedStream` kind. When all of them are completely converted, we can remove the Avro stuff.
* The executive view avoids a ton of shuffling by using `getRecords` with the channel name directly. This seems to make it a lot faster.
* `telemetry.utils.Telemetry` contains some useful functions, `getRecords` and `listOptions`.


src/main/scala/utils/Telemetry.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  def getRecords(sc: SparkContext, submissionDate: DateTime, pingPath: List[String]): RDD[Map[String, Any]] = {
+    // obtain the prefix of the telemetry data source
+    val metadataBucket = Bucket("net-mozaws-prod-us-west-2-pipeline-metadata")
+    val Some(sourcesObj) = metadataBucket.get("sources.json")
+    val metaSources = parse(Source.fromInputStream(sourcesObj.getObjectContent()).getLines().mkString("\n"))
+    val JString(telemetryPrefix) = metaSources \\ "telemetry" \\ "prefix"
+
+    // get a stream of object summaries that match the desired criteria
+    val bucket = Bucket("net-mozaws-prod-us-west-2-pipeline-data")
```
mreid-moz: You should get the bucket from the metadata as well: `val JString(bucket) = metaSources \\ "telemetry" \\ "bucket"`

Uberi:
Updated to use data bucket from metadata.

vitillo:
Thanks for picking this up @Uberi! Could you please rebase it?

src/main/scala/utils/Telemetry.scala
```diff
@@ -0,0 +1,85 @@
...
+      seenPrefixes
+    } else {
+      val matching = seenPrefixes
+        .flatMap(prefix => listS3Keys(bucket, prefix))
+        .filter(prefix => (pattern.head == "*" || prefix.endsWith(pattern.head + "/")))
+      matchingPrefixes(bucket, matching, pattern.tail)
+    }
+  }
+
+  def appendToFile(p: String, s: String): Unit = {
```
vitillo: Where is this function used?

src/main/scala/utils/Telemetry.scala
```diff
@@ -0,0 +1,85 @@
...
+    val pw = new java.io.PrintWriter(new java.io.FileOutputStream(new java.io.File(p),true))
+    try pw.write(s) finally pw.close()
+  }
+
+  def getRecords(sc: SparkContext, submissionDate: DateTime, pingPath: List[String]): RDD[Map[String, Any]] = {
```
vitillo: Is there any specific reason to use a List and not a '/' delimited string?

src/main/scala/views/ExecutiveView.scala
```diff
@@ -0,0 +1,205 @@
+package telemetry.views
```
vitillo: `ExecutiveStream.scala` should be removed.

src/main/scala/views/ExecutiveView.scala
```diff
@@ -0,0 +1,205 @@
...
+    val knownChannels = Set("nightly", "aurora", "beta", "release")
+    for (offset <- 0 to Days.daysBetween(from, to).getDays()) {
+      val currentDate = from.plusDays(offset)
+      val currentDateString = currentDate.toString("yyyyMMdd")
+
+      val schema = buildSchema()
+
+      // get a list of channel values that fall under "Other"
+      // for example, "esr" and "OTHER" should both be considered part of "Other" rather than
+      // having their own categories, while "nightly" and other known channels should have
```
vitillo: Are we doing this somewhere else as well?

vitillo:
@mreid-moz are you happy with the churn rewrite?

mreid-moz:
Taking a look now

docs/CrashAggregateView.md
```diff
@@ -0,0 +1,126 @@
...
+The crash aggregate view is generated by `src/main/scala/views/CrashAggregateView.scala`. It contains stats used for computing crash rates and other stability-related metrics, aggregated by a variety of criteria.
+
+Essentially, each row of the view represents the stats for a particular population. For example, one row might represent crash stats and usage hours for pings from Firefox Beta 46 on 64-bit Windows 7, and so on.
+
+To generate the dataset for April 10, 2016 to April 11, 2016:
+```bash
+sbt "run-main telemetry.views.CrashAggregateView --from 20160410 --to 20160411"
+```
+
+**Note:** Currently, due to [avro-parquet issues](https://issues.apache.org/jira/browse/HIVE-12828), Parquet writing only works under the `spark-submit` commands - the above example will fail. This will be fixed when avro-parquet updates or is removed.
```
mreid-moz: s/Currently/As of 2016-05-03/

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
+package telemetry.streams
+
+import awscala.s3.{S3, Bucket}
```
mreid-moz: This import is unused

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+import com.typesafe.config._
+import org.apache.spark.SparkContext
```
mreid-moz: SparkContext is imported again on the next line, this one can be removed

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.spark.{SparkConf, SparkContext, Accumulator}
```
mreid-moz: `Accumulator` appears to be unused

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.spark.sql.{Row, SQLContext, SaveMode}
+import org.apache.spark.sql.types._
+import org.json4s._
+import org.json4s.jackson.JsonMethods.parse
+import telemetry.DerivedStream.s3
```
mreid-moz: unused

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+    lazy val settings = parse(message.getOrElse("environment.settings", "{}").asInstanceOf[String])
+    lazy val info = parse(message.getOrElse("payload.info", "{}").asInstanceOf[String])
+    lazy val histograms = parse(message.getOrElse("payload.histograms", "{}").asInstanceOf[String])
+
+    lazy val weaveConfigured = Utils.booleanHistogramToBoolean(histograms \ "WEAVE_CONFIGURED")
+    lazy val weaveDesktop = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_DESKTOP")
+    lazy val weaveMobile = Utils.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
+
+    val row = Row( // the entries in this correspond to the schema in `buildSchema()`
+      message.getOrElse("clientId", None) match {
```
mreid-moz: The ordering of fields in `messageToRow` does not match the ordering of fields in `buildSchema`. See `submissionDate` for example. This wasn't a problem when we built an intermediate map, but now the ordering should match.

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+    val parquetBucket = appConf.getString("app.parquetBucket")
+
+    for (offset <- 0 to Days.daysBetween(from, to).getDays()) {
+      val currentDate = from.plusDays(offset)
+      val currentDateString = currentDate.toString("yyyyMMdd")
+
+      val schema = buildSchema()
+      val messages = Telemetry.getRecords(sc, currentDate, List("telemetry", "4", "main", "Firefox"))
+      val rowRDD = messages.flatMap(messageToRow).repartition(100) // TODO: partition by sampleId
+      val records = sqlContext.createDataFrame(rowRDD.coalesce(1), schema)
```
mreid-moz: Do we need the `coalesce(1)` here?  Shouldn't we leave `rowRDD` as-is since it already has the desired number of partitions (in this case 100)?

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+      records.write.mode(SaveMode.Overwrite).parquet(s"s3://$parquetBucket/churn/v1/submission_date_s3=$currentDateString")
```
mreid-moz: Rather than overwriting, would it be better to delete the `submission_date_s3=$currentDateString` prefix before saving?

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
+package telemetry.streams
```
mreid-moz: Should move to `package telemetry.views`

src/main/scala/utils/Telemetry.scala
Uberi: It's somewhat simpler since we don't need to put another wrapper function around this one, and also lets us concat paths together using the usual list operations, which I think we do in a few places. It would be quite simple to switch over, though.

src/main/scala/views/ExecutiveView.scala
Uberi: This is the only instance of bucketing into OTHER I've seen so far - the other streams/views just keep all the channel info as-is.

src/main/scala/views/ChurnView.scala
Uberi: If I'm reading [the docs](https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/sql/DataFrameWriter.html) right, deleting it before saving would be exactly equivalent to using `SaveMode.Overwrite`. Would there be any differences in our case if we delete first rather than overwriting?

Uberi:
Changes:

* Implement all of the above feedback.
* Note: Since I'm not in the AWS cloud services group, I haven't tested these on real data.

src/main/scala/views/ExecutiveView.scala
vitillo: We should try to be backward compatible here and not e.g. rename the esr channel to OTHER, unless there is a very good reason to do so. It would also be helpful to mention why this code is needed in a comment (i.e. for perf reasons).

vitillo: Can we add a comment mentioning why this is necessary (i.e. for perf reasons)?

src/main/scala/views/ChurnView.scala
mreid-moz: I wasn't sure if overwrite would just replace files with the same name, which would cause all sorts of trouble. After reading the link, it looks like it does the right thing. Thanks!

mreid-moz:
Changes look good to me. Can you test it on real data @vitillo?

src/main/scala/views/ExecutiveView.scala
Uberi: I'm not actually sure why it's necessary (maybe it was part of the original stream requirements?). IIRC, this is how the existing executive stream data on S3 is currently laid out. Is it different from this now?

vitillo: Right, that makes sense, thanks.

src/main/scala/utils/Telemetry.scala
```diff
@@ -0,0 +1,80 @@
...
+    val JString(telemetryPrefix) = metaSources \\ "telemetry" \\ "prefix"
+    val JString(dataBucket) = metaSources \\ "telemetry" \\ "bucket"
+
+    // get a stream of object summaries that match the desired criteria
+    val bucket = Bucket(dataBucket)
+    val keys = matchingPrefixes(
+      bucket,
+      List("").toStream,
+      List(telemetryPrefix, submissionDate.toString("yyyyMMdd")) ++ pingPath
+    ).flatMap(prefix => s3.objectSummaries(bucket, prefix)).map(summary => summary.getKey())
```
mreid-moz: Should we include something like the logic in `DerivedStream.groupBySize` to try to balance out the size of each partition / task?

vitillo:
@mreid-moz I can test the executive view, can you take churn?

mreid-moz:
@vitillo Yeah, I'll test out the churn view today.

mreid-moz:
Ok, I ran a test of the `ChurnView` code and found the following for the data on 20160401 (running on a cluster of 20 nodes):

**Test 1: DerivedStream-based code:**
```bash
time spark-submit \
    --master yarn-client \
    --class telemetry.DerivedStream \
    telemetry-batch-view-1.1-orig.jar \
    --from-date 20160401 \
    --to-date 20160401 \
    Churn
```

**Test 2: ChurnView code:**
```bash
time spark-submit \
    --master yarn-client \
    --class telemetry.views.ChurnView \
    telemetry-batch-view-1.1-views.jar \
    --from 20160401 \
    --to 20160401
```


- The old DerivedStream-based code ran in 26m31.701s and generated a correct-looking number of records.
- The View-based code ran in 137m48.952s (about 5 times slower), and generated a dataset with the correct structure, but no actual records (a count of the dataframe was zero).
- The tasks in the ChurnView code were very unbalanced time-wise, with the longest task taking 2.2hr. See this screenshot from the Spark UI:
screen shot 2016-05-12 at 3 06 47 pm
- I left the generated datasets in telemetry-test-bucket/churn/v1 (DerivedStream based) and telemetry-test-bucket/churn/v2 (ChurnView) for further inspection, and I saved the "spark.log" for the test if that's of interest.
- I also have a simple notebook I used to count the resulting datasets if needed.

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+        case _ => null
+      }
+    )
+    Some(row)
+  }
+
+  def buildSchema(): StructType = {
+    StructType(
+      StructField("clientId",            StringType,  false) ::
+      StructField("sampleId",            IntegerType, false) ::
```
mreid-moz: Either s/IntegerType/LongType/ here or output an Integer value in `messageToRow`.  I noticed that Spark types are less forgiving of Int vs. Long conversion than the Avro stuff.

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+      StructField("channel",             StringType,  false) :: // appUpdateChannel
+      StructField("normalizedChannel",   StringType,  false) :: // normalizedChannel
+      StructField("country",             StringType,  false) :: // geoCountry
+      StructField("profileCreationDate", IntegerType, true) :: // environment/profile/creationDate
```
mreid-moz: This one should be a `LongType` as well

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+      StructField("subsessionStartDate", StringType,  true) :: // info/subsessionStartDate
+      StructField("subsessionLength",    IntegerType, true) :: // info/subsessionLength
```
mreid-moz: Ditto

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+      StructField("distributionId",      StringType,  true) :: // environment/partner/distributionId
+      StructField("submissionDate",      StringType,  false) ::
+
+      // bug 1232050
+      StructField("syncConfigured",      BooleanType, true) :: // WEAVE_CONFIGURED
+      StructField("syncCountDesktop",    IntegerType, true) :: // WEAVE_DEVICE_COUNT_DESKTOP
```
mreid-moz: Ditto

src/main/scala/views/ChurnView.scala
```diff
@@ -0,0 +1,168 @@
...
+      StructField("syncCountMobile",     IntegerType, true) :: // WEAVE_DEVICE_COUNT_MOBILE
```
mreid-moz: Ditto

PR #67, opened 2016-05-12
Spark schema main summary to view

mreid-moz:
Per bug 1272334, refactor the Main Summary code to use Spark SQL types directly (rather than using Avro Schema), and switch from the "Derived Stream" approach to the "Views" approach.

This builds on the code in #62, but still has some performance problems to work out.

Processing a single day's data on a 20-node cluster (April 1, 2016 was my reference date) takes a bit more than 90 minutes with this current code.  Using the DerivedStreams approach from a few commits back takes about half that amount of time.

Note that I also added a step to `Telemetry.getMessages` to balance the partition sizes, which reduced the run time from more than 3 hours down to 90 minutes or so.

@vitillo @Uberi Feedback and suggestions appreciated, particularly about what I can do to identify and fix the performance difference.


vitillo:
Please rebase.

mreid-moz:
Merged master.  I'll squash all the WIP commits once this is ready to land :)

build.sbt
```diff
@@ -14,7 +14,7 @@ lazy val root = (project in file(".")).
     retrieveManaged := true,
     ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) },
     libraryDependencies += "org.apache.avro" % "avro" % "1.7.7",
-    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.8.1",
+    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.7.0",
```
mreid-moz: parquet-avro 1.8.1 conflicted with writing a DataFrame to Parquet format.

src/test/scala/ClientCountViewTest.scala
```diff
@@ -75,31 +75,34 @@ class ClientCountViewTest extends FlatSpec with Matchers{
     val sparkConf = new SparkConf().setAppName("KPI")
     sparkConf.setMaster(sparkConf.get("spark.master", "local[1]"))
     val sc = new SparkContext(sparkConf)
-    val sqlContext = new HiveContext(sc)
-    sqlContext.udf.register("hll_create", hllCreate _)
-    sqlContext.udf.register("hll_cardinality", hllCardinality _)
-    import sqlContext.implicits._
+    sc.setLogLevel("WARN")
+    try {
```
mreid-moz: Note: without this try/finally block, a failure anywhere in this test suite would leave the SparkContext around, causing all the other Spark-related tests to fail too.

src/test/scala/Longitudinal.scala
```diff
@@ -176,7 +176,8 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
   }
 
   "Records" can "be serialized" in {
-    ParquetFile.serialize(List(fixture.record).toIterator, fixture.schema)
+    println("TODO: fix Avro+Parquet Serialization test")
```
mreid-moz: Ideally we can just get rid of this by changing `Longitudinal` to use the SparkSQL / DataFrame serialization too.

src/main/scala/utils/Telemetry.scala
```diff
@@ -0,0 +1,93 @@
...
+      matchingPrefixes(bucket, matching, pattern.tail)
+    }
+  }
+
+  def appendToFile(p: String, s: String): Unit = {
+    val pw = new java.io.PrintWriter(new java.io.FileOutputStream(new java.io.File(p),true))
+    try pw.write(s) finally pw.close()
+  }
+
+  def getRecords(sc: SparkContext, submissionDate: DateTime, pingPath: List[String]): RDD[Map[String, Any]] = {
```
vitillo: Why is this function defined here but never used?

src/main/scala/views/MainSummaryView.scala
```diff
@@ -0,0 +1,396 @@
...
+        messageToRow(m) match {
+          case None =>
+            ignoredCount += 1
+            None
+          case x =>
+            processedCount += 1
+            x
+        }
+      })
+      // TODO: partition by sampleId
```
vitillo: Please implement the `sampleId` based partitioning and remove the `coalesce` operation.

vitillo:
Just fyi, I am planning to write a proper telemetry API in the next weeks which is going to provide fast access to telemetry records. @Uberi `getRecords` was just a temporary measure until we have that API.

vitillo:
Where can I find the output of this job?

build.sbt
vitillo: @Uberi that might fix the bug you had seen with the crash dataset.

src/main/scala/utils/Telemetry.scala
mreid-moz: I copied this whole file from Anthony's PR, which only had the `getRecords` function - I added `getMessages` because I needed to access the main Heka fields (Timestamp specifically).

mreid-moz:
You can find the sample data at `telmetry-test-bucket/main_summary/v3`

vitillo:
Do you know why `s3://telemetry-test-bucket/main_summary/v3/submission_date_s3=20160401/` has 103 files (instead of 100) and why it's 1/3 smaller than `s3://telemetry-parquet/main_summary/v2/submission_date_s3=20160401/`?

vitillo:
Also note that v2 has practically no skew and all files have the same size while the same isn't true for v3.

mreid-moz:
The extra 3 files in `s3://telemetry-test-bucket/main_summary/v3/submission_date_s3=20160401/` are the "_SUCCESS", "_metadata" and "_common_metadata" files that are autogenerated by the dataframe serialization. I think it's smaller due to using gzip compression instead of none (or snappy, I'm not sure) used by the Avro/Parquet stuff.

mreid-moz:
Re: skew, I believe that is due to actually repartitioning the records (by sampleId) in v2, whereas the current iteration of v3 just does the "coalesce" which combines existing partitions.

src/main/scala/utils/Telemetry.scala
Uberi: This is a debugging function that I left in accidentally. It's pretty useful when debugging locally, but we probably don't need to keep it around.

build.sbt
Uberi: It does seem like it. Are there any issues with downgrading though? 7c7494ca42eddb81f25248e5419d2381e738bad8 seems to have upgraded for reasons related to functionality.

mreid-moz: I think 1.8.1 added support for writing some of the data structures used by the Longutindal dataset (Map of structs or something, I'm not sure). You can see the errors I got by checking out out this PR and re-enabling the serialization test in ` src/test/scala/Longitudinal.scala`, then running `sbt test`

src/main/scala/views/MainSummaryView.scala
mreid-moz: I've updated this code to use "real" partitioning, just running a test now.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -0,0 +1,399 @@
...
+    // Set up Spark
+    val sparkConf = new SparkConf().setAppName(jobName)
+    sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
+    val sc = new SparkContext(sparkConf)
+    val sqlContext = new SQLContext(sc)
+    val hadoopConf = sc.hadoopConfiguration
+    hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
+
+    // We want to end up with reasonably large parquet files on S3.
+    hadoopConf.setInt("parquet.block.size", 128 * 1024 * 1024)
```
mreid-moz: @vitillo Is 128MB a reasonable size for parquet files? Should we go larger? I've read conflicting advice.

vitillo: Depends, how many files are generated per day (with 128MB as size) and how many days are usually read from an analysis job?

mreid-moz: There are about 20-30GB per day of data (in the latest gzipped-parquet form per this PR). If we partition by sample_id, we should have 100 partitions of ~235MB each. So likely two files per `day+sample_id` partition.  Analysis jobs frequently look at MAU, so using 30 days of data is common (40 days, including data latency for activity date vs. submission date).

vitillo: We should have a single file per day+sample_id then.

mreid-moz: Ok, I'll go with 256MB for now.

vitillo: Let's make it 512 to avoid very small files to be generated.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -0,0 +1,408 @@
...
+    sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
+    val sc = new SparkContext(sparkConf)
+    val sqlContext = new SQLContext(sc)
+    val hadoopConf = sc.hadoopConfiguration
+    hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
+
+    // We want to end up with reasonably large parquet files on S3.
+    val parquetSize = 512 * 1024 * 1024
+    hadoopConf.setInt("parquet.block.size", parquetSize)
+    hadoopConf.setInt("dfs.blocksize", parquetSize)
```
mreid-moz: @vitillo do you think we still need these block size parameters with the fix to properly partition by sample_id?

vitillo: Yes, the default is 128M I think.

mreid-moz:
Ok, I tested that the data can be read from Spark, and that the record counts match exactly for one day's data between the existing v2 data, the v3 data using manual partitioning, and the v4 data using `partitionBy("submission_date", "sample_id")`.

vitillo:
r+ once you confirm that the dataset can be read by Presto as well.

mreid-moz:
I checked, and the v3 and v4 datasets can't be auto-imported into Hive with the current `parquet2hive` utility. I filed [this pr](https://github.com/vitillo/parquet2hive/pull/3) with a fix (and tested that I can access the resulting table in Presto via sql.tmo afterwards.

One other issue I ran into is that `parquet2hive` assumes all S3 [partitioning fields are strings], so using the `sample_id` field in queries is somewhat awkward, since you can't easily use ranges of `sample_id`s.

I manually edited the resulting Hive commands to make `sample_id` a `BIGINT` type field, and that worked, but I'm not sure if there's a good general approach to this.

vitillo:
Casting sample_id to an integer in the query might be good enough.

mreid-moz:
Ok, there's a workaround where you can just cast the partition field to a number in your query if you want to treat it as one. Null values work as expected too. These queries all do the expected thing:
```sql
SELECT count(*) FROM test_main_summary_v4 WHERE sample_id = '5'
SELECT count(*) FROM test_main_summary_v4 WHERE cast(sample_id AS bigint) < 10
SELECT count(*) FROM test_main_summary_v4 WHERE sample_id IS NULL
SELECT count(*) FROM test_main_summary_v4 WHERE cast(sample_id AS bigint) IS NULL
```

mreid-moz:
@vitillo I've cleaned up the history in this branch, it should be ready for final review / merge.

mreid-moz:
Argh, I noticed a problem where the "overwrite" deletes the entire "version" prefix before adding data.

vitillo:
As discussed on IRC we should make sure the longitudinal dataset is not affected by the changes.

mreid-moz:
Ok, I've re-enabled the serialization test for `Longitudinal`, instead disabling the serialization test for `MainSummaryView`.

I confirmed that the MainSummaryView generation itself still works fine.

mreid-moz:
This also restores the "submission_date_s3" field, so that's one less change for consumers of the data between v2 and v3.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -91,13 +91,27 @@ object MainSummaryView {
...
+      // Note we cannot just use 'partitionBy' below to automatically populate
+      // the submission_date partition, because none of the write modes do
+      // quite what we want:
+      //  - "overwrite" causes the entire vX partition to be deleted and replaced with
+      //    the current day's data, so doesn't work with incremental jobs
+      //  - "append" would allow us to generate duplicate data for the same day, so
+      //    we would need to add some manual checks before running
+      //  - "error" (the default) causes the job to fail after any data is
+      //    loaded, so we can't do single day incremental updates.
+      //  - "ignore" causes new data not to be saved.
```
vitillo: You forgot to mention `write` in your comments.

src/test/scala/Longitudinal.scala
```diff
@@ -176,6 +176,7 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester{
   }
 
   "Records" can "be serialized" in {
+    println("TODO: fix Avro+Parquet Serialization test")
```
vitillo: Please remove this line.

vitillo:
@mreid-moz this can be merged; we should probably do that when you are back so we can deploy it shortly afterwards.

mreid-moz:
Summary of where we ended up with this PR:
- Performance is known to be sub-optimal for now.  The original `main_summary` code (before this change) took about 30 minutes to process a day's data. The code that used the `DerivedStreams` approach, but changed to SparkSQL types instead of Avro took about 45 minutes. The current code using the `Views` approach takes about 90 minutes. There are plans to rewrite the S3-iterating code, so we will tackle the performance problem at that time.
- The `MainSummaryView` test coverage does not include testing of the data serialization due to an incompatibility between versions of the `parquet-avro` library.
- The `submission_date_s3` field **is** still present in v3. We also introduce an S3 partition for `sample_id`, which is a string field, but can efficiently be cast to a number when sampling ranges are desired.
- The `src/main/scala/utils/Telemetry.scala` file was copied (with minor modifications) from #62, so one or the other of these PRs will need to be rebased.

If this sounds acceptable, then we're ready to merge.

vitillo:
r+

PR #70, opened 2016-05-24
Use the Telemetry lib instead of getRecords()

Uberi:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1275025


vitillo:
r+

vitillo:
@Uberi I reverted this change as I noticed a possible bug only later and I assume this had been tested on real data. I think what you want is `Telemetry.getMessages` and not `Telemetry.getRecords`.

vitillo:
@Uberi could you file a new PR and mention if change has been tested with real data? Thanks.

Uberi:
Sure, I'll be able to give it a test after https://bugzilla.mozilla.org/show_bug.cgi?id=1275986 is resolved.

PR #72, opened 2016-05-30
Add support for type unions in schemas, and represent settings with these unions

Uberi:
* We currently have support for `optional()` in Avro schemas, but not `nullable()` or `unionOf()`.
* Add support for all of these structures, as well as any others that are built on top of unions.
* Add support for non-string values in `settings.user_prefs`, plus tests for that functionality. Now, setting values can be integers/longs, doubles, and booleans. I think it covers all of the relevant cases; are there any I'm missing?
* Parquet output looks good for sample data. It should accept a superset of the pings it currently accepts. Not yet tested on real data yet since I'm [waiting on AWS stuff](https://bugzilla.mozilla.org/show_bug.cgi?id=1275986).


vitillo:
If this patch depends on features of `avro-parquet` that are available only in versions higher than 1.7, then I am afraid that this might not work in deployment as the `avro-parquet` jar used by Spark (which should be 1.7) is the one that gets actually loaded there.

harterrt:
Sorry for the thread necromancy, but what unavailable features does this patch depend on? If I can get this running without a parquet-avro upgrade are we good to go? I'm cautiously hoping this PR became a solution to Bug 1274008 with time. Thanks!

Uberi:
Hey @harterrt,

If I recall correctly, the exact feature that was missing was support for union types in map values. In other words, the following is only valid in newer versions of `parquet-avro` (code taken from this PR):

```scala
.name("user_prefs").`type`().optional().map().values().unionOf()
    .stringType().and()
    .booleanType().and()
    [...]
.endUnion()
```

I was a bit unclear in the bug, but yeah, this should solve it! Let me know if you have any other questions, or if you'd like me to rebase this on the latest master.

mreid-moz:
Drive-by comment... is there a definitive list of prefs (and their types) that could be used to generate a struct with properly-typed fields for each pref we track in this section of the ping?

This is the approach I plan to take for the `main_summary` dataset (for [Bug 1304104](https://bugzilla.mozilla.org/show_bug.cgi?id=1304104), though admittedly the case is easier there since we can add fields for prefs incrementally as they're needed.

vitillo:
@Dexterp37 Afaik any pref can appear there

Uberi:
Hey @mreid-moz, I don't know if this is the best way, but here is one way to get a list of preferences and their types:

1. https://dxr.mozilla.org/mozilla-central/source/modules/libpref/init/all.js seems to have definitions and default values for preferences.
2. Preferences must [be either strings, ints, or booleans](https://dxr.mozilla.org/mozilla-central/source/modules/libpref/prefapi.h?q=PrefType&redirect_type=direct#53).
3. Janky regexes...
    * Booleans: `pcregrep -oM '(?<=\bpref\(")[^"]+(?="\s*,\s*(?:true|false))' all.js > all-booleans.txt`
    * Ints: `pcregrep -oM '(?<=\bpref\(")[^"]+(?="\s*,\s*[\d-])' all.js > all-integers.txt`
    * Strings: `pcregrep -oM '(?<=\bpref\(")[^"]+(?="\s*,\s*["'"'"'])' all.js > all-strings.txt`
4. Success! Here is a list of preferences that are: [strings](https://pastebin.mozilla.org/8926113), [integers](https://pastebin.mozilla.org/8926114), [booleans](https://pastebin.mozilla.org/8926115).
5. Verify:
    * Check number of prefs detected: `wc -l all-booleans.txt all-integers.txt all-strings.txt` (as of this writing, 2555).
    * Check actual number of prefs: `pcregrep -M '\bpref\(' all.js | wc -l` (as of this writing, 2555).
    * That means all prefs were successfully detected as either a string, integer, or boolean.

Dexterp37:
> https://dxr.mozilla.org/mozilla-central/source/modules/libpref/init/all.js seems to have definitions and default values for preferences.

Unfortunately, this doesn't contain a list of all the preferences we report using the main pings (e.g. "app.feedback.baseURL"). Such list is only available in the TelemetryEnvironment file. I'm afraid that there's no easy way to infer the type though.

harterrt:
Thanks, All. I'm going to try using a typed struct to get these fields, similar to the fix suggested by @mreid-moz. I'll keep Bug 1274008 updated.

PR #74, opened 2016-06-06
Use the Telemetry lib instead of the custom, brittle  getRecords() function

maurodoglio:
This is a resurrection of the code in https://github.com/mozilla/telemetry-batch-view/pull/70 which was backed out because it wasn't tested before merging.
I tested it manually on a sample of 100 messages and it seems to be working:
```
=======================================================================================
JOB COMPLETED SUCCESSFULLY FOR 2016-04-30T00:00:00.000+01:00
255 main pings processed, 5 pings ignored
135 crash pings processed, 0 pings ignored
=======================================================================================
```

### DIFF
--- a/src/main/scala/views/CrashAggregateView.scala
+++ b/src/main/scala/views/CrashAggregateView.scala
@@ -15,6 +15,7 @@ import scala.math.{max, abs}
 import telemetry.heka.{HekaFrame, Message}
 import telemetry.ObjectSummary
 import telemetry.utils.Utils
+import telemetry.utils.Telemetry
 import org.joda.time._
 import org.rogach.scallop._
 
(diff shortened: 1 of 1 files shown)


src/main/scala/views/CrashAggregateView.scala
```diff
@@ -53,7 +54,9 @@ object CrashAggregateView {
       val currentDateString = currentDate.toString("yyyy-MM-dd")
 
       // obtain the crash aggregates from telemetry ping data
-      val messages = getRecords(sc, currentDate, "crash").union(getRecords(sc, currentDate, "main"))
+      val messages = Telemetry.getMessages(sc, currentDate, List("telemetry", "4", "crash"))
+        .union(Telemetry.getMessages(sc, currentDate, List("telemetry", "4", "main")))
+        .map(message => HekaFrame.fields(message) + ("payload" -> message.payload.getOrElse("")))
```
maurodoglio: This is the only change I made to https://github.com/mozilla/telemetry-batch-view/pull/70 to make it work with the new activityDate field

vitillo:
Looks good, thanks.

PR #76, opened 2016-06-08
Implement a flexible API to access Heka files on S3.

vitillo:
The current way of accessing files on S3 is very ugly. This patch adds a simple way to fetch Heka files based on a set of filtering criteria on the dimensions, e.g.:

```scala
Dataset("telemetry")
  .where("sourceName") {
    case "telemetry" => true
  }.where("sourceVersion") {
    case "4" => true
  }.where("docType") {
    case "main" => true
  }.where("appName") {
    case "Firefox" => true
  }.where("submissionDate") {
    case date if date.startsWith("2016") => true
  }.records()
```

### DIFF
--- a/src/main/scala/heka/Dataset.scala
+++ b/src/main/scala/heka/Dataset.scala
@@ -1,0 +1,80 @@
+package telemetry.heka
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.json4s._
+import org.json4s.jackson.JsonMethods.parse
+import telemetry.ObjectSummary
+import telemetry.utils.{AbstractS3Store, S3Store}
+import scala.io.Source
+
+case class Schema(dimensions: List[Dimension])
+case class Dimension(fieldName: String)
+
+class Dataset private (bucket: String,
+                       schema: Schema, prefix: String,
+                       clauses: Map[String, PartialFunction[String, Boolean]],
+                       s3Store: => AbstractS3Store) extends java.io.Serializable {
+  def where(dimension: String)(clause: PartialFunction[String, Boolean]): Dataset = {
+    if (clauses.contains(dimension))
+      throw new Exception(s"There should be only one clause for ${dimension}")
+
+    if (!schema.dimensions.contains(Dimension(dimension)))
+      throw new Exception(s"The dimension ${dimension} doesn't exists")
+
+    new Dataset(bucket, schema, prefix, clauses + (dimension -> clause), s3Store)
+  }
+
+  def summaries(fileLimit: Option[Int] = None): Stream[ObjectSummary] = {
+    def scan(dimensions: List[Dimension], prefixes: Stream[String]): Stream[String] = {
+      if (dimensions.isEmpty) {
+        prefixes
+      } else {
+        val dimension = dimensions.head
+        val clause = clauses.getOrElse(dimension.fieldName, {case x => true}: PartialFunction[String, Boolean])
+        val matched = prefixes
+          .flatMap(s3Store.listFolders(bucket, _))
+          .filter{ prefix =>
+            val value = prefix.split("/").last
+            clause.isDefinedAt(value) && clause(value)
+          }
+        scan(dimensions.tail, matched)
+      }
+    }
+
+    val keys = scan(schema.dimensions, Stream(s"${prefix}/"))
+      .flatMap(s3Store.listKeys(bucket, _))
+
+    fileLimit match {
+      case Some(x) => keys.take(x)
+      case _ => keys
+    }
+  }
+
+  def records(fileLimit: Option[Int] = None)(implicit sc: SparkContext): RDD[Message] = {
+    // Partition the files into groups of approximately-equal size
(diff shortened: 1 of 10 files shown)


vitillo:
@mreid-moz r?

src/main/scala/views/MainSummaryView.scala
```diff
@@ -21,7 +21,6 @@ object MainSummaryView {
     val from = opt[String]("from", descr = "From submission date", required = false)
     val to = opt[String]("to", descr = "To submission date", required = false)
     val bucket = opt[String]("bucket", descr = "Destination bucket for parquet data", required = false)
-    val filter = opt[String]("filter", descr = "S3 filter for source data", required = false)
```
mreid-moz: Can we keep this "filter" parameter, and convert it into a `Dataset` filter? It's quite handy during development.

src/test/scala/DatasetTest.scala
```diff
@@ -0,0 +1,63 @@
...
+    case "telemetry/" => Stream("20160606/", "20160607/")
+    case "20160606/" => Stream("main/", "crash/")
+    case "20160607/" => Stream("other/")
+    case "main/" => Stream("Firefox/")
+    case "crash/" => Stream("Fennec/")
+  }
+}
+
+class DatasetTest extends FlatSpec with Matchers{
+  "S3 partitions" can "can be filtered" in {
```
mreid-moz: nit: you have `can` twice here

mreid-moz:
Please add test coverage for `Dataset.toRDD` and the `fileLimit` parameter of `collect()`

src/main/scala/views/MainSummaryView.scala
vitillo: I have added the `fileLimit` parameter to `Dataset.collect` and `Dataset.toRDD` which you can use to e.g. fetch a single file from S3 for development/testing purposes, which is basically what I have been using to test things manually. Is that good enough for your needs?

mreid-moz: That might be good enough - is it deterministic? Will it always return the same set of files given the same filter and limit)?

src/main/scala/heka/Dataset.scala
```diff
@@ -0,0 +1,79 @@
...
+    val keys = scan(schema.dimensions, Stream(s"${prefix}/"))
+      .flatMap(s3Store.listKeys(bucket, _))
+
+    fileLimit match {
+      case Some(x) => keys.take(x)
+      case _ => keys
+    }
+  }
+
+  def toRDD(fileLimit: Option[Int] = None, s3Store: => AbstractS3Store = S3Store)(implicit sc: SparkContext): RDD[Message] = {
```
mreid-moz: This function has a somewhat confusing name. I would expect it to return an `RDD` of the matching object names, but it fetches the messages inside those objects.  Maybe a name like `messages` would be better? If you like the name as-is, then some scaladoc comments to describe the behaviour would be helpful.

src/main/scala/views/MainSummaryView.scala
vitillo: Yes.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -46,17 +45,10 @@ object MainSummaryView {
...
-    val s3filter = conf.filter.get match {
-      case Some(f) => f
-      case _ => "telemetry/4/main/Firefox"
-    }
-
     // Set up Spark
     val sparkConf = new SparkConf().setAppName(jobName)
     sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
-    val sc = new SparkContext(sparkConf)
+    implicit val sc = new SparkContext(sparkConf)
```
maurodoglio: does it have a get-or-create semantic?

vitillo: It creates a new SparkContext.

src/main/scala/views/MainSummaryView.scala
mreid-moz: Ok, that works for me then. Can you add a `limit` param instead?

mreid-moz:
+1 on the new naming in `Dataset`

src/test/scala/DatasetTest.scala
```diff
@@ -0,0 +1,98 @@
...
+    case "20160607/" => Stream("other/")
+    case "main/" => Stream("Firefox/")
+    case "crash/" => Stream("Fennec/")
+  }
+}
+
+class DatasetTest extends FlatSpec with Matchers{
+  "Partitions" can "be filtered" in {
+    val files = Dataset("telemetry", MockS3Store)
+      .where("submissionDate") {
```
maurodoglio: it would be awesome if you could add a couple of tests (more thinking of them as examples here) for more complex filter conditions, like submissionDate.contains('201605') or (docType == 'main' or docType == 'crash')

mreid-moz:
Looks good, r+

maurodoglio:
👍

Issue #58, opened 2016-04-14
Corrupt SEARCH_COUNTS blobs should not be silently discarded


SamPenrose:
Unparseable SEARCH_COUNT key or sum values are silently discarded:

https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/streams/main_summary/Utils.scala#L105
https://github.com/mozilla/telemetry-batch-view/blob/master/src/test/scala/MainSummaryTest.scala#L299

Let's not do this. How about returning something like this if the key is corrupt:

    "engine" -> "__CORRUPTED",
    "source" -> "__CORRUPTED"
and if sum.toInt fails, setting the value to 0?

mreid-moz:
What if we introduce a new (nullable) field for `key`, and make the other three fields nullable too.
Then if we fail to parse the histogram key into an `engine` / `source`, we set the `key` field instead (leaving the other two null), and if we fail to parse the `sum` field, we set the `count` to null.

This would give us the ability to do "light" investigation of bad keys, as well as showing at least the scale of any problem with unparseable sums.

SamPenrose:
Setting engine and source null instead of to a sentinel value is fine. I don't want key but don't object to it. We could certainly set count to null. If we set it to 0, then integer operations still work and the signal is available. If we set it to null then non-SQL access has to test inside each operation, which is a cost.

C.f. issue #59 , where I propose adding 3 new fields to each struct in search_counts.

mreid-moz:
What's the goal / analytic question you want to solve?

SamPenrose:
My goal is to communicate search results accurately. In the short term, this change would allow me to say "0.8% of provider-source values are unreadable, and 1.1% of counts are unreadable" or whatever the real values are. In the medium term we can decide whether to investigate causes.

mreid-moz:
Ok, let's go with the simplest thing and null out those fields in the case where they contain unexpected data.

Most aggregate operations should be resilient to null values, and you can also use the "fill" functions. 
<img width="532" alt="screen shot 2016-04-15 at 11 29 42 am" src="https://cloud.githubusercontent.com/assets/969479/14564572/8358475e-02fd-11e6-80d1-5431b6ca113b.png">

 Does that seem reasonable?

SamPenrose:
That sounds great, thanks so much.

mreid-moz:
Resolved by https://github.com/mozilla/telemetry-batch-view/pull/79

PR #79, opened 2016-06-11
Add several more fields to main_summary v3.

mreid-moz:
This addresses bug 1270482, 1265874 and 1270505, issue #58 on
github, and adds the mean bookmarks and places counts.

We also change the LOOP_ACTIVITY_COUNTER data from a set of
individual fields to a single field containing a struct for all
the enum values (so that we can use the same approach for keyed
histograms too).


mreid-moz:
r? @vitillo

One note: I tested this on April 1 and 2, and the counts did not match exactly. We may have some debugging to do on the new `Dataset` code.

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -90,17 +90,19 @@ object Utils{
```
vitillo: Please move this functionality out from the `streams` package/folder.

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -120,6 +120,41 @@ object Utils{
     }
   }
 
+  // Return a row with the bucket values for the given set of keys as fields.
+  def enumHistogramToRow(histogram: JValue, keys: IndexedSeq[String]): Option[Row] = {
+    histogram \ "values" match {
+      case JNothing => None
+      case v => {
```
vitillo: No need for a block `{}` here.

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -120,6 +120,41 @@ object Utils{
     }
   }
 
+  // Return a row with the bucket values for the given set of keys as fields.
+  def enumHistogramToRow(histogram: JValue, keys: IndexedSeq[String]): Option[Row] = {
```
vitillo: `searchHistogramToRow` is not returning an `Option` but `enumHistogramToRow` is; In theory we should use `Option` everywhere, but since we have to use `nulls` to mark missing fields in a `Row` I think it's acceptable to have those functions return `null`. Either way, we should be consistent.

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -120,6 +120,41 @@ object Utils{
...
+        })
+        Some(Row.fromSeq(values))
+      }
+    }
+  }
+
+  // Return a map of histogram keys to rows with the bucket values for the given set of keys as fields.
+  def keyedEnumHistogramToMap(histogram: JValue, keys: IndexedSeq[String]): Option[Map[String,Row]] = {
+    histogram match {
+      case JObject(x) => {
```
vitillo: I don't think there is a need for a `case` here, you could push this in the [for loop](https://github.com/json4s/json4s#linq-style).

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -120,6 +120,41 @@ object Utils{
...
+        val pns = scala.collection.mutable.HashMap.empty[String,Row]
+        for ((k, v) <- x) {
+          val values = keys.map(_.toString).map(key => v \ "values" \ key match {
```
vitillo: Could you call `enumHistogramToRow` here?

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -120,6 +120,41 @@ object Utils{
...
+          case JInt(n) => n.toInt
+          case _ => 0
+        })
+        Some(Row.fromSeq(values))
+      }
+    }
+  }
+
+  // Return a map of histogram keys to rows with the bucket values for the given set of keys as fields.
+  def keyedEnumHistogramToMap(histogram: JValue, keys: IndexedSeq[String]): Option[Map[String,Row]] = {
```
vitillo: No need for the block `{}` here, you can just write `...]] = histogram match {...}`

src/main/scala/streams/main_summary/Utils.scala
vitillo: No need for a block `{}` here, you can just write `Option[Row] = histogram \ "values" match {`

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -144,6 +179,27 @@ object Utils{
     }
   }
 
+  // Given histogram h, return floor(mean) of the measurements in the bucket.
+  // That is, the histogram sum divided by the number of measurements taken.
+  def histogramToMean(h: JValue): Option[Int] = {
```
vitillo: Don't use a new block, see comments above.

src/main/scala/streams/main_summary/Utils.scala
```diff
@@ -144,6 +179,27 @@ object Utils{
...
+    h \ "sum" match {
```
vitillo: This method doesn't look very idiomatic, please try to use ["LINQ style"](https://github.com/json4s/json4s#linq-style) access if possible.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -324,12 +324,25 @@ object MainSummaryView {
...
-      hval(histograms \ "LOOP_ACTIVITY_COUNTER", "0"),
-      hval(histograms \ "LOOP_ACTIVITY_COUNTER", "1"),
-      hval(histograms \ "LOOP_ACTIVITY_COUNTER", "2"),
-      hval(histograms \ "LOOP_ACTIVITY_COUNTER", "3"),
-      hval(histograms \ "LOOP_ACTIVITY_COUNTER", "4"),
+      settings \ "defaultSearchEngine" match {
+        case JString(x) => x
+        case _ => null
+      },
+      Utils.enumHistogramToRow(histograms \ "LOOP_ACTIVITY_COUNTER", (0 to 4).map(_.toString)).orNull,
```
vitillo: Since enum histogram are always going to be indexed by integers, we could do the conversion to string inside `enumHistogramToRow`. Furthermore, we should read the number of buckets from the lastest `Histograms.json` file so we don't have to pass it as an argument.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -324,12 +324,25 @@ object MainSummaryView {
...
+        case _ => null
+      },
+      Utils.histogramToMean(histograms \ "PLACES_BOOKMARKS_COUNT").orNull,
+      Utils.histogramToMean(histograms \ "PLACES_PAGES_COUNT").orNull,
+      hsum(histograms \ "PUSH_API_NOTIFICATION_RECEIVED"),
+      hsum(histograms \ "WEB_NOTIFICATION_SHOWN"),
+
+      // Messy list of known enum values.
+      Utils.keyedEnumHistogramToMap(keyedHistograms \ "POPUP_NOTIFICATION_STATS",
+        (0 to 8).union(10 to 11).union(20 to 28).union(30 to 31).map(_.toString)).orNull,
```
vitillo: Do you know why some buckets are missing here?

src/main/scala/views/MainSummaryView.scala
```diff
@@ -340,7 +353,43 @@ object MainSummaryView {
     val searchCountsType = StructType(List(
       StructField("engine", StringType, true), // Name of the search engine
       StructField("source", StringType, true), // Source of the search (urlbar, etc)
-      StructField("count", LongType, true) // Number of searches
+      StructField("count",  LongType,   true)  // Number of searches
```
vitillo: Please remove the extra empty space.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -340,7 +353,43 @@ object MainSummaryView {
...
+      StructField("action_1",                  IntegerType, true), // bucket 1
+      StructField("action_2",                  IntegerType, true), // bucket 2
+      StructField("action_3",                  IntegerType, true), // bucket 3
+      StructField("action_last",               IntegerType, true), // bucket 4
+      StructField("dismissal_click_elsewhere", IntegerType, true), // bucket 5
+      StructField("dismissal_leave_page",      IntegerType, true), // bucket 6
+      StructField("dismissal_close_button",    IntegerType, true), // bucket 7
+      StructField("dismissal_not_now",         IntegerType, true), // bucket 8
+      StructField("open_submenu",              IntegerType, true), // bucket 10
+      StructField("learn_more",                IntegerType, true), // bucket 11
```
vitillo: Please use a consistent spacing.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -409,17 +458,43 @@ object MainSummaryView {
...
+      // LOOP_ACTIVITY_COUNTER histogram per bug 1261829
+      StructField("loop_activity_counter", loopActivityCounterType, true),
 
       // DevTools usage per bug 1262478
       StructField("devtools_toolbox_opened_count", IntegerType, true), // DEVTOOLS_TOOLBOX_OPENED_COUNT
 
+      // client date per bug 1270505
+      StructField("client_submission_date", StringType, true), // Fields[Date], the HTTP Date header sent by the client
+
+      // TODO: use median instead of mean of an exponential like:
```
vitillo: Is this TODO relevant for bookmarks?

src/main/scala/views/MainSummaryView.scala
mreid-moz: I thought about doing that with Histograms.json, but we can't just use the number of buckets in the histogram definition, unfortunately. For example, if you look at the `POPUP_NOTIFICATION_STATS` definition, you'll see that it's defined with 40 buckets, but we only have defined meanings for 22 of them :(

src/main/scala/views/MainSummaryView.scala
mreid-moz: No idea...

src/main/scala/views/MainSummaryView.scala
mreid-moz: I initially indexed with a list of integer indices, then thought it seemed messier inside `enumHistogramToRow`. Now that the `keyedEnumHistogramToMap` function calls `enumHistogramToRow`, I think it makes sense to pass in the exact keys you want to use. Otherwise, I think you'd have to re-stringify them to pass to the next function (or have two separate functions, or ugly matchers).  Would you be OK with extracting the key lists to variables?

mreid-moz:
@vitillo Does this look OK now?

mreid-moz:
I tested this again on `20160401` and got the expected number of records.

src/main/scala/utils/MainPing.scala
```diff
@@ -1,16 +1,14 @@
-package telemetry.streams.main_summary
```
vitillo: This should be a package object.

src/main/scala/utils/MainPing.scala
```diff
@@ -90,58 +88,95 @@ object Utils{
...
-      case _ => None
+      if (buf.isEmpty) None
+      else Some(buf.toList)
+    case _ => None
+  }
+
+  // Return a row with the bucket values for the given set of keys as fields.
+  def enumHistogramToRow(histogram: JValue, keys: IndexedSeq[String]): Row = histogram \ "values" match {
+    case JNothing => null
+    case v => {
```
vitillo: There is no need for a block here.

src/main/scala/utils/MainPing.scala
```diff
@@ -90,58 +88,95 @@ object Utils{
...
+      Row.fromSeq(values)
     }
   }
 
+  // Return a map of histogram keys to rows with the bucket values for the given set of keys as fields.
+  def keyedEnumHistogramToMap(histogram: JValue, keys: IndexedSeq[String]): Option[Map[String,Row]] = {
+    val enums = Map[String, Row]() ++ (for {
+      JObject(x) <- histogram
+      (enumKey, enumHistogram) <- x
+    } yield (enumKey, enumHistogramToRow(enumHistogram, keys))).filter(a => a._2 != null)
```
vitillo: You can move the filter condition inside the for comprehension.

src/main/scala/utils/MainPing.scala
```diff
@@ -90,58 +88,95 @@ object Utils{
...
-          Some(topBucket)
-        } else {
-          None
-        }
-      }
-      case _ => {
+  def enumHistogramToCount(h: JValue): Option[Int] = h \ "values" match {
+    case JNothing => None
+    case JObject(x) => {
+      var topBucket = -1
```
vitillo: A fold would make this less convoluted.

src/main/scala/views/MainSummaryView.scala
```diff
@@ -324,13 +328,25 @@ object MainSummaryView {
...
+        case _ => null
+      },
+      MainPing.enumHistogramToRow(histograms \ "LOOP_ACTIVITY_COUNTER", loopActivityCounterKeys),
       hsum(histograms \ "DEVTOOLS_TOOLBOX_OPENED_COUNT"),
-      Utils.getSearchCounts(keyedHistograms \ "SEARCH_COUNTS").orNull
+      fields.getOrElse("Date", None) match {
+        case x: String => x
+        case _ => null
+      },
+      MainPing.histogramToMean(histograms \ "PLACES_BOOKMARKS_COUNT").orNull,
```
vitillo: Could you make all of your helper functions return directly `null` to be consistent as mentioned in a previous comment?

vitillo:
r+ once the comments have been addressed.

src/main/scala/utils/MainPing.scala
vitillo: You could use the `src/main/scala/utils/main` package.

mreid-moz: I'll move this functionality to a package object in a future PR

src/main/scala/utils/MainPing.scala
mreid-moz: I'll make this less dumb :)

src/main/scala/views/MainSummaryView.scala
mreid-moz: I can't return a value of `null` if the return type is declared as a simple scalar type like `Int` or `Boolean` :(

PR #80, opened 2016-06-11
Add logging support for Heka parsing errors.

vitillo:
This patch adds logging support for errors that occur during the parsing
of Heka files. Note that logs are shown in the Spark UI, see also
https://spark.apache.org/docs/latest/running-on-yarn.html#debugging-your-application.


mreid-moz:
Looks good, r+

PR #82, opened 2016-06-20
Don't check test coverage for generated source.

mreid-moz:
No need to write tedious test coverage for protobuf-generated code.


vitillo:
r+

PR #87, opened 2016-06-27
Parallelize Dataset implementation.

vitillo:
This patch makes the scanning of files on S3 about 13x faster on a c3.4xlarge, in doing so it decreases the run-time of the Longitudinal job by about 30 min (10%).

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/87)

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/DatasetTest.scala
+++ b/src/test/scala/com/mozilla/telemetry/DatasetTest.scala
@@ -55,6 +55,7 @@ object MockS3Store extends AbstractS3Store {
       case "Error/" => Stream(ObjectSummary("error", 1), ObjectSummary("missing", 1))
     }
   }
+
   def listFolders(bucket: String, prefix: String, delimiter: String): Stream[String] = prefix match {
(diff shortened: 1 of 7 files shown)


src/test/scala/com/mozilla/telemetry/DatasetTest.scala
```diff
@@ -63,6 +64,10 @@ object MockS3Store extends AbstractS3Store {
     case "crash/" => Stream("Fennec/")
     case "error/" => Stream("Error/")
   }
+
+  def uploadFile(file: java.io.File, bucket: String, prefix: String, name: String) = ???
```
maurodoglio: Why the '???' ? Is this a way to provide an empty implementation?

maurodoglio:
It looks great! :shipit:

src/test/scala/com/mozilla/telemetry/DatasetTest.scala
vitillo: Yes.

PR #92, opened 2016-07-07
Close S3Objects after reading them.

vitillo:
S3Object opens a connection which is not freed even if the object is garbage
collected so it has to be explicitely closed in order to liberate the connection
to the pool.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1285127.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/92)


maurodoglio:
🚢 it

PR #94, opened 2016-07-08
The client count view should go back 6 months by default.

vitillo:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/94)


mreid-moz:
Looks good to me.

PR #98, opened 2016-08-03
Bug 1287101 - Pass the new `processes` structure through

chutten:
After bug 1218576, this is where child histograms could be.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/98)


maurodoglio:
:shipit: !

PR #114, opened 2016-09-20
Create a derived parquet dataset for the sync ping.

mhammond:
This is a first cut that seems to "work", for some definitions of "work".

Alot of code was cargo-culted from MainSummaryView - I'm not sure how much
is relevant here. Some of the new parsing code I added is almost certainly
not idiomatic scala.

There are some tests. Some .json test files are parsed and their contents
verified. One of these payloads is also round-tripped through a local parquet
file and the content verified.

I also had a test that added a .heka file I grabbed of s3 and verified I could
parse it, but I felt that had little value as I couldn't actually verify the
contents. The commit where I removed it is at:
> https://github.com/mhammond/telemetry-batch-view/commit/ad40720d09d1717660bfc0ea140e67e80998f1b1
and I can reinstate that if you like.

Note that the code is handling the older format (where we sent one sync per
ping) and the current format (where we batch a number of syncs into one ping).
@Dexterp37 questioned whether the parquet structure should match the actual
ping formats, but I think this format is preferred - there's one row per
sync-event rather than one row per ping.

@vitillo, @Dexterp37 suggested you might be a good person to flag for this.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/114)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/SyncView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/SyncView.scala
@@ -1,0 +1,384 @@
+package com.mozilla.telemetry.views
+
+import com.mozilla.telemetry.heka.{Dataset, Message}
+import com.mozilla.telemetry.utils.S3Store
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Row, SQLContext}
+import org.apache.spark.{SparkConf, SparkContext}
+import org.joda.time.{DateTime, Days, format}
+import org.json4s.{DefaultFormats, JValue}
+import org.json4s.JsonAST._
+import org.json4s.jackson.JsonMethods.parse
+import org.rogach.scallop._ // Just for my attempted mocks below.....
+
+object SyncView {
+  def streamVersion: String = "v1"
+  def jobName: String = "sync_summary"
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val from = opt[String]("from", descr = "From submission date", required = false)
+    val to = opt[String]("to", descr = "To submission date", required = false)
+    val outputBucket = opt[String]("bucket", descr = "Destination bucket for parquet data", required = false)
+    val outputFilename = opt[String]("outputFilename", descr = "Destination local filename for parquet data", required = false)
+    val limit = opt[Int]("limit", descr = "Maximum number of files to read from S3", required = false)
+    verify()
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args) // parse command line arguments
+    if (!conf.outputBucket.supplied && !conf.outputFilename.supplied)
+      conf.errorMessageHandler("One of outputBucket or outputFilename must be specified")
+    val fmt = format.DateTimeFormat.forPattern("yyyyMMdd")
+    val to = conf.to.get match {
+      case Some(t) => fmt.parseDateTime(t)
+      case _ => DateTime.now.minusDays(1)
+    }
+    val from = conf.from.get match {
+      case Some(f) => fmt.parseDateTime(f)
+      case _ => DateTime.now.minusDays(1)
+    }
+
+    // Set up Spark
+    val sparkConf = new SparkConf().setAppName(jobName)
+    sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
+    implicit val sc = new SparkContext(sparkConf)
+    val sqlContext = new SQLContext(sc)
+    val hadoopConf = sc.hadoopConfiguration
+
+    // We want to end up with reasonably large parquet files on S3.
+    val parquetSize = 512 * 1024 * 1024
+    hadoopConf.setInt("parquet.block.size", parquetSize)
+    hadoopConf.setInt("dfs.blocksize", parquetSize)
+    hadoopConf.set("parquet.enable.summary-metadata", "false")
+
+    for (offset <- 0 to Days.daysBetween(from, to).getDays) {
+      val currentDate = from.plusDays(offset)
+      val currentDateString = currentDate.toString("yyyyMMdd")
+
+      println("=======================================================================================")
+      println(s"BEGINNING JOB $jobName FOR $currentDateString")
+
+      val ignoredCount = sc.accumulator(0, "Number of Records Ignored")
+      val processedCount = sc.accumulator(0, "Number of Records Processed")
+
+      val messages = Dataset("telemetry")
+        .where("sourceName") {
+          case "telemetry" => true
+        }.where("sourceVersion") {
+          case "4" => true
+        }.where("docType") {
+          case "sync" => true
+        }.where("appName") {
+          case "Firefox" => true
+        }.where("submissionDate") {
+          case date if date == currentDate.toString("yyyyMMdd") => true
+        }.records(conf.limit.get)
+
+      val rowRDD = messages.flatMap(m => {
+        messageToRow(m) match {
+          case Nil =>
+            ignoredCount += 1
+            None
+          case x =>
+            processedCount += 1
+            x
+        }
+      })
+
+      val records = sqlContext.createDataFrame(rowRDD, SyncPingConverter.syncType)
+
+      if (conf.outputBucket.supplied) {
+        // Note we cannot just use 'partitionBy' below to automatically populate
+        // the submission_date partition, because none of the write modes do
+        // quite what we want:
+        //  - "overwrite" causes the entire vX partition to be deleted and replaced with
+        //    the current day's data, so doesn't work with incremental jobs
+        //  - "append" would allow us to generate duplicate data for the same day, so
+        //    we would need to add some manual checks before running
+        //  - "error" (the default) causes the job to fail after any data is
+        //    loaded, so we can't do single day incremental updates.
+        //  - "ignore" causes new data not to be saved.
+        // So we manually add the "submission_date_s3" parameter to the s3path.
+        val s3prefix = s"$jobName/$streamVersion/submission_date_s3=$currentDateString"
+        val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
+
+        records.write.mode("error").parquet(s3path)
+
+        // Then remove the _SUCCESS file so we don't break Spark partition discovery.
+        S3Store.deleteKey(conf.outputBucket(), s"$s3prefix/_SUCCESS")
+        println(s"Wrote data to s3 path $s3path")
+      } else {
+        // Write the data to a local file.
+        records.write.parquet(conf.outputFilename())
+        println(s"Wrote data to local file ${conf.outputFilename()}")
+      }
+
+      println(s"JOB $jobName COMPLETED SUCCESSFULLY FOR $currentDateString")
+      println("     RECORDS SEEN:    %d".format(ignoredCount.value + processedCount.value))
+      println("     RECORDS IGNORED: %d".format(ignoredCount.value))
+      println("=======================================================================================")
+    }
+  }
+
+  // Convert the given Heka message containing a "sync" ping
+  // to a list of rows containing all the fields.
+  def messageToRow(message: Message): List[Row] = {
+    val fields = message.fieldsAsMap()
+
+    val payload = parse(message.payload.getOrElse("{}").asInstanceOf[String])
+    SyncPingConverter.pingToRows(payload)
+  }
+}
+
+// Convert Sync pings, which are defined by the schema at:
+// https://dxr.mozilla.org/mozilla-central/source/services/sync/tests/unit/sync_ping_schema.json
+object SyncPingConverter {
+  /*
+   * The type definitions for the rows we create.
+   */
+  private val failureType = StructType(List(
+    // failures are probably *too* flexible in the schema, but all current errors have a "name" and a second field
+    // that is a string or an int. To keep things simple and small here, we just define a string "value" field and
+    // convert ints to the string.
+    StructField("name", StringType, nullable = false),
+    StructField("value", StringType, nullable = true)
+  ))
+
+  // The record of incoming sync-records.
+  private val incomingType = StructType(List(
+    StructField("applied", LongType, nullable = false),
+    StructField("failed", LongType, nullable = false),
+    StructField("newFailed", LongType, nullable = false),
+    StructField("reconciled", LongType, nullable = false)
+  ))
+
+  // Outgoing records.
+  private val outgoingType = StructType(List(
+    StructField("sent", LongType, nullable = false),
+    StructField("failed", LongType, nullable = false)
+  ))
+
+  // The schema for an engine.
+  private val engineType = StructType(List(
+    StructField("name", StringType, nullable = false),
+    StructField("took", LongType, nullable = false),
+    StructField("status", StringType, nullable = true),
+    StructField("failureReason", failureType, nullable = true),
+    StructField("incoming", incomingType, nullable = true),
+    StructField("outgoing", ArrayType(outgoingType, containsNull = false), nullable = true)
+  ))
+
+  // The status for the Sync itself (ie, not the status for an engine - that's just a string)
+  private val statusType = StructType(List(
+    StructField("sync", StringType, nullable = true),
+    StructField("service", StringType, nullable = true)
+  ))
+
+  // The record of a single sync event.
+  def syncType = StructType(List(
+    // These field names are the same as used by MainSummaryView
+    StructField("app_build_id", StringType, nullable = true), // application/buildId
+    StructField("app_display_version", StringType, nullable = true), // application/displayVersion
+    StructField("app_name", StringType, nullable = true), // application/name
+    StructField("app_version", StringType, nullable = true), // application/version
+    // XXX - how do we record the "platform"?
+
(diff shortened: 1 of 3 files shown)


autolander:

vitillo:
The test-suite is failing. Please make sure that your changes don't break the existing tests. You can run the test-suite with `sbt test` from the command line.

vitillo:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 42 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb3VPNT72_YfmMhirz:-KSb3VPNT72_YfmMhis-:bcs7u6e) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L42)):*
> ```Scala
>     }
> 
>     // XXX - the below is copied verbatim from MainSummaryView - I've no idea how much is relevant here.
> ```

Please remove this comment.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 49 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb3X_iMns45VUJlku6:-KSb3X_iMns45VUJlku7:bpoph7n) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L49)):*
> ```Scala
>     val sqlContext = new SQLContext(sc)
>     val hadoopConf = sc.hadoopConfiguration
>     hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
> ```

Please remove this line as it's not needed with EMRFS.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 56 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb3P3h2zabOYxx2Ec0:-KSb3P3h2zabOYxx2Ec1:b-jubm0c) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L56)):*
> ```Scala
>     hadoopConf.setInt("dfs.blocksize", parquetSize)
>     // Don't write temp files to S3 while building parquet files.
>     hadoopConf.set("spark.sql.parquet.output.committer.class", "org.apache.spark.sql.parquet.DirectParquetOutputCommitter")
> ```

Please remove this line as it's no longer needed with Spark 2.0.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 139 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb3yEZLP0GjrO5VlgV:-KSb3yEZLP0GjrO5VlgW:b-9klvt4) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L139)):*
> ```Scala
> 
>     val payload = parse(message.payload.getOrElse("{}").asInstanceOf[String])
>     lazy val build = parse(fields.getOrElse("application.buildId", "{}").asInstanceOf[String])
> ```

`build` isn't used anywhere.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 144 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb4Imd2qWvmMHOK56d:-KSb4Imd2qWvmMHOK56e:b-s45kog) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L144)):*
> ```Scala
> }
> 
> object SyncPingConverter {
> ```

Please link to the sync schema in a comment.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 145 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb4RATQYLwMMFJ6gGh:-KSb4RAUZpUJYXlzZbCd:b6lxk0e) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L145)):*
> ```Scala
> 
> object SyncPingConverter {
>   private val failureType = StructType(List(
> ```

Please move all type definitions (e.g. `failureType` & `incomingType`) together.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 154 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb4relL5NnVZO9BtwI:-KSb4relL5NnVZO9BtwJ:b-xywxt3) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L154)):*
> ```Scala
> 
>   // XXX - this looks dodgy - I'm sure there's a more scala-ish way to write this...
>   private def failureReasonToRow(failure: JValue): Row = failure match {
> ```

What about using [`.extract`](https://github.com/json4s/json4s#extracting-values) with a case class to parse the JSON? 

Ideally you would need to have a set of case classes that describe the JSON schema (for parsing) *and* a potentially distinct set of case classes that describe the schema of the dataset you want to generate. If you do that, then you can use the Spark 2.0 abstraction [`Dataset`](https://databricks.com/blog/2016/01/04/introducing-apache-spark-datasets.html), which allows you to easily generate Parquet data without the need of manually specifying the schema with `StructTypes` and what not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114)*

vitillo:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 154 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSb4relL5NnVZO9BtwI:-KSb8iBstxYSIdylj04l:b-931sg5) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/12921738de1782c9414d7db83df09cd8f3af4f37/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L154)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> What about using [`.extract`](https://github.com/json4s/json4s#extracting-values) with a case class to parse the JSON? 
> 
> Ideally you would need to have a set of case classes that describe the JSON schema (for parsing) *and* a potentially distinct set of case classes that describe the schema of the dataset you want to generate. If you do that, then you can use the Spark 2.0 abstraction [`Dataset`](https://databricks.com/blog/2016/01/04/introducing-apache-spark-datasets.html), which allows you to easily generate Parquet data without the need of manually specifying the schema with `StructTypes` and what not.
</details>
As shown [here](https://spark.apache.org/docs/2.0.0/sql-programming-guide.html#inferring-the-schema-using-reflection), it turns out you don't have to use the `Dataset` abstraction at all in order to use case classes to define a schema

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114)*

mhammond:
I fixed all comments except:

> What about using .extract with a case class to parse the JSON?

That's actually one of the cases where it doesn't seem to make sense, as the name of the second field depends on the contents of the first.

> As shown here, it turns out you don't have to use the Dataset abstraction at all in order to use case classes to define a schema

I hope you can understand that is beyond my Scala capabilities, especially given none of the existing code in this repo seems to do it the way you suggest. I've blown the majority of today on *just* this and failed to make any headway. Please let me know if this is a blocker so I can escalate it upwards to try and find someone else to take this over.

vitillo:
Fair enough, I am happy to merge this as is. Has this job been run/tested with real data as well?

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-:-KSeT7HmR8uANqqVS7x4:bry2fji)*

vitillo:
Please squash your commits into a single one.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-:-KSeUMQ3DDV_phSGtiOS:b-pwshz2)*

vitillo:
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 15 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSepVCEnIvz8tl5UE9o:-KSepVCFIoBmLFOrzWlH:brp0up1) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L15)):*
> ```Scala
> 
>   "Old Style SyncPing payload" can "be serialized" in {
> ```

Please remove newline.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 22 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSeolnSipJUsvVkZ1N3:-KSeolnTzmx8zG-zYaIf:bcs7u6e) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L22)):*
> ```Scala
>     try {
>       val row = SyncPingConverter.pingToRows(SyncViewTestPayloads.singleSyncPing)
>       // Serialize this one row as Parquet
> ```

Please remove this comment.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 27 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSeaxpQTBEHvjgjBHQ6:-KSeaxpQTBEHvjgjBHQ7:bn0w0nm) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L27)):*
> ```Scala
>       val dataframe = sqlContext.createDataFrame(rdd, SyncPingConverter.syncType)
> 
>       dataframe.show() // See the console here
> ```

Please remove this print.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 56 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSeqK5SxElbZ8dGSzkK:-KSeqK5SxElbZ8dGSzkL:bjc73pg) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L56)):*
> ```Scala
>         name match {
>           case "clients" =>
>             took should be (203)
> ```

Ideally you should compare the contents of the Dataframe to the values parsed from the JSON. Doing so would make it simpler to change the pings in the future and it would also allow you to avoid duplicating the code to test multiple records.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 261 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSepnNA_nczLkJKKBbB:-KSepnNBR0aDUMs_PcCL:brp0up1) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L261)):*
> ```Scala
> 
>   "New Style SyncPing payload" can "be serialized" in {
> ```

Please remove newline.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 268 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSep2zT-UrRmNwZRnIB:-KSep2zT-UrRmNwZRnIC:b-kd1prk) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L268)):*
> ```Scala
>     try {
>       val row = SyncPingConverter.pingToRows(SyncViewTestPayloads.multiSyncPing)
>       // Serialize the rows to Parquet
> ```

Please remove comment.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 273 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSep6DvOOPdKG0BexKm:-KSep6DvOOPdKG0BexKn:bow68mk) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L273)):*
> ```Scala
>       val dataframe = sqlContext.createDataFrame(rdd, SyncPingConverter.syncType)
> 
>       dataframe.show() // See the console here
> ```

Please remove print.

---

*[src/test/scala/com/mozilla/telemetry/SyncViewTest.scala, line 299 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSepQUngzO-HRKhpyFt:-KSepQUoWEn5f3xW0pX5:bow68mk) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/test/scala/com/mozilla/telemetry/SyncViewTest.scala#L299)):*
> ```Scala
>       localDataset.registerTempTable("sync")
>       val localDataframe = sqlContext.sql("SELECT * FROM sync")
>       localDataframe.show() // See the console here
> ```

Please remove print.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114)*

vitillo:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/SyncView.scala, line 103 at r2](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-KSfgJk6JT3OMQMA5No2:-KSfgJk6JT3OMQMA5No3:b7qysj7) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/9e022b23d3d3cf1d9643732b40f49b855d223a89/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L103)):*
> ```Scala
>         //  - "ignore" causes new data not to be saved.
>         // So we manually add the "submission_date_s3" parameter to the s3path.
>         val s3prefix = s"$jobName/$streamVersion/submission_date_s3=$currentDateString"
> ```

Give how little data we have for sync (~600 MB per day) we could also regenerate the Parquet dataset over the past 180 days or so in one go. That's clearly more expensive but it would still be less data than a single day of main pings. By doing that it would be easier to deal with schema changes and backfills in the future and we wouldn't have to partition either as the Parquet data for a single day is very small (~15 MB).

Is the schema going to change often? Who is going to maintain this dataset in the future (say if the schema changes or something breaks)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114)*

mhammond:
Thanks - I made the requested changes.

> Is the schema going to change often? Who is going to maintain this dataset in the future (say if the schema changes or something breaks)?

IIUC there are 2 schemas that can change here:

* The schema of the sync ping itself. I suspect most changes here will be addition of extra data, which shouldn't force a change to this code until we actually need that new data on re:dash.

* The schema written to parquet. This is largely an unknown to me, and I suspect is likely to be driven by performance - for example, it may turn out that to get the dashboards we need in a reasonable time, we need the data re-structured - eg, something like the longitudinal datasets across a single sync user.

The latter is largely unknown to me, and I raised a possibility with Dexter that it might make sense to manually run this job once, giving us a static snapshot of the data in re:dash that we can use to explore our dashboards, and only enabling it as a cron job once we are somewhat confident of the data. I don't think anyone in our team has enough experience with re:dash to try and make that determination without access to the data.

However, in both cases, I suspect the work will need to be done by people on our team, simply as that is the only option we have. In practice, I think that will be me, as it is very difficult to justify too many people taking this massive learning curve hit for things which are unlikely to apply to any other work we are likely to do for Mozilla in the medium term.

vitillo:
:lgtm:

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/114#-:-KSqKlueH299zYwBS4Xw:bnfp4nl)*

PR #117, opened 2016-09-22
Bug 1300224 - Create a view for Addons data

mreid-moz:
This view contains one row per active addon, as well as one row with
a null value for addon_id for each client with zero addons.

All fields for each addon are present except the "description" field.

---

This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/117)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala
@@ -1,0 +1,254 @@
+package com.mozilla.telemetry.views
+
+import com.mozilla.telemetry.heka.{Dataset, Message}
+import com.mozilla.telemetry.utils.S3Store
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.util.LongAccumulator
+import org.apache.spark.{SparkConf, SparkContext}
+import org.joda.time.{DateTime, Days, format}
+import org.json4s._
+import org.json4s.jackson.JsonMethods.parse
+import org.rogach.scallop._
+
+import scala.util.Try
+
+case class AddonData(blocklisted: Option[Boolean],
+                     description: Option[String],
+                     name: Option[String],
+                     userDisabled: Option[Boolean],
+                     appDisabled: Option[Boolean],
+                     version: Option[String],
+                     scope: Option[Integer],
+                     `type`: Option[String],
+                     foreignInstall: Option[Boolean],
+                     hasBinaryComponents: Option[Boolean],
+                     installDay: Option[Integer],
+                     updateDay: Option[Integer],
+                     signedState: Option[Integer],
+                     isSystem: Option[Boolean])
+
+object AddonsView {
+  def streamVersion: String = "v1"
+  def jobName: String = "addons"
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val from = opt[String]("from", descr = "From submission date", required = false)
+    val to = opt[String]("to", descr = "To submission date", required = false)
+    val outputBucket = opt[String]("bucket", descr = "Destination bucket for parquet data", required = true)
+    val limit = opt[Int]("limit", descr = "Maximum number of files to read from S3", required = false)
+    val channel = opt[String]("channel", descr = "Only process data from the given channel", required = false)
+    val appVersion = opt[String]("version", descr = "Only process data from the given app version", required = false)
+    verify()
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args) // parse command line arguments
+    val fmt = format.DateTimeFormat.forPattern("yyyyMMdd")
+    val to = conf.to.get match {
+      case Some(t) => fmt.parseDateTime(t)
+      case _ => DateTime.now.minusDays(1)
+    }
+    val from = conf.from.get match {
+      case Some(f) => fmt.parseDateTime(f)
+      case _ => DateTime.now.minusDays(1)
+    }
+
+    // Set up Spark
+    val sparkConf = new SparkConf().setAppName(jobName)
+    sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
+    implicit val sc = new SparkContext(sparkConf)
+    val hadoopConf = sc.hadoopConfiguration
+    hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
+
+    // We want to end up with reasonably large parquet files on S3.
+    val parquetSize = 512 * 1024 * 1024
+    hadoopConf.setInt("parquet.block.size", parquetSize)
+    hadoopConf.setInt("dfs.blocksize", parquetSize)
+    // Don't write metadata files, because they screw up partition discovery.
+    // This is fixed in Spark 2.0, see:
+    //   https://issues.apache.org/jira/browse/SPARK-13207
+    //   https://issues.apache.org/jira/browse/SPARK-15454
+    //   https://issues.apache.org/jira/browse/SPARK-15895
+    hadoopConf.set("parquet.enable.summary-metadata", "false")
+
+    val spark = SparkSession
+      .builder()
+      .appName("AddonsView")
+      .getOrCreate()
+
+    for (offset <- 0 to Days.daysBetween(from, to).getDays) {
+      val currentDate = from.plusDays(offset)
+      val currentDateString = currentDate.toString("yyyyMMdd")
+      val filterChannel = conf.channel.get
+      val filterVersion = conf.appVersion.get
+
+      println("=======================================================================================")
+      println(s"BEGINNING JOB $jobName FOR $currentDateString")
+      if (filterChannel.nonEmpty)
+        println(s" Filtering for channel = '${filterChannel.get}'")
+      if (filterVersion.nonEmpty)
+        println(s" Filtering for version = '${filterVersion.get}'")
+
+      val schema = buildSchema
+      val ignoredCount = sc.longAccumulator("Number of Records Ignored")
+      val processedCount = sc.longAccumulator("Number of Records Processed")
+      val badAddonCount = sc.longAccumulator("Number of Bad Addon records skipped")
+      val messages = Dataset("telemetry")
+        .where("sourceName") {
+          case "telemetry" => true
+        }.where("sourceVersion") {
+          case "4" => true
+        }.where("docType") {
+          case "main" => true
+        }.where("appName") {
+          case "Firefox" => true
+        }.where("submissionDate") {
+          case date if date == currentDate.toString("yyyyMMdd") => true
+        }.where("appUpdateChannel") {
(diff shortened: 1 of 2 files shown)


mreid-moz:
This was tested with Spark 1.6, outputting in `s3://telemetry-test-bucket/addons/v1`. Haven't yet tested with Spark 2, though tests are passing.

mreid-moz:
DO NOT MERGE - this code doesn't appear to work with Spark 2 (yet).

vitillo:
Generally this dataset feels too "low-level" and not something end-users should have access to in stmo. The commit description doesn't explain why need a separate dataset in this form though so I don't fully understand why certain choices were made.

As briefly discussed over IRC this should ideally end up in the main_summary as joining the two datasets is very expensive. We could store the data in an unnormalised form but then we would have to include environment fields that are already present in main_summary dataset, so that doesn't make much sense either.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 17 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSas1pa8S3t2UTvnyO6:-KSas1pa8S3t2UTvnyO7:b-7dmzu0) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L17)):*
> ```Scala
> 
> case class AddonData(blocklisted: Option[Boolean],
>                      description: Option[String],
> ```

Do we really need the description? I recall Dave mentioning that it's not a useful field for analyses.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 28 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSarxySRBcLYv0fmf8J:-KSarxyT1LcJ6ZPPrMxw:bbpei5m) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L28)):*
> ```Scala
>                      installDay: Option[Integer],
>                      updateDay: Option[Integer],
>                      signedState: Option[Option[Integer]],
> ```

This should be `Option[Integer]`.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 186 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSatJkUPt-7pNI9Vm6C:-KSatJkVjUxPaMIfAqFx:b-c9az79) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L186)):*
> ```Scala
> 
>     val activeAddons = addons \ "activeAddons"
>     val addonIds = activeAddons match {
> ```

What about trying to extract this data with something like `.extract[Map[String, AddonData]]`?

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 187 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSasP3zg9JtNK_a734p:-KSasP3zg9JtNK_a734q:b69gmcd) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L187)):*
> ```Scala
>     val activeAddons = addons \ "activeAddons"
>     val addonIds = activeAddons match {
>       case JObject(foo) => foo map(x => x._1)
> ```

Please don't use `foo` as variable name.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 213 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSauQWzbAuoQur3cgkc:-KSauQWzbAuoQur3cgkd:b-cjojig) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L213)):*
> ```Scala
> 
>     if (addonIds.isEmpty)
>       // Output a row with a null addon_id for this submission. This way we can identify pings without any addons.
> ```

This looks weird and should be resolved by including this data in `main_summary`.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 225 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSavEO8xs4tsaOMmITg:-KSavEO8xs4tsaOMmITh:b-f3f382) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L225)):*
> ```Scala
>       StructField("document_id",           StringType, nullable = false), // documentId
>       StructField("client_id",             StringType, nullable = false), // clientId
>       StructField("sample_id",             LongType,   nullable = false), // Fields[sampleId]
> ```

Note that this dataset can't be used as is to rebuild a user's history as there is no data available about sub-session chaining. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-:-KSaqh5RV8pm-N_BcFju:b-3xc2a6)*

mreid-moz:
This dataset is aimed at Spark-based analysis, not stmo. It, similar to `main_summary`, is too heavy to be conveniently queried via Presto.

The intention here was a proof-of-concept to investigate addon data from main pings, particularly DAU for individual addons.

I agree that we will want to move this data into main_summary, such that an Addon-oriented dataset like this can be trivially created via a query against main_summary.

For the purposes of STMO, a stripped-down dataset using HLLs to count unique clients per addon would make more sense going forward.

---

Review status: 1 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 17 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSas1pa8S3t2UTvnyO6:-KSfj3Tt1RtcPJzVAO8K:b-aiozi8) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L17)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> Do we really need the description? I recall Dave mentioning that it's not a useful field for analyses.
</details>
This is just a wrapper class for parsing the addon JSON structure, so it should include all available fields. We don't keep the description field when storing the data.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 28 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSarxySRBcLYv0fmf8J:-KSfjA0DxLCuF7P88PQq:b-886gsx) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L28)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> This should be `Option[Integer]`.
</details>
Fixed

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 186 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSatJkUPt-7pNI9Vm6C:-KSfkVY0OxZdGFgv--zM:b-8qrtbf) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L186)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> What about trying to extract this data with something like `.extract[Map[String, AddonData]]`?
</details>
I wanted to do that, but didn't want to discard an entire record worth of addons if there was a single malformed field in one addon's data. If the counting of bad addon data is to be trusted, it is a non-issue (there were zero bad addons encountered during September).

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 187 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSasP3zg9JtNK_a734p:-KSfjJvmizA0wZKggiEO:b8f448o) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L187)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> Please don't use `foo` as variable name.
</details>
Oops, left this temp name in from an early iteration.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 225 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSavEO8xs4tsaOMmITg:-KSfkzlQLVhCQR2ZPGtH:b5hv0i8) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L225)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary>
> Note that this dataset can't be used as is to rebuild a user's history as there is no data available about sub-session chaining. 
</details>
Correct. You'd have to join with another dataset based on `document_id`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-:-KSbklxfmbqUbmlqcMIj:b-vffqvg)*

vitillo:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/AddonsView.scala, line 186 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117#-KSatJkUPt-7pNI9Vm6C:-KSfrJPjJSs5rdPACUMD:b-umtxvu) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acfd903e31623489971c9397f83b3b017d57f823/src/main/scala/com/mozilla/telemetry/views/AddonsView.scala#L186)):*
<details><summary><i>Previously, mreid-moz (Mark Reid) wrote…</i></summary>
> I wanted to do that, but didn't want to discard an entire record worth of addons if there was a single malformed field in one addon's data. If the counting of bad addon data is to be trusted, it is a non-issue (there were zero bad addons encountered during September).
</details>
I would rather have cleaner code than dealing with malformed data if the latter occurs only rarely (or not at all). Plus, we could count/log how many times parsing fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/117)*

mreid-moz:
Superseded by #134

PR #119, opened 2016-09-23
Update explicit scala version in jar path (E10s dataset)

dzeber:
The command to build the E10s dataset makes reference to the explicit scala version in its args. Updating for future reference, as it changed with the update to Spark 2.0.

It also bears mentioning that this now needs to be run on an EMR-5.0.0 cluster supporting Spark 2.0. I wasn't sure if it was worth adding that to the readme (eg. updating the link to a.t.m.o). If this will soon become the default for our clusters, it may not be necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/119)


autolander:

vitillo:
Looks good, thanks!

PR #124, opened 2016-10-13
Get scalar data from clients with old pings

Dexterp37:
The scalars section was added by bug 1282091, which
landed on July 2016. Pings before that date don't
have the scalars section. If we have a ping with
the missing section among the pings for a client,
the longitudinal view drops all the data from that
client, if if newer pings have the scalar section.
This patch prevents that from happening.
Bonus: this also adds some documentation about
running this code on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/124)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,25 @@ If you run into memory issues during compilation time issue the following comman
 ```bash
 export _JAVA_OPTIONS="-Xss4M -Xmx2G"
 ```
+
+**Running on Windows**
+
+Executing scala/Spark jobs could be particularly problematic on this platform. Here's a list of common issues and the relative solutions:
+
+**Issue:** *I see a weird reflection error or an odd exception when trying to run my code.*
+
+This is probably due to *winutils* being missing or not found. Winutils are needed by HADOOP and can be downloaded from [here](https://github.com/steveloughran/winutils).
+
+**Issue:** *java.net.URISyntaxException: Relative path in absolute URI: ...*
+
+This means that *winutils* cannot be found or that Spark cannot find a valid warehouse directory. Add the following line at the beginning of your entry function to make it work:
+
+```scala
+System.setProperty("hadoop.home.dir", "C:\\path\\to\\winutils")
+System.setProperty("spark.sql.warehouse.dir", "file:///C:/somereal-dir/spark-warehouse")
+```
+
+**Issue:** *The root scratch dir: /tmp/hive on HDFS should be writable. Current permissions are: ---------*
+
+See [SPARK-10528](https://issues.apache.org/jira/browse/SPARK-10528). Run "winutils chmod 777 /tmp/hive" from a privileged prompt to make it work.
+
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -75,6 +75,25 @@ object LongitudinalView {
   // columns.
   private val ScalarColumnNamePrefix = "scalar_"
 
+  // In theory, we shouldn't really need to define this types. However, on Spark 2.0
+  // and Avro 1.7.x this is the only sane way to have Arrays with null elements.
+  // See https://github.com/mozilla/telemetry-batch-view/pull/124#issuecomment-254218256
+  // for context.
+  private val UintScalarSchema = SchemaBuilder
+    .record("uint_scalar").fields()
+      .name("value").`type`().optional().longType()
+    .endRecord()
+
+  private val BoolScalarSchema = SchemaBuilder
+    .record("bool_scalar").fields()
+      .name("value").`type`().optional().booleanType()
+    .endRecord()
+
+  private val StringScalarSchema = SchemaBuilder
+    .record("string_scalar").fields()
+      .name("value").`type`().optional().stringType()
+    .endRecord()
+
   private class ClientIdPartitioner(size: Int) extends Partitioner {
     def numPartitions: Int = size
 
@@ -508,17 +527,20 @@ object LongitudinalView {
       val key = getParquetFriendlyScalarName(k.toLowerCase, "parent")
       value match {
         case s: UintScalar if !s.keyed =>
-          builder.name(key).`type`().optional().array().items().longType()
+          // TODO: After migrating to Spark 2.1.0, change this and the next
+          // definitions to builder.name(key).`type`().optional().items().nullable().theTypeDef()
+          // Context: https://github.com/mozilla/telemetry-batch-view/pull/124#issuecomment-254218256
+          builder.name(key).`type`().optional().array().items(UintScalarSchema)
         case s: UintScalar =>
-          builder.name(key).`type`().optional().map().values().array().items().longType()
+          builder.name(key).`type`().optional().map().values().array().items(UintScalarSchema)
         case s: BooleanScalar if !s.keyed =>
-          builder.name(key).`type`().optional().array().items().booleanType()
+          builder.name(key).`type`().optional().array().items(BoolScalarSchema)
         case s: BooleanScalar =>
-          builder.name(key).`type`().optional().map().values().array().items().booleanType()
+          builder.name(key).`type`().optional().map().values().array().items(BoolScalarSchema)
         case s: StringScalar if !s.keyed =>
-          builder.name(key).`type`().optional().array().items().stringType()
+          builder.name(key).`type`().optional().array().items(StringScalarSchema)
         case s: StringScalar =>
-          builder.name(key).`type`().optional().map().values().array().items().stringType()
+          builder.name(key).`type`().optional().map().values().array().items(StringScalarSchema)
         case _ =>
           throw new Exception("Unrecognized scalar type")
       }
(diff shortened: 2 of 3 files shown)


autolander:

Dexterp37:
@vitillo , with this change if a particular ping doesn't have a specific scalar or a user doesn't send a scalar, it gets set to a default value (0 for uint, false for boolean and "" for strings). This is not desirable. Any idea about how to allow null values in the schema and write them if no scalar is available?

vitillo:
Yeah, that's pretty bad as it could cause analyst to reach wrong conclusions. You should be able to store `null` values within the arrays, have you tried that?

README.md
```diff
@@ -43,3 +43,25 @@ If you run into memory issues during compilation time issue the following comman
 ```bash
 export _JAVA_OPTIONS="-Xss4M -Xmx2G"
 ```
+
+**Running on Windows**
```
vitillo: This change is not logically related to the rest of the patch. Please split the PR into multiple commits.

Dexterp37:
Yeah, I tried by doing that:

- Changed the schema definition for the UintScalar to `builder.name(key).`type`().optional().array().items().nullable().longType()` (~line 511).
- Changed the default values for missing scalars to `null` or `None`, i.e. `vectorizeScalar_(name, payloads, s => s.asInstanceOf[BigInt], null)` (~line 737).

This doesn't seem to work, as it tries to cast the `null`/`None` to a Number and raises an exception:

> java.lang.ClassCastException: scala.None$ cannot be cast to java.lang.Number

vitillo:
Where is that exception thrown?

Dexterp37:
@vitillo, here's the full exception, triggered by `private val path = ParquetFile.serialize(List(record).toIterator, schema):` from LongitudinalTest.scala.

> An exception or error caused a run to abort: scala.None$ cannot be cast to java.lang.Number 
java.lang.ClassCastException: scala.None$ cannot be cast to java.lang.Number
	at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:202)
	at org.apache.parquet.avro.AvroWriteSupport.writeArray(AvroWriteSupport.java:127)
	at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:216)
	at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:112)
	at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:87)
	at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:44)
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.write(InternalParquetRecordWriter.java:121)
	at org.apache.parquet.hadoop.ParquetWriter.write(ParquetWriter.java:258)
	at com.mozilla.telemetry.parquet.ParquetFile$$anonfun$serialize$2.apply(ParquetFile.scala:31)
	at com.mozilla.telemetry.parquet.ParquetFile$$anonfun$serialize$2.apply(ParquetFile.scala:30)
	at scala.collection.Iterator$class.foreach(Iterator.scala:893)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
	at com.mozilla.telemetry.parquet.ParquetFile$.serialize(ParquetFile.scala:30)
	at com.mozilla.telemetry.views.LongitudinalTest$$anon$1.<init>(LongitudinalTest.scala:205)
	at com.mozilla.telemetry.views.LongitudinalTest.<init>(LongitudinalTest.scala:182)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.scalatest.tools.Runner$.genSuiteConfig(Runner.scala:2644)
	at org.scalatest.tools.Runner$$anonfun$37.apply(Runner.scala:2461)
	at org.scalatest.tools.Runner$$anonfun$37.apply(Runner.scala:2460)
	at scala.collection.immutable.List.map(List.scala:273)
	at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:2460)
	at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1044)
	at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1043)
	at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:2722)
	at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1043)
	at org.scalatest.tools.Runner$.run(Runner.scala:883)
	at org.scalatest.tools.Runner.run(Runner.scala)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.runScalaTest2(ScalaTestRunner.java:138)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.main(ScalaTestRunner.java:28)

vitillo:
@Dexterp37 Alright, this can't be checked in as is so you would have to understand if this is a limitation of Avro, Parquet or a bug in their respective implementation. If that's the case, we might have to fall back to another approach, like converting the integers to floats and using nans.

Dexterp37:
You, thanks Roberto. I'll sort this issue out!

Dexterp37:
I think I've figured the problem. Apparently, in order to support Null values in arrays, one should in theory simply make the changes that I suggested in [this comment](https://github.com/mozilla/telemetry-batch-view/pull/124#issuecomment-253545242), by replacing for example

`builder.name(key).`type`().optional().array().items().longType()`

to 

`builder.name(key).`type`().optional().array().items().nullable().longType()`

to allow null values, and then writing `nulls` if no scalar is found. `SchemaBuilder` seems to produce the right schema, as the Avro definition becomes (taken from a scalar from the tests):

>   {
>     "name" : "scalar_parent_mock_scalar_uint",
>     "type" : [ "null", {
>       "type" : "array",
>       "items" : [ "long", "null" ]
>     } ],
>     "default" : null
>   }
  
The "items" entry is a Union type between Longs and Nulls, as expected. When `AvroSchemaConverter` converts this to a Parquet-friendly schema, it generates

>   optional group scalar_parent_mock_scalar_uint (LIST) {
>     repeated int64 array;
>   }
  
using the `repeated` keyword instead of `optional`, thus requiring non-null longs in the scalar and failing to write Parquet files.

This issue is addressed by [PARQUET-285](https://issues.apache.org/jira/browse/PARQUET-285) available in parquet-avro 1.8.0, by using a 3-level structure for list, documented [here ](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md). By using that version of *org.apache.avro* and *org.apache.parquet*, the file is written correctly with `null` values in the array of scalar values. Here's the correct schema generated by the 1.8.0 version:

>   optional group scalar_parent_mock_scalar_uint (LIST) {
>     repeated group list {
>       optional int64 element;
>     }
>   }

Unfortunately, this format change for this very specific use case breaks Spark parquet file loading. A fix is discussed in [SPARK-16344](https://issues.apache.org/jira/browse/SPARK-16344) and will make it to Spark 2.1.0.

Dexterp37:
By nesting the scalar data inside a bogus structure, we manage to get the expected schema layout, thus we're able to use nulls.

>   optional group scalar_parent_mock_scalar_uint (LIST) {
>     repeated group array {
>       optional int64 value;
>     }
>   }

We should probably still change that once we move to Spark 2.1.0. I'm currently testing that on a Spark cluster.

Dexterp37:
This seems to behave correctly, as it shows the [null]s instead of chopping off the data.

> scala> val tst = spark.sql("select scalar_parent_browser_engagement_max_concurrent_tab_count as the_scalar FROM scalars_new").where("scalar_parent_browser_engagement_max_concurrent_tab_count is not null")
> tst: org.apache.spark.sql.Dataset[org.apache.spark.sql.Row] = [the_scalar: array<struct<value:bigint>>]
> 
> scala> tst.show
> +--------------------+
> |          the_scalar|
> +--------------------+
> |[[2], [2], [1], [1]]|
> |          [[1], [1]]|
> |[[1], [7], [8], [...|
> |[[5], [1], [1], [...|
> |               [[8]]|
> |[[1], [2], **[null].**..|
> |[[14], [4], [5], ...|
> |     [[3], [4], [3]]|
> |               [[6]]|
> |[[1], [4], [4], [...|
> |[[11], [7], [2], ...|
> |     [[1], [4], [1]]|
> |     [[5], [1], [1]]|
> |          [[1], [2]]|
> |[[9], [1], [7], [...|
> |[[8], [6], [1], [...|
> |[[1], [1], [1], [...|
> |[[14], [3], [1], ...|
> |[[1], [4], [2], [...|
> |       [[2], [null]]|
> +--------------------+
> only showing top 20 rows

@vitillo , this seems ready for a review :-D

vitillo:
@Dexterp37 In your example query, it seems like each individual scalar value ends up in an array; shouldn't it end up in a record?

vitillo:
Looks good, thanks!

PR #126, opened 2016-10-19
Bug 1302264 - Support gzip-encoded s3 objects

whd:
I didn't see an easy way of adding tests for this since the abstraction used to mock S3 is at the store level for the current tests.

Unfortunately this is a little more difficult to do in the Python Dataset API but I should have a PR for that tomorrow.

---

This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/126)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/S3.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/S3.scala
@@ -2,6 +2,7 @@ package com.mozilla.telemetry.utils
 
 import java.io.{File, InputStream}
 import java.util.UUID
+import java.util.zip.GZIPInputStream
 import awscala.s3.{Bucket, S3}
 import org.apache.log4j.Logger
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


vitillo:
I am not happy about letting this in without tests. Have you considered using something like ScalaMock to add a new test for this case? Has this been tested manually and if so how?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/126#-:-KUS-aRZWNyULWJDrSLj:bg8gxkb)*

whd:
When I realized I would have to add new tests instead of modifying existing ones I decided to throw this up to get an idea for what would recommend on that score.

I tested this manually by comparing `Dataset(`telemetry-sample`)` with `Dataset(`telemetry-sample-2`)` (which is the new gzipped data set) in the test job I used to develop the patch. I only verified that the counts are the same.

whd:
I have taken a closer look at this, and am still not sure how to proceed.

I looked at ScalaMock as you suggested. If I could remove the `AbstractS3Store` and instead simply mock the `awscala.s3.{Bucket, S3}` stuff during testing, then we would be testing the actual implementation of `S3Store` instead of mocking it, which would be great. However, it seems the code was written specifically to use `AbstractS3Store` because right now [mocking singleton objects is non-trivial](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/heka/Dataset.scala#L73-L74) and [not possible with ScalaMock 3](https://paulbutcher.com/2014/04/15/scalamock-status-report/). So currently *none* of the actual implementation of `S3Store` is tested, which includes `getKey()` and my changes to it.

At this point it looks like the work involved in setting up the test infrastructure to test this relatively small change is beyond my limited Scala capacity. I suppose I can add a test that verifies that `HekaFrame.parse` can read gzip input streams, which I verified empirically, but that wouldn't actually be testing the new code per se.

vitillo:
OK, let's address the more general testing issues in a separate PR.

PR #133, opened 2016-10-26
Bug 1311174 - Replace PUSH_API_NOTIFICATION_RECEIVED with PUSH_API_NOTIFY in main_summary

sunahsuh:
PUSH_API_NOTIFICATION_RECEIVED is and has always been opt-in, but PUSH_API_NOTIFY is opt-in as of Firefox 50, which is going out to the release channel on 11/8

Once that's out to release, we can start using PUSH_API_NOTIFY as our telemetry-based source of push notification mau, dau and ER

---

This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/133)


mreid-moz:
The code changes look fine to me, have you tested to make sure the updated schema is compatible with previous runs? We may need to increment the `streamVersion`.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala, line 513 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133#-KV0cxK7xdAryF1PaSp5:-KV0cxK7xdAryF1PaSp6:b-au6faz) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/be30121174fc5b79060b8b73f89931465d63ffd9/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L513)):*
> ```Scala
> 
>       // Push metrics per bug 1270482
>       StructField("push_api_notify", IntegerType, nullable = true), // WEB_NOTIFICATION_SHOWN
> ```

The histogram in the comment should be `PUSH_API_NOTIFY`, I think

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133#-:-KV0dEWsjebfPLydN_kO:bx2c171)*

sunahsuh:
> The code changes look fine to me, have you tested to make sure the updated schema is compatible with previous runs? We may need to increment the streamVersion.

To summarize what we found today:

- If you load a dataset in Spark with different schemas, it *appears* the first schema takes precedence by default. The new column's data is not accessible.
- If you load a dataset in Spark with different schemas but [set `mergeSchema` to True](http://spark.apache.org/docs/latest/sql-programming-guide.html#schema-merging), the resulting dataframe will include both columns and fill in NULL for rows where the data doesn't exist.
- If you load the dataset in this example via parquet2hive, the table will take the schema from the newest file but _the data from the old column is accessible under the new column_. p2h essentially treats our case as a column rename.

vitillo:
>  If you load the dataset in this example via parquet2hive, the table will take the schema from the newest file but the data from the old column is accessible under the new column. p2h essentially treats our case as a column rename.

I think this might happen just by coincidence when the old and new column have the same position in the table. Is this behavior specified somewhere? I am trying to understand if we can rely on it or not.

sunahsuh:
@fbertsch is writing a bunch of schema evolution tests in an ipython notebook which will cover other cases with p2h. Not sure if this behavior is outlined anywhere else though.

sunahsuh:
Frank's results suggest hive is just doing index-based column access https://bugzilla.mozilla.org/show_bug.cgi?id=1313182. Given that, it looks like our best bet for this specific case is to just add a new column at the end and leave the old column in there.

sunahsuh:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala, line 513 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133#-KV0cxK7xdAryF1PaSp5:-KVAlNoh6hpu2UsulH1s:b-896fix) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/be30121174fc5b79060b8b73f89931465d63ffd9/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L513)):*
<details><summary><i>Previously, mreid-moz (Mark Reid) wrote…</i></summary>
> The histogram in the comment should be `PUSH_API_NOTIFY`, I think
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133)*

mreid-moz:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133)*

sunahsuh:
This is deployable now that the config change to use name-based instead of index-based columns in presto [bug 1314622](https://bugzilla.mozilla.org/show_bug.cgi?id=1314622) is active on the cluster

mreid-moz:
:lgtm:

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/133#-:-KW4cd5X3BgTBbkbG-ju:bnfp4nl)*

PR #140, opened 2016-11-04
Add a dimension to track profile age in CrashAggregateView.

chutten:
We suspect that crashiness in the first little while of a user's experience
with Firefox has a disproportionate effect on engagement and retention.

To test this hypothesis, we need to see how crashy things are based on the ages
of the profiles experiencing the crashes.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/140)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala
@@ -103,7 +103,8 @@ object CrashAggregateView {
     List("environment.addons", "activeExperiment", "branch"),
     List("environment.settings", "e10sEnabled"),
     List("environment.settings", "e10sCohort"),
-    List("environment.system", "gfx", "features", "compositor")
+    List("environment.system", "gfx", "features", "compositor"),
+    List("environment.profile", "creationDate")
   )
 
   // names of the comparable dimensions above, used as dimension names in the database
@@ -120,7 +121,8 @@ object CrashAggregateView {
     "experiment_branch",
     "e10s_enabled",
     "e10s_cohort",
-    "gfx_compositor"
+    "gfx_compositor",
+    "profile_age"
   )
 
   val statsNames = List(
@@ -299,9 +301,11 @@ object CrashAggregateView {
               Some(topLevelField)
             } else { // JSON field, the rest of the path tells us where to look in the JSON
               val is_gfx_compositor = dimensionNames(index).equals("gfx_compositor")
+              val is_profile_date = dimensionNames(index).equals("profile_age")
               val dimensionValue = path.tail.foldLeft(parse(topLevelField))((value, fieldName) => value \ fieldName) // retrieve the value at the given path
               dimensionValue match {
                 case JString(value) if is_gfx_compositor && value == "none" => None
+                case JInt(value) if is_profile_date => Some(Math.min(Math.max(-1, Math.round(Days.daysBetween(new DateTime(0).plusDays(value.toInt), activityDate).getDays() / 7)), 53).toString)
                 case JString(value) => Some(value)
                 case JBool(value) => Some(if (value) "True" else "False")
                 case JInt(value) => Some(value.toString)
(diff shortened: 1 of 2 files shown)


autolander:

chutten:
@maurodoglio for discussion, perf testing, etc as previously discussed

maurodoglio:
LGTM. Do you mean to add the profile age to crash_aggregates permanently or is it for a one-off analysis?

chutten:
This is intended to be a permanent change. One of the things we're intending to look at for 2017 is how crash rates in the early days/weeks of a Firefox user's experience changes their engagement and likelihood to stay with Firefox. (the idea, I guess, being that a crashy experience in the first week or two will scare users off). This will require analyses over time.

When I first proposed this you suggested we'd need to test its performance (speed, size) impact to see if it'd perhaps be better to reduce the resolution (from a day to a week or a fortnight). Is this still something we should do? If so, how?

maurodoglio:
To test the impact of this change on the daily job I would run it on a cluster using the method described [here](https://github.com/mozilla/telemetry-batch-view#generating-datasets) for distributed execution. Maybe if you make it run on a date from, say, 1 month ago or so, you can then run queries on stmo and compare its perf with queries on another date. Unfortunately I can't think of a better way 😢

chutten:
Method:
So I sbt assembled the mozilla/master jar and ran it against 2016-10-05, then sbt assembled the chutten/profile_age jar and ran it against 2016-10-06

Strangely, it appears as though they're basically the same size as parquet:
```console
[hadoop@ip-172-31-1-159 telemetry-batch-view]$ aws s3 ls s3://telemetry-test-bucket/crash_aggregates/v1/submission_date=2016-10-05/
2016-11-10 19:08:22          0 _SUCCESS
2016-11-10 19:08:21   77816358 part-r-00000-6398f7fb-cabf-456c-aa16-f2a5fedd06fa.snappy.parquet
[hadoop@ip-172-31-1-159 telemetry-batch-view]$ aws s3 ls s3://telemetry-test-bucket/crash_aggregates/v1/submission_date=2016-10-06/
2016-11-10 19:52:12          0 _SUCCESS
2016-11-10 19:52:10   76627579 part-r-00000-36965722-86e8-4855-aa55-f99913e33718.snappy.parquet
```

I'm now asking @robotblake for assistance in getting these to a point where I can use redash to query them. Hopefully they're large enough that queries against them will be measurably different in duration.

chutten:
After chatting with @robotblake about how best to go about this, we now have two datasets containing crash aggregates from the month of September:

```
s3://telemetry-test-bucket/chutten_with -- should have dimensions['profile_age'] in days
s3://telemetry-test-bucket/chutten_without -- should be normal
```

Once this gets into Presto and re:dash, then the fun (analysis) can begin. With the sizes being so similar in the earlier test, and given how presto does its thing, I'm told it is likely that we won't see too much of a difference in memory/perf on similar queries. Here's hoping!

maurodoglio:
I'll have a look later today

maurodoglio:
@chutten is this ready for another review?

chutten:
Current status: The patches as written aren't going to work. The job runs and runs and runs... then slowly just eats memory until the 24-hour timer expires on the cluster. The going theory is that there are too many profile ages, which is making the job difficult to execute to completion.

So I'm going to have to bucket them. I'm thinking we can express them as integer weeks, clamping them between -1 and 52, and that'll put a limit on how fragmented the rows can be. 

I'll not be getting to that until the new year, so if anyone's looking at this code and wants to take care of it in the interim, be my guest :)

chutten:
Alrighty, after several false starts, this code actually works and generates parquet files roughly 6x the size of previous, containing one additional dimension "profile_age" which is an integer of weeks, and clamped between [-1, 53]

Is this acceptable? It took a long while to derive on a 20-node cluster (roughly 3x as long as previous).

We can coarsen this up even further if necessary to measure fortnights instead of weeks, and/or limit the acceptable range further. But as it stands, it allows for a graph like this:
agevscrashrate

Via a notebook [like this](https://gist.github.com/chutten/0c4dd1a6d0c55b9d8be012a30f4e0df7).

maurodoglio:
@chutten is this for a one-off study or you want to add it to crash_aggregates permanently?

chutten:
@maurodoglio It is hard to say. We're looking at "new users" and wondering about their stability, but it is unclear at this time if we'll need an ongoing measure of this, or just spot checks.

From this shall I deduce that a threefold increase in time deriving the dataset isn't something we're going to want to entertain?

maurodoglio:
Yeah, I was trying to understand if we need to spend time to try to make this faster or not.

vitillo:
Can we close this?

chutten:
For now, yes. It may serve as a useful artefact in the future, but it can do that as well closed as open.

PR #150, opened 2016-12-01
Bug 1321404 - Remove Churn view from telemetry-batch-view

acmiyaguchi:
This view is no longer in use, as per [bug 1321404](https://bugzilla.mozilla.org/show_bug.cgi?id=1321404).

r? @mreid-moz

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/150)


mreid-moz:
Looks good to me.

PR #152, opened 2016-12-14
Bug 1321424 - 'lock' output directory to prevent duplicate runs

sunahsuh:
Adds a placeholder file to the prefix directory before doing the heavy lifting for longitudinal so that any parallel jobs will trigger the require statement checking for an empty prefix. This is the first quick fix for Bug 1321424 before we work on longer-term fixes for the zombie jobs issue.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/152)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -145,8 +145,14 @@ object LongitudinalView {
     val clsName = "longitudinal" // Needed for retro-compatibility with existing data
     val prefix = s"${clsName}/v${opts.to()}"
     val outputBucket = opts.outputBucket()
+    val lockfileName = "RUNNING"
 
     require(S3Store.isPrefixEmpty(outputBucket, prefix), s"s3://${outputBucket}/${prefix} already exists!")
+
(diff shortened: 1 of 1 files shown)


vitillo:
How has this been tested?

sunahsuh:
I've run the beginning of the job, verified the upload, run another job to verify this file triggers the require statement, and then ran the job with the reduce statement commented out to verify the delete works as expected.

As it stands, our views aren't very testable since they're generally long procedural blobs. What we'd do, ideally, is create a job class and run this functionality in a startup/teardown hook, but that's part of a longer refactoring process for our batch views. For now my goal is to get this in the longitudinal job before Sunday's run.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -145,9 +145,15 @@ object LongitudinalView {
     val clsName = "longitudinal" // Needed for retro-compatibility with existing data
     val prefix = s"${clsName}/v${opts.to()}"
     val outputBucket = opts.outputBucket()
+    val lockfile_name = "RUNNING"
```
vitillo: Please use camelCasing.

sunahsuh:
Whoops. Context switching ftl. Fixed!

vitillo:
lgtm

PR #158, opened 2017-01-03
Bug 1309574 - Port topline report to use main_summary dataset

acmiyaguchi:
This ports the topline summary to use the main summary dataset. There is currently a hiccup in the way that I am counting the number of crash pings, but a review of everything else would be nice.

In particular:
* Is this the right place for this job?
* Does the test suite have appropriate coverage? Too specific? Important cases missing?
* Is this approach sane?

feedback? @mreid-moz

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/158)


autolander:

acmiyaguchi:
The report runs without issue on a month of data with a 10 machine cluster (roughly 2+ hours). I haven't validated the results against the original script yet, but I believe the margin of error shouldn't be too wide.

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,492 @@
...
+    var expect = "Other"
+
+    val cols = Seq("country", "channel", "os")
+    for (col <- cols) {
+      val result = df.head().getAs[String](col)
+      assert(result == expect)
+    }
+  }
+
+
```
mreid-moz: nit: please cleanup extra empty lines throughout

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,492 @@
...
+      fake_ping.copy(client_id = "baz", is_default_browser = false)
+    ).toDF()
+    val reportData = ToplineSummary invokePrivate createReportDataset(data)
+    val expect = 3
+
+    val df = ToplineSummary invokePrivate clientValues(reportData, fake_date)
+    val result = df.head().getAs[Long]("actives")
+
+    assert(expect == result)
+  }
```
mreid-moz: Please add some test coverage for the 'join' cases above - namely if a set of dimensions is present in crash data but not main_summary, it should still appear in the result with zero hours, searches, etc. Same for dims present in main_summary but not crashes.

mreid-moz:
Please rebase this PR to resolve the conflict in `build.sbt` too.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,403 @@
...
+        $"client_id",
+        $"submission_date",
+        $"is_default_browser",
+        $"search_counts",
+        normalizeCountry($"country").alias("country"),
+        convertProfileCreation($"profile_creation_date").alias("profile_creation_date"),
+        normalizeChannel($"channel").alias("channel"),
+        normalizeOS($"os").alias("os"),
+        convertHours($"subsession_length".cast(DoubleType)).alias("hours")
+      )
```
acmiyaguchi: I wonder if this is what's causing this report to over count.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,410 @@
...
+    * @return dataframe with aggregated search counts with a column per search engine
+    */
+  private def searchAggregates(reportData: DataFrame): DataFrame = {
+    val s = session
+    import s.implicits._
+
+    val searchSchema = MainSummaryView.buildSearchSchema
+
+    val searchData = reportData
+      .where($"search_counts".isNotNull)
```
SamPenrose: The daily and monthly search rollups aggregate NULLs separately as "NO_SEARCHES" so that for e.g. geo and locale aggregations we don't drop profiles merely because they didn't search during a period. I can't tell whether that concern applies in this code.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,410 @@
...
+      List("^[Windows|WINNT]", "^Darwin", "(?:Linux|BSD|SunOS)")
+        .map(pattern => pattern.r),
+      List("Windows", "Mac", "Linux"),
+      "Other",
+      os)
+  }
+
+  // Also used to create the pivot table
+  private val SearchLabels = List("google", "bing", "yahoo", "other")
+  private val normalizeSearch = udf {
```
SamPenrose: Not blocking this commit: we should move towards a common way of rolling up major engines, so that e.g. internal BD dashboards and this one get the same totals. In particular, the string matching here will include a wide range of false positives.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,410 @@
...
+
+    val searchData = reportData
+      .where($"search_counts".isNotNull)
+      .withColumn("search_counts", explode($"search_counts"))
+      .select(
+        $"country",
+        $"channel",
+        $"os",
+        normalizeSearch($"search_counts.engine").alias("engine"),
+        $"search_counts.count".alias("count"))
```
SamPenrose: Do we catch negative values?

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,495 @@
...
+    val data = Seq(fake_ping.copy(subsession_length = 1800)).toDF()
+    val expect = 0.5
+
+    val df = ToplineSummary invokePrivate createReportDataset(data)
+    val result = df.head().getAs[Double]("hours")
+
+    assert(expect == result)
+  }
+
+  it should "ignore negative numbers" in {
```
mreid-moz: Per Sam's comment, please add a test to make sure negative search counts are also ignored.

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,495 @@
...
+    import sqlContext.implicits._
+    val main_df = uniqueMain(Seq(fake_ping.copy(country="DE"), fake_ping.copy(os="Windows"))).toDF()
+    val reportData = ToplineSummary invokePrivate createReportDataset(main_df)
+    val crashData = uniqueCrashes(Seq(fake_crash, fake_crash)).toDF()
+
+    val df = ToplineSummary invokePrivate easyAggregates(reportData, crashData)
+
+    // there should be 3 entries based on channel, os, country
+    assert(df.count() == 3)
+    df.show()
```
mreid-moz: remove the `df.show()` line

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,495 @@
...
+  it should "use an outer join" in {
+    import sqlContext.implicits._
+    val main_df = uniqueMain(Seq(fake_ping.copy(country="DE"), fake_ping.copy(os="Windows"))).toDF()
+    val reportData = ToplineSummary invokePrivate createReportDataset(main_df)
+    val crashData = uniqueCrashes(Seq(fake_crash, fake_crash)).toDF()
+
+    val df = ToplineSummary invokePrivate easyAggregates(reportData, crashData)
+
+    // there should be 3 entries based on channel, os, country
+    assert(df.count() == 3)
```
mreid-moz: Please check explicitly that
- if a set of dimensions is only in the "report" dataset but not in the "crash" dataset, a line is still included in the output
- if a set of dimensions is only in the "crash" dataset but not in the "report" dataset, a line is still included in the output
- if a set of dimensions is in both datasets, it is still included in the output.

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,495 @@
...
+                            with Matchers
+                            with PrivateMethodTester
+                            with DataFrameSuiteBase {
+  private val dateFormat = format.DateTimeFormat.forPattern("yyyyMMdd")
+  private val fake_date = "20161201"
+  private val past_date = "20161101"
+  private val future_date = "20161203"
+  private val end_period_date = "20161208"
+
+  private val fake_ping =
```
mreid-moz: Please update variable names in this file to use the normal scala conventions (so `fakePing` instead of `fake_ping` as well as other underscore-separated names).

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
acmiyaguchi: This job isn't affected by that issue, however it does seems to be good practice to choose a more transparent default value.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
acmiyaguchi: We don't, it should be straightforward to implement.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
acmiyaguchi: I wonder what would be a good way of solving this problem, since there are numerous google's in the set of all search engines.

One way of approaching this might be to define major search engines by a name grammar (or mapping) and frequency of occurrence over a time period. This could be implemented as a `normalized_engine_name` in a similar fashion to the `normalized_channel`.

SamPenrose: Partner Engineering maintains the correct mappings, but I don't believe they formally manage it as a dependency for us and others (i.e. BD).

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
acmiyaguchi: Actually, I believe this issue is handled by outer joins to the rest of the aggregates, so we don't end up losing any information while doing the roll up.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,410 @@
...
+    * This is done by exploding the nested search_counts structure and applying a pivot over the engines.
+    *
+    * @param reportData dataframe containing the necessary columns
+    * @return dataframe with aggregated search counts with a column per search engine
+    */
+  private def searchAggregates(reportData: DataFrame): DataFrame = {
+    val s = session
+    import s.implicits._
+
+    val searchSchema = MainSummaryView.buildSearchSchema
```
acmiyaguchi: I wonder why I put that there, away it goes!

acmiyaguchi:
@mreid-moz Bumping this for review again.

src/main/scala/com/mozilla/telemetry/views/ToplineSummary.scala
```diff
@@ -0,0 +1,401 @@
...
+  /* Converts days into seconds, for positive values */
+  private val convertDaysToSeconds: UserDefinedFunction = udf {
+    (timestamp: Long) =>
+      timestamp match {
+        case ts if ts > 0 => ts * SecondsInDay
+        case _ => 0
+      }
+  }
+
+  private val clampToPositive = udf { x: Long => if (x < 0) 0 else x }
```
mreid-moz: nit: Please add a blank line after this

src/test/scala/com/mozilla/telemetry/ToplineSummaryTest.scala
```diff
@@ -0,0 +1,537 @@
...
+import org.scalatest._
+
+import scala.util.Try
+
+
+case class Crashes(document_id: String,
+                   country: String,
+                   channel: String,
+                   os: String)
+
```
mreid-moz: nit: Please use a consistent number of blank lines between class definitions (either 1 or 2)

acmiyaguchi:
Failing tests were fixed by removing sbt buffering between test suites.

This has made bucket and prefix required when launching the program. It is assumed that the prefix does not include the version number in it.

mreid-moz:
> This has made bucket and prefix required when launching the program. It is assumed that the prefix does not include the version number in it.
Other jobs include both the name (often `jobName`) and version (often `schemaVersion`) of the dataset in the scala code - let's do the same here and make `prefix` optional (or remove it)

mreid-moz:
Looks good, thanks!

mreid-moz:
Argh, I should've just merged this when I approved it. Please rebase on `master`.

PR #159, opened 2017-01-04
Bug 1302264 - Bytes support

whd:
This should be everything code-wise we need to complete the migration to using the hindsight-generated datasets for telemetry-batch-view jobs.

I've added as a test case the current test ping as decoded by the new infra, which moves `Payload` to `Fields[submission]`. This ping technically fails on the new infra due to having a bad documentId but that doesn't affect these tests.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/159)


whd:
I've squashed this to one commit and renamed the new test file to include that it is generated by hindsight. That addresses all the comments except  the one about `fields("payload")` being referenced, which I think is mistaken as that particular code is the same on master.

vitillo:
Reviewed 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala, line 61 at r3](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j5PIWqwkclkl29w98:-K_kVKAfhK0yEYG5CD_0:b-571a4f) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/51d9b25896818b54f784af2491dd1b1635c2ce1a/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala#L61)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

This doesn't look right. `fields.get("submissions")` returns an `Option[String]`, not a `String`.
</blockquote></details>

This still looks wrong: `message.payload.getOrElse(fields.get("submission"))` yields either a String or Option[Any]; the latter will not match `case value: String`.

---

*[src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala, line 164 at r3](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j4fVkAxTm2uZMnPdl:-K_kW5yo8a6Z5dFPmMGO:b-gj8hws) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/51d9b25896818b54f784af2491dd1b1635c2ce1a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala#L164)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

This doesn't look right. `fields.get("submissions")` returns an `Option[String]`, not a `String`.
</blockquote></details>

See somment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159)*

vitillo:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala, line 61 at r3](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j5PIWqwkclkl29w98:-K_kXom5qM6vUhUxAScl:b-x2xog4) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/51d9b25896818b54f784af2491dd1b1635c2ce1a/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala#L61)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

This still looks wrong: `message.payload.getOrElse(fields.get("submission"))` yields either a String or Option[Any]; the latter will not match `case value: String`.
</blockquote></details>

Actually in the latter case it's `Option[Serializable]` since `Serializable` is the common ancestor of `String` and `Option[T]`. That said, my comment still applies.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159)*

vitillo:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala, line 61 at r3](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j5PIWqwkclkl29w98:-K_kZWjmyamzfy32naCp:b80imsb) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/51d9b25896818b54f784af2491dd1b1635c2ce1a/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala#L61)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

Actually in the latter case it's `Option[Serializable]` since `Serializable` is the common ancestor of `String` and `Option[T]`. That said, my comment still applies.
</blockquote></details>

Err, I meant the whole expression yields an `Option[Serializable]` since it can be either a `String` or an `Option[Any]`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159)*

whd:
I've replaced `fields.get("submission")` with `fields.getOrElse("submission", None)`, which is more what I was going for, is that sufficient?

vitillo:
Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/heka/package.scala, line 14 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j-dco_y711_hixuRR:-K_p8ojEWpiqAqmrh9Fu:bs1cda9) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/975032c4be9264c6be38fbe6d03b2a4a17ed6e0e/src/main/scala/com/mozilla/telemetry/heka/package.scala#L14)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

Please leave a comment on why it's OK to assume that BYTES fields can always be considered UTF-8 strings.
</blockquote></details>

The comment is still missing.

---

*[src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala, line 61 at r5](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-Ka1volUdI_0oRNtiAlA:-Ka1volUdI_0oRNtiAlB:b6agkdw) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acd60e154dd2f26ccd1e2816fc2ff40e6ec8e395/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala#L61)):*
> ```Scala
>           fields.get("docType") match {
>             case Some("crash") => {
>               val payload = message.payload.getOrElse(fields.getOrElse("submission", None)) match {
> ```

s/None/"{}"

---

*[src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala, line 163 at r5](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-Ka1vyegw3oS3tnYlsyd:-Ka1vyegw3oS3tnYlsye:b6agkdw) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/acd60e154dd2f26ccd1e2816fc2ff40e6ec8e395/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala#L163)):*
> ```Scala
>         (message) =>
>           val fields = message.fieldsAsMap
>           val payload = message.payload.getOrElse(fields.getOrElse("submission", None)) match {
> ```

s/None/"{}"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159)*

whd:
I've updated the PR with s/None/"{}", which I assume is better stylistically.

Regarding the missing comment, I assumed it was unnecessary after the squash because the code only made the assumption you wanted a comment about in an intermediate commit.

vitillo:
Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/main/scala/com/mozilla/telemetry/heka/package.scala, line 14 at r1](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-K_j-dco_y711_hixuRR:-KaCZ3CBkh_nfMNvSoS7:bxpp5b2) ([raw file](https://github.com/mozilla/telemetry-batch-view/blob/975032c4be9264c6be38fbe6d03b2a4a17ed6e0e/src/main/scala/com/mozilla/telemetry/heka/package.scala#L14)):*
<details><summary><i>Previously, vitillo (Roberto Agostino Vitillo) wrote…</i></summary><blockquote>

The comment is still missing.
</blockquote></details>

Please add a comment. A bytes field could be anything, like a Hyperloglog. Whoever touches this code needs to know why it can only be an UTF string.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159)*

whd:
I am not sure what you want the comment to contain. As you say a bytes field could be anything, but in the specific case of it being a UTF8 string it's decoded as such. If we need support for bytes fields containing something other than a UTF8 string the code can be updated and I think it's obvious how. With that said I've made a guess as to what an acceptable comment would be and updated the PR. If it's not to your liking please supply one that is ;-)

I've also rebased against master again.

vitillo:
:lgtm:

---

Reviewed 4 of 4 files at r7.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/159#-:-KaIX2K9bM94xrwilHGt:bnfp4nl)*

vitillo:
@whd Please rebase your PR.

whd:
Rebased.

PR #167, opened 2017-01-23
Generate protobuf classes at compile time.

vitillo:
The externally generated protobuf class wasn't compatible with Flink (it was
throwing some serialization errors). This patch fixes that and also
ensures that a change in the protobuf file doesn't require an external
recompiliation step.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/167)


src/main/scala/com/mozilla/telemetry/heka/package.scala
```diff
@@ -2,27 +2,27 @@ package com.mozilla.telemetry
...
+      case Field.ValueType.BOOL => f.valueBool(0)
+      case Field.ValueType.DOUBLE => f.valueDouble(0)
+      case Field.ValueType.INTEGER => f.valueInteger(0)
       case _ => ""
     }
   }
 
   implicit def messageToRichMessage(m: Message): RichMessage = new RichMessage(m)
-}
+}
```
maurodoglio: missing blank line

src/main/scala/com/mozilla/telemetry/heka/HekaFrame.scala
```diff
@@ -50,7 +50,7 @@ object HekaFrame{
         val uncompressedLength = Snappy.uncompressedLength(messageBuffer)
         val uncompressedMessage = new Array[Byte](uncompressedLength)
         Snappy.uncompress(messageBuffer, 0, header.messageLength, uncompressedMessage, 0)
-        Message.parseFrom(uncompressedMessage, 0, uncompressedLength)
+        Message.parseFrom(uncompressedMessage)
```
maurodoglio: Is this the difference between the old protobuff library and the new one?

vitillo: The should both do the same thing but the former method invocation isn't available in the newly generated Scala classes.

vitillo:
Even if tests run successfully don't merge it just yet; I would like to manually test it first on one of our jobs.

src/main/scala/com/mozilla/telemetry/heka/HekaFrame.scala
maurodoglio: I C

maurodoglio:
Reviewed 6 of 8 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-batch-view/167)*

PR #168, opened 2017-01-23
Bug 1329023 - Filter out invalid unicode characters and characters outside the BMP

thomcc:
I'm not sure how to check and see if this fixes the issue. If it doesn't, then it shouldn't be merged. If it does, then I'll add a unit test and remove the unnecessary parts of the diff.

Ping @acmiyaguchi, who is actively working on the issue this patch is trying to avoid.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-batch-view/168)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/SyncView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/SyncView.scala
@@ -4,7 +4,6 @@ import com.mozilla.telemetry.heka.{Dataset, Message}
 import com.mozilla.telemetry.utils.S3Store
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{Row, SparkSession}
-import org.apache.spark.util.LongAccumulator
 import org.apache.spark.{SparkConf, SparkContext}
 import org.joda.time.{DateTime, Days, format}
 import org.json4s.{DefaultFormats, JValue, string2JsonInput}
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
I'll run this patch using similar logic to [this airflow script](https://github.com/mozilla/telemetry-airflow/blob/master/jobs/sync_view.sh).

The date in question is 20161208, and then results should be uploaded to `s3://net-mozaws-prod-us-west-2-pipeline-analysis/amiyaguchi/sync_summary/v1/submission_date_s3=20161208`. I'll report back on whether this works or not.

Thanks for the help :)

acmiyaguchi:
It doesn't look like this patch mitigates the issue.

https://gist.github.com/acmiyaguchi/4610f1a19748908a7314feb0d550fcae

thomcc:
That says it's still the engine.name field, right? If so, I think this would probably mean that it's not the input data, or at least not strictly it. Since that patch [only allows](https://github.com/mozilla/telemetry-batch-view/pull/168/commits/9994f27932b8fa95d461e6e235e7df84f04ac2a0#diff-0577dbcf8deb04bdb2d199d7c733e941R390) engine names in the hardcoded [list of valid names](https://github.com/mozilla/telemetry-batch-view/pull/168/commits/9994f27932b8fa95d461e6e235e7df84f04ac2a0#diff-0577dbcf8deb04bdb2d199d7c733e941R190) through... :(

PR #170, opened 2017-01-25
Fix coverage status.

vitillo:
Some tests stopped running after the introduction of the slow tag which
caused the coverage to drop.


autolander:

src/main/scala/com/mozilla/telemetry/views/AddonsView.scala
```diff
@@ -82,8 +82,9 @@ object AddonsView {
 
       println(s"JOB $jobName COMPLETED SUCCESSFULLY FOR $currentDateString")
       println("=======================================================================================")
-      sc.stop()
     }
+
+    sc.stop()
```
maurodoglio: Hah!

PR #172, opened 2017-01-27
Rename the System case class to SystemData

Dexterp37:
This is needed as, otherwise, we're not able to call functions from System.* (setProperty, getProperty, out, ...). Using some of these functions is needed to develop and test on Windows.

Unless there's another way to call System.* stuff, even if there's a name clash, I'd suggest renaming the case class. Feel free to suggest a different solution if available! (I'm no Scala guru :D)

r?@mdoglio

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
@@ -38,7 +38,7 @@ case class SystemGfx(
     features: Option[SystemGfxFeatures])
 
 
-case class System(os: SystemOs, gfx: Option[SystemGfx])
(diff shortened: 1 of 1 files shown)


maurodoglio:
That was a dumb choice I made indeed! let's have a chat about that

vitillo:
Can't the class simply be private?

maurodoglio:
@Dexterp37 I'm closing it in favor of a proper fix (See [bug 1334457](https://bugzilla.mozilla.org/show_bug.cgi?id=1334457)). In the mean time, you can avoid overriding a package name using the curly braces notation:
```scala
import com.mozilla.telemetry.views.{System => PingSystem}
```

PR #174, opened 2017-02-08
Use moztelemetry library.

vitillo:
The Dataset API has recently been moved to its own package on a Maven
repo so that it can be used by other projects.


build.sbt
```diff
@@ -32,33 +29,24 @@ lazy val root = (project in file(".")).
...
 dependencyOverrides += "com.google.protobuf" % "protobuf-java" % "2.5.0"
 
-// Compile proto files
-PB.targets in Compile := Seq(
-  scalapb.gen() -> (sourceManaged in Compile).value
-)
-
-// Exclude generated classes from the coverage
-coverageExcludedPackages := "com\\.mozilla\\.telemetry\\.heka\\.(Field|Message|Header)"
-
```
mreid-moz: The `run in Compile <<= ...` below generates a deprecation warning. It's unrelated, but kind of annoying. Might as well fix it while we're here.

Looks like replacing with this does the trick:
`run in Compile := Defaults.runTask(fullClasspath in Compile, mainClass in (Compile, run), runner in (Compile, run)).evaluated`

vitillo: Done

PR #182, opened 2017-02-27
Bug 1342817 - Use `submission` instead of `payload` to get ping content

maurodoglio:
The CEP recently changed the name/place where the ping content is stored.


maurodoglio:
I'm not planning to do so, unless someone asks for it explicitly. I'll backfill the last month of data, which should be enough for now. If someone asks for more, I'll write another patch.

PR #183, opened 2017-02-28
Bug 1331702 - Add is_wow64 to main summary

acmiyaguchi:
[Bug 1331702](https://bugzilla.mozilla.org/show_bug.cgi?id=1331702)

Wow64 is used to tell the CPU architecture of Windows machines apart. This adds `is_wow64` field to main summary.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -258,13 +258,6 @@ object MainSummaryView {
     asInt(scalars \ (prefix + engagementMetric))
   }
 
-  def getWow64(system: JValue): Option[Row] = {
-    system \ "isWow64" match {
-      case JBool(wow64) => Some(Row(wow64))
-      case _ => None
-    }
-  }
-
   def asInt(v: JValue): Integer = v match {
     case JInt(x) => x.toInt
     case _ => null
@@ -371,7 +364,10 @@ object MainSummaryView {
           case JInt(x) => x.toLong
           case _ => null
         },
-        getWow64(system).orNull,
+        system \ "isWow64" match {
+          case JBool(x) => x
+          case _ => null
+        },
         profile \ "creationDate" match {
           case JInt(x) => x.toLong
           case _ => null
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -364,6 +371,7 @@ object MainSummaryView {
           case JInt(x) => x.toLong
           case _ => null
         },
+        getWow64(system).orNull,
```
sunahsuh: If this is just emitting a boolean there's no need to create a separate get method and if you did, you wouldn't want it to return Option[Row], it should be Boolean -- in fact, I'm surprised you were able to run this without getting an error when you tried to apply the schema on the DataFrame conversion when testing (did you test this on real data output to telemetry-test-bucket? You can pass in `--channel nightly` to run on just nightly pings)
Check out https://github.com/acmiyaguchi/telemetry-batch-view/blob/5506f9a85fe00b255ac71a231cfbfc086dac26be/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L427 for an example of another boolean field

acmiyaguchi: You're right, I don't think it makes sense. I also pulled it out into a separate method because I wanted to be able to write some sort of test for it, but it might not be necessary with real data.

I didn't test against real data because I thought it would take too long, but `--channel nightly` looks like a sane way to do things.

sunahsuh: Please make sure all your PRs against main_summary and other core datasets are tested on real data -- we have a lot of downstream consumers for whom an outage of main_summary would be a big deal, and our tests unfortunately don't exercise the whole job at the moment.

acmiyaguchi:
I've made the relevant changes to `is_wow64`. I've done some manual validation against the dataset that is covered [in this notebook](https://gist.github.com/acmiyaguchi/e5e0d55d4f7a1d9a6485a97eeec96285). 

@sunahsuh r?

PR #187, opened 2017-03-03
Bug 1339655 - Add SSL_HANDSHAKE_RESULT to main_summary

acmiyaguchi:
[Bug 1339655](https://bugzilla.mozilla.org/show_bug.cgi?id=1339655)

This adds `ssl_handshake_result` to the `main_summary` table. This [validation report](https://gist.github.com/acmiyaguchi/b7deaad26bac276a9fb7d3257d9c6151) shows that it contains valid values.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -290,6 +290,7 @@ object MainSummaryView {
       lazy val parentEvents = payload \ "payload" \ "processes" \ "parent" \ "events"
 
       val loopActivityCounterKeys = (0 to 4).map(_.toString)
+      val sslHandshakeResultKeys = (0 to 671).map(_.toString)
 
       // Messy list of known enum values for POPUP_NOTIFICATION_STATS.
       val popupNotificationStatsKeys = (0 to 8).union(10 to 11).union(20 to 28).union(30 to 31).map(_.toString)
@@ -521,7 +522,12 @@ object MainSummaryView {
         getBrowserEngagement(parentScalars, "total_uri_count"),
         getBrowserEngagement(parentScalars, "unfiltered_uri_count"),
         getBrowserEngagement(parentScalars, "unique_domains_count"),
-        getEvents(parentEvents).orNull
+        getEvents(parentEvents).orNull,
+
+        // bug 1339655
+        MainPing.enumHistogramBucketCount(histograms \ "SSL_HANDSHAKE_RESULT", sslHandshakeResultKeys.head).orNull,
+        MainPing.enumHistogramSumCounts(histograms \ "SSL_HANDSHAKE_RESULT", sslHandshakeResultKeys.tail),
+        MainPing.enumHistogramToMap(histograms \ "SSL_HANDSHAKE_RESULT", sslHandshakeResultKeys)
       )
       Some(row)
     } catch {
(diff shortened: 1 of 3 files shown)


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -736,7 +738,10 @@ object MainSummaryView {
       StructField("total_uri_count", IntegerType, nullable = true),
       StructField("unfiltered_uri_count", IntegerType, nullable = true),
       StructField("unique_domains_count", IntegerType, nullable = true),
-      StructField("events", ArrayType(buildEventSchema, containsNull = false), nullable = true) // payload.processes.parent.events
+      StructField("events", ArrayType(buildEventSchema, containsNull = false), nullable = true), // payload.processes.parent.events
+
+      // bug 1339655
+      StructField("ssl_handshake_result", IntegerType, nullable = true) // SSL_HANDSHAKE_RESULT
```
sunahsuh: 1. Once we turn an enum to Count, does calling it "SSL Handshake Result" still make sense?
2. Does flatting this to a count give you reasonable, actionable information for analysis? I suspect that's not the case and turning the enum histogram into a struct a la `LOOP_ACTIVITY_COUNTER` is the way to go here

acmiyaguchi: Thanks, those are both valid points. I was confused about the [672 buckets](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/Histograms.json#2021,2026), since existing histograms seem to have enumerations that can be translated into a schema by hand. I also couldn't find `LOOP_ACTIVITY_COUNTER` in `Histograms.json` when trying to reference `buildLoopSchema`.

I'll run through this again, this time creating a new field for each bucket and their counts with fields `result_0`-`result_671`.

sunahsuh: I suggest collapsing the buckets to a map? Otherwise you're making an already wide table much wider (and not-fun-to-read code as well)

LOOP_ACTIVITY_COUNTER was removed a few versions back (and, in fact, we should probably remove it from the schema..)

acmiyaguchi: I think I need to spend a little more time grokking this, because I'm having a hard time visualizing what this would look like in its final form.

I don't think wide tables are a bad thing with parquet other than making `printSchema` results noisier. I would imagine having a nested row with all the columns would make it easier to aggregate over individual or groups of buckets. 

If I'm understanding correctly, collapsing to a map would essentially be taking the set of columns (#0-#671) and converting it into a parquet type of map of (bucket #, count). This would be like a keyed histogram, but a 'keyed count' instead.

I *think* that because parquet is column based, compression should be fairly good if the frequency count of each bucket is low. Mapping would probably be just as good or better for data compression because most buckets are probably empty (I don't know what the data should look like). Either way, I think the code in main_summary is going to be straightforward either way.

I'm going to ask for more information about the nature of the analysis and see what approach makes the most sense.

acmiyaguchi: After reading through the relevant bug page and taking a look at the [histogram viewer](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2017-03-02&keys=__none__!__none__!__none__&max_channel_version=nightly%252F54&measure=SSL_HANDSHAKE_RESULT&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2017-02-16&table=1&trim=1&use_submission_date=0), I think the type for `search_counts` is the best way to represent this because of the sparsity of results. 

Creating statistics from arrays of (bucket #, count) is managable with SQL, so I think this is the approach I'll implement.

sunahsuh: Okay, not opposed to this, but I think it also makes sense to get @mreid-moz's thoughts here :)

mreid-moz: This histogram definition appears to be complex enough to warrant special treatment. The enum values are defined as:
> "SSL handshake result, 0=success, 1-255=NSS error offset, 256-511=SEC error offset + 256, 512-639=NSPR error offset + 512, 640-670=PKIX error, 671=unknown err"

I think we should add two scalar fields for convenience:
- `ssl_handshake_result_success` which would contain the count from the "0" bucket of the histogram
- `ssl_hanshake_result_failure` which would contain the total count for all error types (or simply the histogram `sum` minus the success count above).

Then we should also add a Map field `ssl_handshake_result` that contains the full detail of the histogram in (Bucket -> Count) entries.

acmiyaguchi:
I've updated the the PR to include three fields: `ssl_handshake_result_success`, `ssl_handshake_result_failure`, and `ssl_handshake_result`. I've also updated the [validation notebook](https://gist.github.com/acmiyaguchi/b7deaad26bac276a9fb7d3257d9c6151) to include an example query.

@sunahsuh r?

sunahsuh:
Looks great :)

PR #190, opened 2017-03-15
Bug 1318709 - Create pre-release Longitudinal

fbertsch:
This is now per-process and each opt-in or opt-out, not both.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/histograms/Histogram.scala
+++ b/src/main/scala/com/mozilla/telemetry/histograms/Histogram.scala
@@ -15,26 +15,30 @@ case class EnumeratedHistogram(keyed: Boolean, nValues: Int) extends HistogramDe
 case class LinearHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
 case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
 
-object Histograms {
+class HistogramsClass {
   private val processTypes = List("content", "gpu")
-  private case class HistogramLocation(path: List[String], suffix: String)
+  private case class HistogramLocation(processType: String, path: List[String], suffix: String)
   private def generateLocations(histogramKey: String)(processType: String): HistogramLocation = {
     HistogramLocation(
+      processType,
       List("payload", "processes", processType, histogramKey),
       "_" + processType.toUpperCase
     )
   }
 
   // Locations for non-keyed histograms
-  private val histogramLocations = 
-    HistogramLocation(List("payload.histograms"), "") ::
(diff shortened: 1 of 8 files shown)


src/main/scala/com/mozilla/telemetry/histograms/Histogram.scala
```diff
@@ -120,7 +126,7 @@ object Histograms {
 
       val pretty = for {
         (k, v) <- result
-        if v.getOrElse("releaseChannelCollection", Some("opt-in")) == Some("opt-out")
+        if (includeOptin && v.getOrElse("releaseChannelCollection", Some("opt-in")) == Some("opt-in")) || ((!includeOptin) && v.getOrElse("releaseChannelCollection", Some("opt-in")) == Some("opt-out"))
```
harterrt: Let's refactor this logic to a separate function or simplify. 

It looks like this is splitting the longitudinal sets into two distinct sets of opt-in or non-opt-in (as discussed in the bug). If that's the case then `includeOptin` is misleading me. When I hear `includeOptin` I expect it to augment the set, not split. 

I'm having some difficulty typing `Optin`, it feels like a misspelling of `Option`. Does it make sense to use `OptIn` instead?

src/main/scala/com/mozilla/telemetry/histograms/Histogram.scala
```diff
@@ -129,10 +135,19 @@ object Histograms {
         val high = v.getOrElse("high", None).asInstanceOf[Option[Int]]
         val nBuckets = v.getOrElse("n_buckets", None).asInstanceOf[Option[Int]]
 
-        def addSuffixes(key: String, histogram: HistogramDefinition): List[(String, HistogramDefinition)] = {
-          suffixes.map(suffix => (key + suffix, histogram))
+        def addProcessSuffixes(processType: Option[String])(key: String, histogram: HistogramDefinition): List[(String, HistogramDefinition)] = {
```
harterrt: Can we remove this intermediate function by changing the way we create `suffixes`? See the comment above where we define suffixes.

src/main/scala/com/mozilla/telemetry/scalars/Scalars.scala
```diff
@@ -42,6 +45,8 @@ package object Scalars {
       // Emit a scalar definition for each scalar.
       val scalarDefinitions = for {
         (scalarName, v) <- flattenedScalars
+        if (optinOnly && v.asInstanceOf[util.LinkedHashMap[String, Any]].getOrElse("release_channel_collection", "opt-in") == "opt-in") ||
```
harterrt: Same as above, let's refactor this logic to be easier to read.

fbertsch:
Removing per-process stuff, so going to file a new PR.

PR #192, opened 2017-03-24
Sync events view

sunahsuh:
- Moves Events out of MainSummaryView into a utility class
- Adds a job to output Sync Events into a flattened view

No timestamp column yet since I feel like it'd be confusing if we don't use the session start date, which we'll add once it's added to the sync ping.


sunahsuh:
RFAL
Re: refactoring, we've had a bug open for a few months on this https://bugzilla.mozilla.org/show_bug.cgi?id=1313701

PR #194, opened 2017-04-01
Write the real addon id to the addon mapping

Dexterp37:
This changes the format of the addon_mapping.json
file to print both the default addon name and the
real addon id for a given hashed addon id. The new
format is:

{
  "<hashed_id>" : {
    "name" : "<addon default name>",
    "id" : "<addon id>"
  },
  ...
}

The old format was:

{
  "<hashed_id>" : "<addon default name>",
  ...
}


Dexterp37:
@vitillo , this is needed in order to make TAAR return addon-ids instead of blindly returning the addon names. 

I've tested this locally and the files have the expected format. We don't really need the addon names in this new mapping, since we have the ids, but I left them to make it easier to fix what's broken in the dependant services.

Once this is deployed, both TAAR and the page on t.m.o will need to be updated.

PR #196, opened 2017-04-14
Bug 13490813 - Create Prerelease Longitudinal

fbertsch:
This is similar to my previous PR but without the per-process stuff. As such we won't be able to actually create this dataset yet, but the ability is there. I just wanted to use the scalar and histogram parts I created in here for some main_summary work.

Also, now the options are: an opt-out only dataset, or an opt-out and opt-in dataset. No more of the confusing "just opt-in" parts.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -104,9 +104,13 @@ object LongitudinalView {
   }
 
   private class Opts(args: Array[String]) extends ScallopConf(args) {
-    val from = opt[String]("from", descr = "From submission date", required = false)
+    val from = opt[String]("from", descr = "From submission date. Defaults to 6 months before `to`.", required = false)
     val to = opt[String]("to", descr = "To submission date", required = true)
     val outputBucket = opt[String]("bucket", descr = "bucket", required = true)
+    val includeOptin = opt[String]("optin-only", descr = "Boolean, whether to include opt-in data only. Defaults to false.", required = false)
+    val channels = opt[String]("channels", descr = "Comma separated string, which channels to include. Defaults to the four main channels.", required = false)
+    val samples = opt[String]("samples", descr = "Comma separated string, which samples to include. Defaults to 42.", required = false)
+    val telemetrySource = opt[String]("source", descr = "Source for Dataset.from_source. Defaults to telemetry-sample.", required = false)
     verify()
   }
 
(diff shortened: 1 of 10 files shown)


fbertsch:
@harterrt will you be able to get to this r anytime soon? If not we can pass it on to someone else.

harterrt:
Hey, Frank. Sorry for the radio silence :(. 

This is on my radar, but I probably will not be able to review until Friday. I see now that this is blocking some of your work on main_summary. Is that time-line reasonable or should I hand off this review?

fbertsch:
@harterrt Friday works. Thanks for letting me know :)

src/test/scala/com/mozilla/telemetry/histograms/HistogramsTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  val fixture = {
+    new {
+      val histograms = new HistogramsClass{
+        override protected val getURL = (a: String, b: String) => Source.fromFile("src/test/resources/ShortHistograms.json")
+      }
+
+      private val optStatuses = List(false, true)
+
+      val names = optStatuses map {
+        case (optStatus) =>
```
harterrt: Can you drop the case statement here?

src/test/scala/com/mozilla/telemetry/histograms/HistogramsTest.scala
```diff
@@ -0,0 +1,45 @@
...
+            optStatus -> histograms.definitions(optStatus).map{
+              case (k, value) =>
```
harterrt: Is the case statement necessary here?

src/test/scala/com/mozilla/telemetry/LongitudinalTest.scala
```diff
@@ -197,34 +224,57 @@ class LongitudinalTest extends FlatSpec with Matchers with PrivateMethodTester {
...
       sc.setLogLevel("WARN")
+
       private val sqlContext = new SQLContext(sc)
-      val rows = sqlContext.read.load(filename).collect()
-      val row =  rows(0)
+
+      // Opt-out only schema
+      private var includeOptin = false
+      private var histogramDefs = histograms.definitions(includeOptin)
+      private var scalarDefs = scalars.definitions(includeOptin)
```
harterrt: This is personal opinion so these suggestions are totally optional :).

These three lines may be more readable if you used named arguments like `definitions(includeOptin = false)`. Please also consider changing the `var`s to `val`s. Same with the next three statements.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -684,21 +713,25 @@ object LongitudinalView {
     root.set(avroField, fieldValues.asJava)
   }
 
+  def getStandardHistogramSchema(schema: Schema): Schema = {
```
fbertsch: Hate repeating that random string. This function makes it more clear what we're doing.

src/test/scala/com/mozilla/telemetry/histograms/HistogramsTest.scala
fbertsch: Good practice. See http://docs.scala-lang.org/cheatsheets/#pattern_matching

src/test/scala/com/mozilla/telemetry/LongitudinalTest.scala
fbertsch: It was good advice :)

src/test/scala/com/mozilla/telemetry/histograms/HistogramsTest.scala
harterrt: Learn something new every day! Thanks.

src/test/scala/com/mozilla/telemetry/scalars/ScalarsTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  val fixture = {
+    new {
+      val scalars = new ScalarsClass{
+        override protected val getURL = (a: String, b: String) => Source.fromFile("src/test/resources/Scalars.yaml")
+      }
+
+      private val optStatuses = List(true, false)
+
+      val names = optStatuses map {
+        case (optStatus) =>
```
harterrt: Please remove this case statement as well.

fbertsch:
@harterrt fixed - RFAL

PR #197, opened 2017-04-17
Remove duplicate addon class definitions

wlach:


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -19,6 +19,22 @@ import com.mozilla.telemetry.histograms._
 import com.mozilla.telemetry.scalars._
 import com.mozilla.telemetry.utils._
 
+// this represents add-on data as serialized by the longitudinal view
+case class ParquetAddon(addon_id: String,
```
vitillo: When importing the current definition from the `views` namespace it isn't clear at all to which definition of Addon this case class refers to. I think the definition should live in a namespace that reflects the dataset it belongs to, e.g. `com.mozilla.telemetry.views.longitudinal.Addon`. You could also consider making the Addon definition an inner class of LongitudinalView.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -19,6 +19,22 @@ import com.mozilla.telemetry.histograms._
...
+                        blocklisted: Boolean,
+                        name: String,
```
vitillo: This looks wrong, please use the [same definition](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala#L365) generated by the Longitudinal dataset. In particular, optional types should remain such. Since you are at it, it would be helpful to investigate why using non optional types didn't cause the tests of the cross sectional dataset to fail.

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -25,20 +25,6 @@ import scala.sys.process._
 
 private case class Rating(clientId: Int, addonId: Int, rating: Float)
 private case class Addons(client_id: Option[String], active_addons: Option[Map[String, Addon]])
-private case class Addon(blocklisted: Option[Boolean],
```
vitillo: The addon recommender uses the [Addon definition](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L543) in the `main_summary` dataset , not the one defined in the `longitudinal` one. Even if they currently match, there is no guarantee that they will keep doing so in the future. If you want to enforce a certain common structure across datasets at read time, then you have to do so also at write time or this patch could introduce bugs in the future.

src/test/scala/com/mozilla/telemetry/AddonsViewTest.scala
```diff
@@ -1,31 +1,16 @@
 package com.mozilla.telemetry
 
-import com.mozilla.telemetry.views.AddonsView
+import com.mozilla.telemetry.views.{ParquetAddon, AddonsView}
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.{SparkConf, SparkContext}
 import org.scalatest.{FlatSpec, Matchers}
 
-case class Addon(addon_id: String,
```
vitillo: See my comment for `AddonRecommender`.

wlach:
Need to look into this more closely, I think I misunderstood some things the first time around. May put this on short term hold since I don't think (even if done correctly) it would be all that much of a win.

maurodoglio:
It sounds like this pull request can be closed for now.

PR #199, opened 2017-04-24
Bug 1359268 - Add the target device OS and version to the sync event telemetry when possible.

thomcc:
It's not practical for the redash queries to pull this data out of the sync ping (for numerous and unfortunate reasons).

(cc @DavisMTL who has been trying to work around this being missing all day)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
@@ -156,8 +156,10 @@ object SyncEventConverter {
   ) ++ eventFields
     ++ List(
       StructField("event_device_id", StringType, nullable = true), // present in most events
-      StructField("event_flow_id", StringType, nullable = true) // present in most events
-    )
+      StructField("event_flow_id", StringType, nullable = true), // present in most events
+      StructField("event_device_version", StringType, nullable = true), // present in most events
+      StructField("event_device_os", StringType, nullable = true) // present in most events
+  )
   )
 
   def pingToRows(ping: JValue): List[Row] = {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
```diff
@@ -219,12 +221,51 @@ object SyncEventConverter {
...
+            }
+            if (devID != null && devVer != null && devOS != null) {
+              Some(Map[String, (String, String)]((devID, (devVer, devOS))))
+            } else {
+              None
+            }
+          })
+        }
+        case _ => List()
+      }) ++ List(Map.empty[String, (String, String)])).reduce((a, b) => a ++ b)
```
sunahsuh: It took me a minute to realize there's an extra close-parens here that means the reduce acts on the whole output instead of just the empty Map -- is there a way to add a newline or change the indentation to make that a bit clearer?

src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
```diff
@@ -219,12 +221,51 @@ object SyncEventConverter {
...
+          }
+
+        List(
+          deviceID,
+          x getOrElse ("flowID", null),
+          deviceVersion,
+          deviceOS
+        )
+      }
       case _ => List(null, null)
```
sunahsuh: Please update the catch-all guard to match the expanded fields

sunahsuh:
Looks good other than those two comments -- the backfill only takes ~20 minutes a day on a 1-node machine so I'm inclined to blow away the whole dataset and redo the whole thing with the new fields. Let me know if you'd like support on that

thomcc:
Updated to match changes.

Yes, replacing the current data w/ a backfill sounds good to me.

sunahsuh:
Okay -- I don't have time to get everything kicked off today but I'll get a backfill started first thing tomorrow

PR #200, opened 2017-04-25
Remove `date` column from ToplineSummaryView

acmiyaguchi:
This column was previously added for compatibility with the original `v4_reformat.py` script. This was implemented incorrectly (it should have used yyyy-MM-dd) and is redundant because of report_start.

The topline_dashboard etl job will handle the date field instead.


acmiyaguchi:
I think yyyyMMdd is more consistent with how we partition other datasets.

PR #201, opened 2017-04-27
Rename ToplineSummary to ToplineSummaryView

acmiyaguchi:
This naming convention is more consistent with other classes in this project. This also fixes failing airflow jobs that used ToplineSummaryView instead of ToplineSummary.


mreid-moz:
Err, approved too soon - build appears to be failing.

PR #203, opened 2017-05-01
Add generalized longitudinal job

fbertsch:
This will allow us to make Longitudinal-type datasets out of any parquet datasets. I'll be using this for the mobile ones to begin with. With Presto now supporting lambda-expressions, these datasets will be easy-to-use in STMO.

The basic idea here is to give it a dataset name, a grouping column, and any number of ordering columns. Most datasets will be grouped by client_id, and ordered by submission_date or activity_date.

Ideas for more test cases are welcome!

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
@@ -1,0 +1,159 @@
+package com.mozilla.telemetry.utils
+
+import java.math.BigDecimal
+import java.sql.{Date, Timestamp}
+import org.apache.spark.sql.expressions.{MutableAggregationBuffer, UserDefinedAggregateFunction}
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types._
+
+class CollectList(inputStruct: StructType, orderCols: List[String], maxLength: Option[Int]) extends UserDefinedAggregateFunction {
+  /**
+   * This transforms the groups to arrays. It is similar to collect_list, but
+   * we can't use that because it throws out null values, which we want to retain.
+   * Additionally, it sorts and trims them before outputting.
+   **/
+
+  private val outputSchema = StructType(
+    inputStruct.fields.map{
+      sf => StructField(sf.name, ArrayType(sf.dataType), sf.nullable)
+    }
+  )
+
+  private val numFields = inputStruct.fields.length
+
+  override def inputSchema: StructType = inputStruct
+
+  override def bufferSchema: StructType = outputSchema
+
+  override def dataType: DataType = outputSchema
+
+  override def deterministic: Boolean = false
+
+  override def initialize(buffer: MutableAggregationBuffer): Unit = {
+    (0 to numFields - 1).foreach{
+      n => buffer(n) = Nil
+    }
+  }
+
+  override def update(buffer: MutableAggregationBuffer, input: Row): Unit = {
+    (0 to numFields - 1).foreach{
+      n => buffer(n) = buffer.getSeq(n) :+ input(n)
+    }
+  }
+
+  override def merge(buffer1: MutableAggregationBuffer, buffer2: Row): Unit = {
+    (0 to numFields - 1).foreach{
+      n => buffer1(n) = buffer1.getSeq(n) ++ buffer2.getSeq(n)
+    }
+  }
+
+  /**
+   * SORTING
+   *
+   * We want a descending, stable sort - this includes null values.
+   * As such, our sorting is:
+   *
+   * case e1==null, e2==null  => given ordering
+   * case e1, e2==null => e1 before e2
+   * case e1==null, e2 => e2 before e1
+   * case e1 > e2 => e1 before e2
+   * case e1 < e2 => e2 before e1
+   * case e1 == e2 => given ordering
+   **/
+
+  private def sortInt(arr: Seq[Integer])(e0: Int, e1: Int): Boolean = {
+    val v1 = arr(e0)
+    val v2 = arr(e1)
+    (v1 != null) && ((v2 == null) || v1 > v2)
+  }
+
+  private def sortString(arr: Seq[String])(e0: Int, e1: Int): Boolean = {
+    val v1 = arr(e0)
+    val v2 = arr(e1)
+    (v1 != null) && ((v2 == null) || v1 > v2)
+  }
+
+  private def sortLong(arr: Seq[java.lang.Long])(e0: Int, e1: Int): Boolean = {
+    val v1 = arr(e0)
+    val v2 = arr(e1)
+    (v1 != null) && ((v2 == null) || v1 > v2)
+  }
+
+  private def sort(order: Seq[Int], orderType: DataType, seq: Seq[Any]): Seq[Int] = {
+    orderType match {
+      case _: IntegerType => order.sortWith(sortInt(seq.asInstanceOf[Seq[Integer]]))
+      case _: StringType => order.sortWith(sortString(seq.asInstanceOf[Seq[String]]))
+      case _: LongType => order.sortWith(sortLong(seq.asInstanceOf[Seq[java.lang.Long]]))
+      case other =>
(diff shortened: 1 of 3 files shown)


fbertsch:
@harterrt I figured out a way to make it *wayyy* simpler :)

harterrt:
Cool, should I hold off on reviewing for a while then?

fbertsch:
@harterrt nope, the simpler version is the current version. Carry on!

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  private def sortInt(arr: Seq[Integer])(e0: Int, e1: Int): Boolean = {
+    val v1 = arr(e0)
+    val v2 = arr(e1)
+    (v1 == null && v2 == null) || ((v1 != null) && ((v2 == null) || v1 >= v2))
+  }
+
+  private def sort(order: Seq[Int], orderType: DataType, seq: Seq[Any]): Seq[Int] = {
+    orderType match {
+      case _: IntegerType => order.sortWith(sortInt(seq.asInstanceOf[Seq[Integer]]))
```
harterrt: Do we need an Option here?

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  "longitudinal" must "contain all groups" in {
+    val clientIds = fixture.longitudinal.map(_.get(fixture.fieldIndex("client_id"))).toList
+    assert(clientIds == List("1", "2"))
+  }
+
+  "longitudinal" must "correctly aggregate strings" in {
+    val oss = fixture.longitudinal.map(_.getSeq(fixture.fieldIndex("os")).toList).toList
+    assert(oss == List(List("nothing", "Android", "iOS"), List("Android", null)))
```
harterrt: Shouldn't the OS list for client 2 be : `List(null, "Android")`?

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
```diff
@@ -0,0 +1,144 @@
...
+    val numNulls =
+      (1 to (row.size - 1))
+      .map(row.getSeq(_).contains(null))
+      .map(if(_) 1 else 0)
+      .reduce(_ + _)
+
+    assert(numNulls == 2)
+  }
+
+  "longitudinal" must "sort null last" in {
```
harterrt: This seems wrong to me

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -0,0 +1,150 @@
...
+   * As such, our sorting is:
+   *
+   * case e1==null, e2==null  => e1 before e2
+   * case e1, e2==null => e1 before e2
+   * case e1==null, e2 => e2 before e1
+   * case e1 >= e2 => e1 before e2
+   * case e1 < e2 => e2 before e1
+   **/
+
+  private def sortString(arr: Seq[String])(e0: Int, e1: Int): Boolean = {
```
harterrt: Do we have any datasets where we would want to sort by an `Option[Int]`?

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
fbertsch: So "Sort null last" means that, for a give sorting column[s], null values will always be last. So in your example, the results would be as follows:

## Sorting by data_a
grouping | data_a | data_b
--------- | ------- | ------
a | [6, null] | [null, 9]

## Sorting by data_b
grouping | data_a | data_b
--------- | ------- | ------
a | [null, 6] | [9, null]

fbertsch: For reference, see the big "SORTING" block comment in the CollectList class.

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
fbertsch: Right now both Integer and String allow null values. Some others - such as Long - don't, so we're going to have to deal with an Option[Long] for that sorting case.

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
fbertsch: Nope, Integer allows for null values.

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
fbertsch: It's sorting by `submission_date_s3`, which has `"20160101"` for `"Android"`, and `null` for `null`. The `null` value is sorted last, so the OS list is `List("Android", null)`.

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -0,0 +1,150 @@
...
+      case _: LongType      => sortTrimTypedArray[Long](buffer.getSeq(arrayIndex), sortOrder)
+      case _: FloatType     => sortTrimTypedArray[Float](buffer.getSeq(arrayIndex), sortOrder)
+      case _: DoubleType    => sortTrimTypedArray[Double](buffer.getSeq(arrayIndex), sortOrder)
+      case _: DecimalType   => sortTrimTypedArray[BigDecimal](buffer.getSeq(arrayIndex), sortOrder)
+      case _: StringType    => sortTrimTypedArray[String](buffer.getSeq(arrayIndex), sortOrder)
+      case _: BinaryType    => sortTrimTypedArray[Array[Byte]](buffer.getSeq(arrayIndex), sortOrder)
+      case _: BooleanType   => sortTrimTypedArray[Boolean](buffer.getSeq(arrayIndex), sortOrder)
+      case _: TimestampType => sortTrimTypedArray[Timestamp](buffer.getSeq(arrayIndex), sortOrder)
+      case _: DateType      => sortTrimTypedArray[Date](buffer.getSeq(arrayIndex), sortOrder)
+      case _: ArrayType     => sortTrimTypedArray[Seq[Any]](buffer.getSeq(arrayIndex), sortOrder)
```
fbertsch: Maps seem to work, based on the test I have. I can add another if you think the specific check is not covered.

fbertsch:
@harterrt answered most queries above, made some minor changes. RFAL.

fbertsch:
@harterrt made a few changes:
1. Added sorting for Long type
2. Made sorting of arrays O(n) rather than O(nlog(n))
3. Fixed secondary ordering test
4. Sorting functions return false if the values are equal - per sorting spec

sunahsuh:
Commenting to register my interest in using this for experiments. For my needs, I'll need multiple grouping columns (`client_id`, `experiment_name`, `experiment_branch`) but I'm happy to make the change to allow that after this is merged.

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
harterrt: Missed that test. This is good to go.

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
harterrt: As discussed offline, I (incorrectly) thought this was ordered by `order`. No change needed.

src/test/scala/com/mozilla/telemetry/views/GeneralizedLongitudinalTest.scala
harterrt: As with above, you're right. I didn't realize this was ordered by `submission_date` and not by `order`.

PR #206, opened 2017-05-04
Bug 1361819 - Missing "esr" and "Other" data in recent longitudinal tables

harterrt:
Makes `channels` option default to all channels


harterrt:
Thanks, Frank!

Added a second commit that makes SBT executable - let's see if the tests pass 🤞

PR #207, opened 2017-05-08
Update client_count to use current main_summary version

fbertsch:
It wasn't downstream, so I missed it during my first pass.


mreid-moz:
@fbertsch please rebase on the current master

PR #209, opened 2017-05-10
Bug 1329541: ... `session_start_date` is null for all records

harterrt:
This change adds session_start_date to the longitudinal dataset.

Note that this change removes the calls to `return` and consequentially
changes the behavior of the code in the case of missing
`subsession_start_date` values. However, these values are not missing in
practice, and this appears to be an undefined edge case currently.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -895,19 +895,22 @@ object LongitudinalView {
     }
   }
 
-  private def subsessionStartDate2Avro(payloads: List[Map[String, Any]], root: GenericRecordBuilder, schema: Schema) {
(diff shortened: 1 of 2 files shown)


harterrt:
Not yet, I'll rebase and do a test run this afternoon before merging.

harterrt:
I ran the updated job over a week of data. The data is at:
s3://telemetry-test-bucket/longitudinal/v20170506/

I verified that the job runs to completion and that this field is now populated. I'm waiting on a computation to verify there are the same number of rows in this dataset and the prod version.

harterrt:
Well, it looks like my new run of the longitudinal table includes 70 client_ids which were not represented in the production run on Sunday. It looks like each of these clients exclusively had pings from 2017-05-06, which is the final day of aggregation. My suspicion is that the additional rows come from the change in run date, and not the change in code.

@vitillo, can you confirm this is still good to merge?

Investigation details here:
https://gist.github.com/harterrt/bc7cba55469909a9cb1c17be225f4699

PR #211, opened 2017-05-12
Bug 1364617 - Remove crash aggregates from ToplineSummary

acmiyaguchi:
[Bug 1364617 - Remove crash aggregates from ToplineSummary](https://bugzilla.mozilla.org/show_bug.cgi?id=1364617)

These aggregates aren't being shown on the executive dashboard, so this removes it.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
@@ -146,50 +146,6 @@ object ToplineSummaryView {
         convertSecondToHours($"subsession_length".cast(DoubleType)).alias("hours"))
   }
 
-  /**
-    * Create the dataset required for counting the number of crashes
-    *
-    * This is currently implemented by searching through the raw pings. This could be done using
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
```diff
@@ -371,11 +321,11 @@ object ToplineSummaryView {
       val dataset: DataFrame = s.read.parquet(MainSummaryURL)
 
       val reportData = createReportDataset(dataset, from, to)
-      val crashData = createCrashDataset(from, to)
 
-      val finalReport = easyAggregates(reportData, crashData)
+      val finalReport = easyAggregates(reportData)
         .join(clientValues(reportData, from), Seq("country", "channel", "os"), "outer")
         .withColumnRenamed("country", "geo")
+        .withColumn("crashes", lit(0))
```
mreid-moz: Please add a comment in the code here linking to the bug and describing why we set crashes to zero.

acmiyaguchi:
@mreid-moz RFAL

PR #217, opened 2017-05-18
WIP: Experiments analysis job

sunahsuh:
This is the first PR to get reviews on the unfinished experiment analysis job. Currently, the job reads in a longitudinal-shaped parquet file (with two additional two columns: `experiment_name` and `experiment_branch`), aggregates histograms and scalars, runs a few simple summary statistics on the aggregate histograms and outputs a parquet file with a row for each grouping of: experiment branch, subpopulation (not yet implemented), and metric.

I'm specifically hoping to get suggestions for improvement on:
Class organization: this is a bit of a mess -- partly because I hacked up a first version to fake some output for the experiment-viewer team to work with and built up from there. I feel like I've been staring at this too long to be able to have a good perspective on this at this point. This includes structure as well as things like naming (I don't particularly like <MetricType>Analyzer class names, for instance and suggestions would be great.)
Performance: any obvious perf improvements? This currently runs reasonably fast, although aggregation time is significant, and the exploded dataset takes up a surprising amount of memory right now.
Histogram gotchas: I haven't worked much with the longitudinal dataset so I may be missing subtleties in working with histograms.

Suggested review order:
- com.mozilla.telemetry.views.ExperimentAnalysisView: This is the entrypoint to the job. The view loads the longitudinal-shaped data, loads the current list of histograms and scalars, and runs the aggregation & analysis on each metric, applies the output schema to the result and saves it to disk

- com.mozilla.telemetry.experiments.analyzers.MetricAnalyzer: This is the shared base for the histogram & scalar analyzers. `analyze()` should probably be the only public method (I'll change this for the next PR). This class encapsulates the common steps for aggregating and analyzing histogram and scalars. The companion object provides an interface for passing in a metric definition and getting the corresponding child analyzer back.

- com.mozilla.telemetry.experiments.analyzers.histograms: This is the specialized handling for the various types of histograms. Each histogram type is normalized into an Array of `Long` to standardize aggregation and summary statistics. There's a bit of duplicated code in the classes which I'd had abstracted out in a complicated trait-based inheritance tree but I'm much happier now to trade off the duplication for simplicity (but I'm happy to take suggestions on improving this.)

- com.mozilla.telemetry.experiments.analyzers.scalars: The scalar counterpart to the above. Scalars are *not* normalized into a common format before aggregation since we have things like string scalars. Note BooleanScalars are not yet implemented, in large part because we don't have any boolean scalars out in the wild yet so I don't have real data to work with while writing the code (but I'll develop this with tests before this job is done.)

- com.mozilla.telemetry.utils.userdefinedfunctions: These are the User Defined sparksql functions that aggregate histograms and scalars, and implement the base UserDefinedAggregateFunction interface. I note that Frank merged in a com.mozilla.telemetry.utils.udfs package and I should either move these there or under com.mozilla.telemetry.experiments.udfs. I'm a bit concerned about putting them until utils.udfs since our histograms need to be normalized into an Array of `Long` before running them through the aggregator. I suspect a good future refactoring that'd make this generally useful for folks would be to split out histogram/scalar aggregation so that you can feed in a raw longitudinal column and it'd do all the normalization and such under the hood.

- com.mozilla.telemetry.experiments.analyzers.MetricAggregate: This class takes an aggregated metric and spits out some summary statistics. I need to write unit tests for this class so I can check my math. Also, for Linear/exponential histograms, this should probably be taking the midpoint between buckets as the "value". I only have this working for histograms at the moment, and this doesn't make sense for all histograms (e.g. enum histograms) so this class might need a bit of refactoring.

-  com.mozilla.telemetry.histograms.HistogramDefinition & com.mozilla.telemetry.scalars.ScalarDefinition & com.mozilla.telemetry.util.MetricDefinition: Defines a base trait for HistogramDefinition and ScalarDefinition that exposes the "keyed" value, which greatly simplifies the analysis code. Thinking about this some more now, we should probably move the Histogram.scala, Scalars.scala and MetricDefinition.scala files under a com.mozilla.telemetry.metrics package and normalize the pluralization for the filenames (possibly as a separate PR to be merged in earlier?)

- com.mozilla.telemetry.ExperimentAnalysisViewTest: Pretty barebones for the moment, but exercises a keyed and not-keyed version of each histogram type. This should really be renamed as a test for HistogramAnalyzer, and I'll add separate ones for ScalarAnalyzer, MetricAggregate, and the UDFs with smaller unit tests.

The big outstanding task for this job is to write the distance function and the permutation test. I've added some scaffolding for where it's going to go but I'm sure I'll need to make changes as I implement it here. Other large outstanding tasks for this job are:
- Grouping by subpopulations (currently, these are just a list of OSes.)
- Supporting opt-in metrics (made easier by the recent changes to histogram definitions.)
- Implementing any other statistic functions that might be useful

Sorry for dumping this big of a PR -- future ones will be smaller and more frequent.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
@@ -11,6 +11,8 @@ import scala.util.{Failure, Success, Try}
 
 
 abstract class MetricAnalyzer(name: String, md: MetricDefinition, df: DataFrame) extends java.io.Serializable {
+  type AggregateType
+
   val reducer: UserDefinedAggregateFunction
   def handleKeys: Column
   val keyedUDF: UserDefinedFunction
@@ -67,15 +69,38 @@ abstract class MetricAnalyzer(name: String, md: MetricDefinition, df: DataFrame)
 
   def runSummaryStatistics(rows: List[Row]): List[List[Row]] = {
     // TODO: fill me in!
-    List()
+    List.fill(rows.length)(List())
   }
 
   def runTestStatistics(filtered: DataFrame, rows: List[Row]): List[List[Row]] = {
     // TODO: fill me in!
-    List()
+    List.fill(rows.length)(List())
   }
 
-  def toFinalSchema(rows: List[Row], summary_stats: List[List[Row]], test_stats: List[List[Row]]): List[Row]
+  def aggToMap(values: AggregateType): Map[Long, Row]
+
+  def toFinalSchema(rows: List[Row], summary_stats: List[List[Row]], test_stats: List[List[Row]]): List[Row] = {
+    println(rows)
+    (rows zip summary_stats zip test_stats).map {
+      case (((row, row_summary_stats), row_test_stats)) => {
+        println(row)
+        val stats = (row_summary_stats ++ row_test_stats) match {
+          case Nil => null
+          case x => x
+        }
+        row match {
+          case Row(experiment:    String,
+                   branch:        String,
+                   subgroup:      String,
+                   n:             Long,
+                   histogram_seq: AggregateType,
+                   metric_name:   String,
+                   metric_type:   String) => {println(histogram_seq); Row(experiment, branch, subgroup, n, metric_name,
+                                               metric_type, aggToMap(histogram_seq), stats)}
+        }
+      }
+    }
+  }
 }
 
 object MetricAnalyzer {
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
@@ -1,0 +1,95 @@
+package com.mozilla.telemetry.experiments.analyzers
+
+import com.mozilla.telemetry.utils.MetricDefinition
+import org.apache.spark.sql.expressions.{UserDefinedAggregateFunction, UserDefinedFunction}
+import org.apache.spark.sql.functions.{col, count, explode, lit}
+import org.apache.spark.sql.{Column, DataFrame, Row}
+import com.mozilla.telemetry.scalars._
+import com.mozilla.telemetry.histograms._
+
+import scala.util.{Failure, Success, Try}
+
+
+abstract class MetricAnalyzer(name: String, md: MetricDefinition, df: DataFrame) extends java.io.Serializable {
+  val reducer: UserDefinedAggregateFunction
+  def handleKeys: Column
+  val keyedUDF: UserDefinedFunction
+
+  def analyze(): List[Row] = {
+    println(name)
+    val filtered = formatAndFilter match {
+      case Some(x: DataFrame) => x.persist()
+      case _ => return List()
+    }
+    val rows = aggregate(explodeMetric(filtered)).rdd.collect.toList
+    val summary_stats = runSummaryStatistics(rows)
+    val test_stats = runTestStatistics(filtered, rows)
+    filtered.unpersist()
+    toFinalSchema(rows, summary_stats, test_stats)
+  }
+
+  // We want to keep this output since we'll need this to do the experimental distance metrics
+  // Question: for permutation tests, are the null values significant? e.g. should we be doing the tests on the DF before
+  // filtering out nulls?
+  def formatAndFilter: Option[DataFrame] = {
+    Try(df.select(
+      col("experiment").as("experiment_name"),
+      col("branch").as("experiment_branch"),
+      lit("All").as("subgroup"),
+      col(name).as("metric"))
+      .filter(col("metric").isNotNull)
+    ) match {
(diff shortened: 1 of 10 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,104 @@
...
+    val conf = new Conf(args)
+
+    val spark = SparkSession
+      .builder()
+      .master("local[*]")
+      .appName(jobName)
+      .getOrCreate()
+
+    val hadoopConf = spark.sparkContext.hadoopConfiguration
+    hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
```
vitillo: As we are using EMRFS this line isn't needed, I think.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    // DEBUG -- remove before committing
+    spark.sparkContext.setLogLevel("WARN")
+    val parquetSize = 512 * 1024 * 1024
```
vitillo: Why did you have to tune the blocksize?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.rogach.scallop.ScallopConf
+import org.apache.spark.sql.types._
+
+object ExperimentAnalysisView {
+  def schemaVersion: String = "v1"
+  def jobName: String = "experiment_analysis"
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    // TODO: change to s3 bucket/keys
```
vitillo: Is this TODO still valid?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,104 @@
...
+    } ++
+    histogramList.flatMap {
+      case(name: String, hd: HistogramDefinition) =>
+        MetricAnalyzer.getAnalyzer(
+          name.toLowerCase, hd, rows
+        ).analyze()
+      case _ => List()
+    }
+
+    output.foreach(println)
```
vitillo: Please remove this print. If you want to log something then you should use log4j.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,120 @@
...
+
+abstract class MetricAnalyzer(name: String, md: MetricDefinition, df: DataFrame) extends java.io.Serializable {
+  type AggregateType
+
+  val reducer: UserDefinedAggregateFunction
+  def handleKeys: Column
+  val keyedUDF: UserDefinedFunction
+
+  def analyze(): List[Row] = {
+    println(name)
```
vitillo: Please remove prints.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,120 @@
...
+    val summary_stats = runSummaryStatistics(rows)
+    val test_stats = runTestStatistics(filtered, rows)
+    filtered.unpersist()
+    toFinalSchema(rows, summary_stats, test_stats)
+  }
+
+  // We want to keep this output since we'll need this to do the experimental distance metrics
+  // Question: for permutation tests, are the null values significant? e.g. should we be doing the tests on the DF before
+  // filtering out nulls?
+  def formatAndFilter: Option[DataFrame] = {
```
vitillo: I find the DataFrame API to be pretty horrible in general. I wonder if you could use the Dataset API for this job?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  def runTestStatistics(filtered: DataFrame, rows: List[Row]): List[List[Row]] = {
+    // TODO: fill me in!
+    List.fill(rows.length)(List())
+  }
+
+  def aggToMap(values: AggregateType): Map[Long, Row]
+
+  def toFinalSchema(rows: List[Row], summary_stats: List[List[Row]], test_stats: List[List[Row]]): List[Row] = {
```
vitillo: More prints to remove :) I iterate my sentiment that this job could be simplified if it would be written to use the DataFrame API.

fbertsch:
@sunahsuh should we close this now?

sunahsuh:
Yeah, sure (I've been referring back to it while writing v2 but there's no real reason to keep it open :))

PR #220, opened 2017-05-20
add isWebExtension and multiprocessCompatible to addons struct

SamPenrose:
Add two optional fields to the addon struct in main_summary.


src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
```diff
@@ -573,13 +579,13 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
         val dataframe = spark.createDataFrame(sc.parallelize(rows.toSeq), schema)
         val tempFile = com.mozilla.telemetry.utils.temporaryFileName()
-        dataframe.write.parquet(tempFile.toString)
-
-        // Then read it back
-        val data = spark.read.parquet(tempFile.toString)
-
-        data.count() should be (1)
-        data.filter(data("document_id") === "foo").count() should be (1)
+//        dataframe.write.parquet(tempFile.toString)
```
fbertsch: I know the tests are failing because of this section, but we should probably find the root cause before we make any commits commenting it out.

SamPenrose: Whoops! That was not in there on purpose -- my apologies. Will uncomment and debug.

PR #223, opened 2017-05-23
Bug 1366253 - Expose active experiments in main summary

wlach:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -214,6 +214,20 @@ object MainSummaryView {
     }
   }
 
+  def getExperiments(experiments: JValue): Option[List[Row]] = {
+    println(experiments)
```
mreid-moz: Please remove this `println`

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
```diff
@@ -25,6 +25,9 @@ case class Attribution(source: Option[String],
                        campaign: Option[String],
                        content: Option[String])
 
+case class Experiment(id: Option[String],
```
mreid-moz: Hm, per the [Telemetry Environment](https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/data/environment.html) docs, I don't think the experiment structs have an `id` field.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -214,6 +214,20 @@ object MainSummaryView {
...
+    implicit val formats = DefaultFormats
+    Try(experiments.extract[Map[String, Experiment]]) match {
+      case Success(experiments) => {
+        val rows = experiments.map { case (experimentId, experimentData) =>
+          Row(experimentId, experimentData.branch.orNull)
```
mreid-moz: Per discussion in the associated bug, I think we want the experiments field to mirror the `Map[String, Experiment]` structure, rather than converting to a list of rows.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -772,7 +796,11 @@ object MainSummaryView {
       StructField("plugins_notification_user_action", buildPluginNotificationUserActionSchema, nullable = true),
       StructField("plugins_infobar_shown", IntegerType, nullable = true),
       StructField("plugins_infobar_block", IntegerType, nullable = true),
-      StructField("plugins_infobar_allow", IntegerType, nullable = true)
+      StructField("plugins_infobar_allow", IntegerType, nullable = true),
+
+      // bug 1366253 - active experiments
+      StructField("experiments", ArrayType(buildExperimentSchema, containsNull = false), nullable = true) // One per item in environment.experiments
```
mreid-moz: This field should use `MapType`

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
wlach: Oh grump, I misread the bug. I'll give this another try.

src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
```diff
@@ -773,7 +799,8 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
             "scalar_parent_mock_scalar_string"       -> null,
             "scalar_parent_mock_scalar_uint"         -> null,
             "scalar_parent_mock_uint_optin"          -> null,
-            "scalar_parent_mock_uint_optout"         -> null
+            "scalar_parent_mock_uint_optout"         -> null,
+            "experiments"                            -> Some(Some(Map()))
```
wlach: So I figured I should test the case where an experiments field wasn't present, so added this. However, I'm confused about why it needs to be `Some(Some(Map))` and not just `Some(Map)`. If I make it the latter, I get the following error:

```bash
[info] Heka records
[info] - can be summarized *** FAILED ***
[info]   experiments: Some(Map()) was not equal to Some(Some(Some(Map()))) (MainSummaryViewTest.scala:808)
```

@sunahsuh @mreid-moz any ideas?

wlach:
I think I fixed the issues, though one thing is confusing me. See comment. :)

src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
sunahsuh: Seems like the expected value should be `Map()` instead of Some(Map())

wlach: Indeed, that's what I can't figure out. :(

sunahsuh: Sorry, let me clarify: there's no option involved at all in the output - you call `orNull` immediately after invocation do it should be `experiments -> Map ()` as the expected value

wlach:
Ok, I think we're actually good now

src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
wlach: I think null actually makes more sense for the expected value, though it took some more changes to make that happen. Anyway, I think we're good now.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -214,6 +214,22 @@ object MainSummaryView {
     }
   }
 
+  def getExperiments(experiments: JValue): Option[Map[String, String]] = {
+    implicit val formats = DefaultFormats
+    Try(experiments.extract[Map[String, Experiment]]) match {
+      case Success(experiments) => {
+        if (experiments.size > 0) {
+          Some(experiments.map { case (experimentId, experimentData) =>
```
sunahsuh: Style nit: since you have a line break anyway it's probably a bit more idiomatic to have the case guard and the output on the same line. The other little idiomatic thing I'd add is that Scala style tends toward terse names for short-lived variables (e.g. `case (id, data) => id -> data.branch.orNull`)

wlach:
Great, fixed the nits, rebased on master and will merge.

PR #224, opened 2017-05-23
Add Quantum Release criteria to main_summary

fbertsch:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -245,6 +245,8 @@ object MainSummaryView {
       lazy val weaveMobile = MainPing.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
       lazy val parentScalars = payload \ "payload" \ "processes" \ "parent" \ "scalars"
       lazy val parentEvents = payload \ "payload" \ "processes" \ "parent" \ "events"
+      lazy val childHistograms = payload \ "payload" \ "processes" \ "parent" \ "histograms"
```
mreid-moz: I believe `childHistograms` should use `payload \ "payload" \ "processes" \ "content" \ "histograms"`, same with `childKeyedHistograms` below.

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
```diff
@@ -159,6 +159,19 @@ object MainPing{
     }
   }
 
+  // Return the number of counts greater than threshold
```
mreid-moz: This comment could use a bit more description -- maybe an example, as well as a note that if you don't use bucket boundary values, results may not be exact (like if you used `6` as the threshold in the `json4` test case, it would exclude all the values in the 5 to 99 bucket).

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
```diff
@@ -159,6 +159,19 @@ object MainPing{
...
+  def histogramToThresholdCount(histogram: JValue, threshold: Int): Long = {
+    implicit val formats = org.json4s.DefaultFormats
+
+    histogram \ "values" match {
+      case JNothing => 0
+      case v => v
+        .extract[Map[String, Int]]
+        .filterKeys(s => s.toInt >= threshold)
```
mreid-moz: Consider using the `toInt(s: String)` function here for resilience to bad data. If a histogram has a corrupted key that can't be converted to an int, it'll raise an exception.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -590,7 +610,8 @@ object MainSummaryView {
       StructField("install_day",           IntegerType, nullable = true),
       StructField("update_day",            IntegerType, nullable = true),
       StructField("signed_state",          IntegerType, nullable = true),
-      StructField("is_system",             BooleanType, nullable = true)
+      StructField("is_system",             BooleanType, nullable = true),
+      StructField("is_web_extension",      BooleanType, nullable = true)
```
mreid-moz: This field is already being added in #220

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -518,8 +520,26 @@ object MainSummaryView {
         MainPing.enumHistogramToRow(histograms \ "PLUGINS_NOTIFICATION_USER_ACTION", pluginNotificationUserActionKeys),
         hsum(histograms \ "PLUGINS_INFOBAR_SHOWN"),
         hsum(histograms \ "PLUGINS_INFOBAR_BLOCK"),
-        hsum(histograms \ "PLUGINS_INFOBAR_ALLOW")
+        hsum(histograms \ "PLUGINS_INFOBAR_ALLOW"),
 
+        // bug 1366838 - Quantum Release Criteria
+        MainPing.histogramToThresholdCount(histograms \ "GC_MAX_PAUSE_MS", 150),
```
mreid-moz: Per the above comment on the function definition, do we need to ensure that the threshold values used here map exactly to histogram bucket boundaries? Or are approximate counts good enough? Based on the [histogram simulator](https://telemetry.mozilla.org/histogram-simulator/#low=1&high=1000&n_buckets=50&kind=linear&generate=normal), `GC_MAX_PAUSE_MS` has a nearby bucket boundary of 147.

mreid-moz:
Er, never mind re: updating the docs, they've moved :)

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
fbertsch: I had considered this but came to no conclusions. I sent an email out to clarify, I'm waiting to push the new code until we get confirmation from the stakeholders.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
maurodoglio: Should it be called contentHistograms?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -245,6 +245,8 @@ object MainSummaryView {
       lazy val weaveMobile = MainPing.enumHistogramToCount(histograms \ "WEAVE_DEVICE_COUNT_MOBILE")
       lazy val parentScalars = payload \ "payload" \ "processes" \ "parent" \ "scalars"
       lazy val parentEvents = payload \ "payload" \ "processes" \ "parent" \ "events"
+      lazy val childHistograms = payload \ "payload" \ "processes" \ "parent" \ "histograms"
+      lazy val childKeyedHistograms = payload \ "payload" \ "processes" \ "parent" \ "keyedHistograms"
```
maurodoglio: Are you sure this is the right path to childKeyedHistograms ?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -772,7 +793,22 @@ object MainSummaryView {
       StructField("plugins_notification_user_action", buildPluginNotificationUserActionSchema, nullable = true),
       StructField("plugins_infobar_shown", IntegerType, nullable = true),
       StructField("plugins_infobar_block", IntegerType, nullable = true),
-      StructField("plugins_infobar_allow", IntegerType, nullable = true)
+      StructField("plugins_infobar_allow", IntegerType, nullable = true),
+
+      // bug 1366838 - Quantum Release Criteria
+      StructField("gc_max_pause_ms_parent_150_above_count", LongType, nullable = true),
```
fbertsch: @maurodoglio and I have agreed on the following naming convention: <histogram_name>_<process>_above_<threshold>

<process> = [main, content]

fbertsch:
@mreid-moz, @maurodoglio - I added an isQuantumReady field, and expanded to make sure we include all possible RC (150, 250, 2500 for each process-histogram). No word yet on whether this is acceptable off bucket boundaries, but we'll have to move forward with the assumption it is.

fbertsch:
I think this is complete. There are two others that I don't need to include here:
1. Mean Startup Time - using `firstPaint` (included), only when `subsession_counter == 1`
2. Crashes - All kinds already included

@maurodoglio can you verify my isQuantumReady field? We should have the same definition (even use the same code, if possible!)

PR #225, opened 2017-05-24
Allow adding experiment as a longitudinal column

sunahsuh:
Adds another CLI option to pass in an experiment_id, which should allow us to create a dataset of this shape to feed into the analysis job.


sunahsuh:
Ah, damn, I have to make a few more changes to make this work with the folder schema for the new experiment source. I'll reopen this when it's actually ready.

PR #227, opened 2017-05-25
Automatically add hists to main_summary

fbertsch:
Currently this only includes Linear and Exponential opt-out hists.
The format is essentially exactly what we recieve from the client,
see the case class for the schema.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
@@ -1,8 +1,12 @@
 package com.mozilla.telemetry.utils
 
 import com.mozilla.telemetry.metrics._
-import org.apache.spark.sql.Row
+import org.json4s.jackson.JsonMethods._
+import org.json4s.DefaultFormats
 import org.json4s.JsonAST._
+import org.apache.spark.sql.Row
+
+import scala.util.{Success, Failure, Try}
 
 case class Addon(id: Option[String],
                  blocklisted: Option[Boolean],
@@ -308,5 +312,42 @@ object MainPing{
     } catch {
       case e: Exception => None
     }
+  }
+
+  def extractHistogramMap(histogram: JValue): Map[Integer, Integer] = {
+    implicit val formats = DefaultFormats
+
+    Try((histogram \ "values").extract[Map[Integer, Integer]]) match {
+      case Success(h) => h
+      case _ => null
+    }
+  }
+
+  def histogramsToRow(histograms: JValue, definitions: List[(String, HistogramDefinition)]): Row = {
+    implicit val formats = DefaultFormats
+
+    val values = definitions.map{
+      case (name, definition) =>
+        definition match {
+          case LinearHistogram(keyed, low, high, nBuckets) => (name, keyed)
+          case ExponentialHistogram(keyed, low, high, nBuckets) => (name, keyed)
+          case EnumeratedHistogram(keyed, _) => (name, keyed)
+          case BooleanHistogram(keyed) => (name, keyed)
+          case other =>
+            throw new UnsupportedOperationException(s"${other.toString()} histogram types are not supported")
+        }
+    }.map{
+      case (name, keyed) =>
+        keyed match {
+          case true =>
+            Try((histograms \ name).extract[Map[String,JValue]].mapValues(extractHistogramMap _).map(identity)) match {
+              case Success(keyedHistogram) => keyedHistogram
+              case _ => null
+          }
+          case false => extractHistogramMap(histograms \ name)
+        }
+    }
+
+    Row.fromSeq(values)
   }
 }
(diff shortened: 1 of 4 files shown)


src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
```diff
@@ -25,6 +30,35 @@ case class Attribution(source: Option[String],
...
+      Histogram.toOption[Long](row, 4)
+    )
+  }
+}
+
+case class Histogram(range: Option[List[Int]],
+                     bucketCount: Option[Int],
+                     histogramType: Option[Int],
+                     values: Option[Map[Int, Int]],
+                     sum: Option[Long]){
```
sunahsuh: - I'm a bit concerned that if the schema hews too closely to incoming data, we're going to have disparate schemas for all the histograms again
- I wonder if we should do a bit more data validation here (e.g. does it make sense to allow these fields to be Options?)

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -630,13 +646,33 @@ object MainSummaryView {
...
+          case other =>
+            throw new UnsupportedOperationException(s"${other.toString()} histogram types are not supported")
+        }
+    }.map{
+      case (name, keyed) =>
+        keyed match {
+          case true => StructField(name.toLowerCase, MapType(StringType, HistogramSchema), nullable = true)
+          case false => StructField(name.toLowerCase, HistogramSchema, nullable = true)
+        }
+    }
```
fbertsch: I think this is correct? Happy to fix if I'm wrong.

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
fbertsch: I waffled on the data validation question. Originally I had it more strict, where if any field was missing/incorrect all would be `null`. Now it tries it's best to munge the data to match the class - you can see the last test is for the validation case, where only one element is present. My rationale is that the [schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/main/main.4.schema.json#L797) does not require any of them, so we do allow histograms missing fields, e.g. `bucket_count`.

fbertsch: Can you describe the "disparate schemas" problem? I don't think I'm familiar.

sunahsuh: > My rationale is that the schema does not require any of them, so we do allow histograms missing fields, e.g. bucket_count.
Okay, that sounds fine to me
> Can you describe the "disparate schemas" problem? I don't think I'm familiar.
Hmm, taking a closer look at the actual data coming in and it looks like everything adheres to the same schema (I was under the impression the other histogram types came in different shapes) -- if that's the case, why not add all the histograms?

fbertsch: @sunahsuh I'm in full agreement. I see no reason not to add all opt-out histograms before the backfill.

sunahsuh: Chatted briefly with @vitillo about this, and I think the only concern is count histograms, in that we should probably turn them into scalars

vitillo: For the MVP we could ignore count histogram, which ideally should be moved to scalars on the client side.

sunahsuh:
I just tried running this on yesterday's nightly data and ran @fbertsch's good friend the Constant Pool Error.

fbertsch:
@sunahsuh it looks like we have plenty of wiggle room here. We can include up to 80% of opt-out histograms and it still runs without erroring out.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -1,5 +1,6 @@
 package com.mozilla.telemetry.views
 
+import org.apache.spark.sql.catalyst.ScalaReflection
```
sunahsuh: This is now unused

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -74,7 +86,11 @@ object MainSummaryView {
         println(s" Filtering for version = '${filterVersion.get}'")
 
       val scalarDefinitions = Scalars.definitions(includeOptin = true).toList.sortBy(_._1)
-      val schema = buildSchema(scalarDefinitions)
+
+      // for now, just linear and exponential opt-out histograms
```
sunahsuh: This comment is out of date with the latest changes

fbertsch:
@sunahsuh RFAL

sunahsuh:
👍 :shipit:

PR #231, opened 2017-05-30
Allow overwrite for main_summary

fbertsch:
I'll have a new airflow job just for this backfill.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -42,25 +42,25 @@ object MainSummaryView {
     }
 
     // Set up Spark
-    val sparkConf = new SparkConf().setAppName(jobName)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -127,7 +128,7 @@ object MainSummaryView {
       // Then write to S3 using the given fields as path name partitions. If any
       // data already exists for the target day, cowardly refuse to run. In
       // that case, go delete the data from S3 and try again.
-      partitioned.write.partitionBy("sample_id").mode("error").parquet(s3path)
+      partitioned.write.partitionBy("sample_id").mode(conf.writeMode()).parquet(s3path)
```
mreid-moz: I think we should just change the job to `overwrite` instead of `error`. That would be more in line with the assumption that tasks are idempotent.

fbertsch:
Tested this out on the cluster, no problems.

PR #232, opened 2017-05-30
Pick out main_summary rows tagged with experiments and write them to …

sunahsuh:
…a separate dataset

Very simple job -- explodes experiment id and branch from the `experiments` field in main_summary and writes the data partitioned by experiment_id, submission_date_s3 for the sake of convenience and as a prelude to the analysis job.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
@@ -1,0 +1,88 @@
+package com.mozilla.telemetry.views
+
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.SparkSession
+import org.joda.time.{DateTime, Days, format}
+import org.rogach.scallop._
+
+object ExperimentSummaryView {
+  def schemaVersion: String = "v1"
+  def jobName: String = "experiments"
+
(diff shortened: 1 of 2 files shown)


sunahsuh:
Sigh. I'll change the printlns to info logs.

bsmedberg:
I just wanted to note something in case it affects your storage/budget: it's likely that with pref rollouts experiments will no longer be rare: everyone will end up having a few or many experiment listings all the time. Are you worried that this will lead to significant data duplication?

vitillo:
@bsmedberg I am not worried at this point as most experiments should generally be short-lived and target a small fraction of the user base. That said, we can revisit this decision (mostly painlessly) if it turns out to be a problem.

bsmedberg:
To be clear, we're going to be running controlled rollouts to 100% of release 55 using this value, so I know for a fact that this will affect basically 100% of the population come August.

vitillo:
How many "experiments" are you expecting to rollout to 100% concurrently?

bsmedberg:
Hard to crystal-ball: if the insights team has their way, nightly will have tens or more pref flip studies going on. Beta will have fewer, but still expect 100% of beta to be involved in some studies all the time.

Not so sure about release, but I think it's reasonable to expect that 100% of release will be in at least one study pretty regularly.

PR #233, opened 2017-05-30
Preserve column types and ordering on ToplineSummary

acmiyaguchi:
The `topline_dashboard` job [relies on column ordering for conversion when applying a schema](https://github.com/mozilla/python_mozetl/blob/master/mozetl/topline/topline_dashboard.py#L180). The upstream ToplineSummary should stay somewhat consistent with this schema.

This also fixes the `crashes` column, which gets saved as an IntegerType. This is incompatible with the LongType.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
@@ -327,9 +327,24 @@ object ToplineSummaryView {
         .withColumnRenamed("country", "geo")
         // Bug 1364617 removed generating crash aggregates.
         // This column defaults to 0 to maintain the schema, but it is just a dummy value.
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/views/ToplineSummaryView.scala
```diff
@@ -327,9 +327,24 @@ object ToplineSummaryView {
         .withColumnRenamed("country", "geo")
         // Bug 1364617 removed generating crash aggregates.
         // This column defaults to 0 to maintain the schema, but it is just a dummy value.
-        .withColumn("crashes", lit(0))
+        .withColumn("crashes", lit(0).cast(LongType))
```
acmiyaguchi: Yep, that's much simpler.

PR #234, opened 2017-05-31
Bug 1368886 - Fix Boolean UserPref null type

fbertsch:
Hmm, I can't ask @SamPenrose for a review. Feel free to take a look. The test fails without the fix, with the same error as bug 1368886.


SamPenrose:
Tests pass, code scans cleanly. I'm not and shouldn't be an official reviewer, but r+ from me and also thank you for the fix.

PR #235, opened 2017-06-01
Fix null type for User Prefs

fbertsch:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -214,15 +214,15 @@ object MainSummaryView {
...
+      case _ => null 
     }
     val anme = prefs \ "extensions.allow-non-mpc-extensions" match {
       case JBool(x) => x
       case _ => null
     }
     val row = Row(pc, anme)
     row match {
-      case Row(None, null) => None
+      case Row(null, null) => None
```
sunahsuh: Shouldn't the output here also be null?

fbertsch: The result is nulled, see line 530.

sunahsuh: Okay, great

PR #236, opened 2017-06-02
Experiment main summary

sunahsuh:
According to Osmose, pref rollouts *will* be a different action so this will work for our purposes

A note that I took actual output from the Normandy server for the test fixtures, but it's a public, unauthed API

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
@@ -1,0 +1,88 @@
+package com.mozilla.telemetry.views
+
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.SparkSession
+import org.joda.time.{DateTime, Days, format}
+import org.rogach.scallop._
+
+object ExperimentSummaryView {
+  def schemaVersion: String = "v1"
+  def jobName: String = "experiments"
+
+  private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val from = opt[String]("from", descr = "From submission date", required = false)
+    val to = opt[String]("to", descr = "To submission date", required = false)
+    val inputBucket = opt[String]("inbucket", descr = "Source bucket for main_summary data", required = false)
+    val outputBucket = opt[String]("bucket", descr = "Destination bucket for parquet data", required = true)
+    verify()
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args)
+    // parse command line arguments
+    val fmt = format.DateTimeFormat.forPattern("yyyyMMdd")
+    val to = conf.to.get match {
+      case Some(t) => fmt.parseDateTime(t)
+      case _ => DateTime.now.minusDays(1)
+    }
+    val from = conf.from.get match {
+      case Some(f) => fmt.parseDateTime(f)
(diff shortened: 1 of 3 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
```diff
@@ -0,0 +1,121 @@
...
+    val mainSummary = spark.read.parquet(input)
+    mainSummary
+      .select(col("*"), explode(col("experiments")).as(Array("experiment_id", "experiment_branch")))
+        .where(col("experiment_id").isin(experiments:_*))
+      .withColumn("submission_date_s3", lit(date))
+      .repartition(col("experiment_id"))
+      .write
+      // Usage characteristics will most likely be "get all pings from an experiment for all days"
+      .partitionBy("experiment_id", "submission_date_s3")
+      .mode("error")
```
fbertsch: I would change this to `overwrite` to enable easy backfilling.

fbertsch:
@sunahsuh I'm okay with not fixing the minor nits, but I am worried about the output partioning. If we can we should test on multiple days.

src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
```diff
@@ -0,0 +1,121 @@
...
+      .parquet(output)
```
sunahsuh: Hmm, yeah, good catch -- so the reason I chose experiment_id first is that the usage characteristics are almost always going to be "grab all the pings from this experiment". I switched the write mode to append which will allow for experiment-first partitioning, but this poses a risk where duplicate runs will silently write duplicate outputs. If we do end up running into this issue, I have a few ideas re: mitigation, but I'd like to see that we run into this first.

src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
```diff
@@ -0,0 +1,121 @@
...
+      .getOrCreate()
+
+    val parquetSize = (512 * 1024 * 1024).toLong
+    spark.conf.set("parquet.block.size", parquetSize)
+    spark.conf.set("dfs.blocksize", parquetSize)
+    return spark
+  }
+
+  def writeExperiments(input: String, output: String, date: String, experiments: List[String], spark: SparkSession): Unit = {
+    val mainSummary = spark.read.parquet(input)
```
sunahsuh: Semantically, this would remove all mention Main Summary from this method, which I think actually makes it a bit harder to understand.

sunahsuh:
RFAL

fbertsch:
@sunahsuh re: duplicate outputs, how will we catch that?

sunahsuh:
My current thinking is that since we already have a list of experiments that we care about, it wouldn't be too hard to check for an existing output directory for each experiment_id/date and delete the extant directories before exploding/writing.

fbertsch:
@sunahsuh, do you want to add that here? Or were you planning on doing that in bash?

sunahsuh:
I was going to skip it unless we knew it'd become a problem, but I guess the context has changed since I've opened this PR, if we're adding this to airflow before the main_summary backfill (although, there aren't that many days that we should have output for this job). I'll write that up in a jiffy and ping here when it's updated.

sunahsuh:
Okay, added a deletion step to the job; RFAL

fbertsch:
@sunahsuh have you tested the new deletion code?

sunahsuh:
Yeah -- worked on the test bucket output I'd already run

PR #237, opened 2017-06-05
Move Histograms and Scalars under metrics package with common MetricD…

sunahsuh:
…efinition ancestor

I've found myself working around this issue where I want to access the "keyed" value in both HistogramDefinitions and ScalarDefinitions yet it's not accessible from the base class. This fixes that issue and moves Histograms.scala and Scalars.scala under a common package, which I feel makes more sense.

Tests pass, and I checked moztelemetry to make sure there aren't references to these files, but if there are other external dependencies that I don't know about please let me know.


fbertsch:
Nice, this is a great idea. I was running into a similar situation.

PR #239, opened 2017-06-05
Avoid a warning about unreachable code.

mreid-moz:


mreid-moz:
@acmiyaguchi *nudge*

mreid-moz:
The `case v => v \ bucket match {` line had no type-specific match criteria, it would match any value for `v`, which is effectively the same as the `case _` case, so the `case _` was unreachable. Adding the `v: JValue` criteria means that anything that's not a `JValue` would fall through to the `case _` case.

PR #241, opened 2017-06-06
Fix main_summary running on old data

fbertsch:
Backfill was failing because of these.


fbertsch:
@mreid-moz, I haven't had a chance to test this. I don't think I'll have time today but can do it tomorrow morning.

fbertsch:
okay, have confirmed this works as expected.

PR #242, opened 2017-06-07
Add working histogram aggregation job

sunahsuh:
Much smaller PR than last time: this is the first iteration of the experiment metric aggregation and analysis job based on Main Summary and Datasets.

Currently, this only aggregates the whitelisted histograms in main summary. The next PR will include scalar aggregation, and the one after that probably summary stats.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
@@ -1,0 +1,93 @@
+package com.mozilla.telemetry.experiments.analyzers
+
+import com.mozilla.telemetry.metrics.HistogramDefinition
+import org.apache.spark.sql._
+import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.expressions.Aggregator
+
+import scala.collection.{Map => CMap}
+import scala.util.{Failure, Success, Try}
+
+
+case class HistogramRow(experiment_id: String, branch: String, subgroup: String, metric: Option[CMap[Int, Int]])
+case class KeyedHistogramRow(experiment_id: String, branch: String, subgroup: String,
+                             metric: Option[CMap[String, CMap[Int, Int]]]) {
+  def collapsedMetric: Option[CMap[Int, Int]] = {
+    // The OKRs state that we collapse keyed metrics (for now)
+    metric match {
+      case Some(m) => Some(m.values.reduce(sumCMaps[Int](_: CMap[Int, Int], _: CMap[Int, Int])))
+      case _ => None
+    }
+  }
+
+  def toHistogramRow: HistogramRow = {
+    HistogramRow(experiment_id, branch, subgroup, collapsedMetric)
+  }
+}
+
+object HistogramAggregator
+  extends Aggregator[HistogramRow, Map[Int, Long], Map[Long, HistogramPoint]] {
+  def zero: Map[Int, Long] = Map[Int, Long]()
+
+  def reduce(b: Map[Int, Long], h: HistogramRow): Map[Int, Long] = {
+    h.metric match {
+      case Some(m: CMap[Int, Int]) => b ++ m.map { case (k: Int, v: Int) => k -> (v.toLong + b.getOrElse(k, 0L)) }
+      case _ => b
+    }
+  }
+
+  def merge(l: Map[Int, Long], r: Map[Int, Long]): Map[Int, Long] = {
+    l ++ r.map { case (k, v) => k -> (v + l.getOrElse(k, 0L)) }
+  }
+
+  def finish(b: Map[Int, Long]): Map[Long, HistogramPoint] = {
+    val sum = b.values.sum
+    b.map { case (k, v) => k.toLong -> HistogramPoint(v.toDouble / sum.toDouble, v.toDouble, None) }
+  }
+
+  def bufferEncoder: Encoder[Map[Int, Long]] = ExpressionEncoder()
+  def outputEncoder: Encoder[Map[Long, HistogramPoint]] = ExpressionEncoder()
+}
+
+class HistogramAnalyzer(name: String, hd: HistogramDefinition, df: DataFrame) extends java.io.Serializable {
+  import df.sparkSession.implicits._
+
+  def analyze(): Option[Dataset[HistogramAnalysis]] = {
+    val ds = format match {
+      case Some(d: DataFrame) => collapseKeys(d)
+      case _ => return None
+    }
+
+    val histogramAggregate = HistogramAggregator.toColumn.name("histogram_aggregate")
+    val output = ds.filter(r => r.metric.isDefined)
+      .groupByKey(x => MetricKey(x.experiment_id, x.branch, x.subgroup))
+      .agg(histogramAggregate, count("*"))
+      .map { toOutputSchema }
+
+    Some(output)
+  }
+
+  private def format: Option[DataFrame] = {
+    Try(df.select(
+      col("experiment_id"),
+      col("experiment_branch").as("branch"),
+      lit("All").as("subgroup"),
+      col(name).as("metric"))
+    ) match {
+      case Success(x) => Some(x)
+      // expected failure, if the dataset doesn't include this metric (e.g. it's newly added)
+      case Failure(_: org.apache.spark.sql.AnalysisException) => None
+      case Failure(x: Throwable) => throw x
+    }
+  }
+
+  private def collapseKeys(formatted: DataFrame): Dataset[HistogramRow] = {
+    if (hd.keyed) formatted.as[KeyedHistogramRow].map(_.toHistogramRow) else formatted.as[HistogramRow]
+  }
+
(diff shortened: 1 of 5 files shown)


src/test/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzerTest.scala
```diff
@@ -0,0 +1,82 @@
...
+  import spark.implicits._
+
+  val m1 = CMap(0 -> 1, 1 -> 2, 2 -> 3)
+  val m2 = CMap(0 -> 2, 1 -> 4, 2 -> 6)
+  val m3 = CMap(0 -> 1, 1 -> 2, 2 -> 3, 100 -> 1)
+  val m4 = CMap(0 -> 1, 1 -> 2, 2 -> 3, 5 -> 0)
+
+  val df = Seq(
+    ExperimentDataset("experiment1", "control", Some(m1), Some(CMap("key1" -> m1, "key2" -> m2))),
+    ExperimentDataset("experiment1", "control", None, None),
```
fbertsch: I think these should be `null` instead.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,116 @@
...
+        .where(col("submission_date_s3") === date)
+        .select("experiment_id")
+        .distinct()
+        .collect()
+        .toList
+        .map(r => r(0).asInstanceOf[String])
+    }
+
+    val histogramList = conf.histo.get match {
+      case Some(h) => Map(h -> Histograms.definitions()(h.toUpperCase))
```
sunahsuh: > Is this compiling? The return type of filterHistogramDefinitions is List[(String, HistogramDefinition)], while this is a Map[String, HistogramDefinition].

Yeah, they both mix in the Iterable trait, which is all you need for `.map` -- I'll change this to a list for clarity though.

> Also, do we need to check that the histogram definition is indeed available in the data?
This is done in the analysis part (which is why the return type of `.analyze` is an Option)

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.spark.sql.expressions.Aggregator
+
+import scala.collection.{Map => CMap}
+import scala.util.{Failure, Success, Try}
+
+
+case class HistogramRow(experimentId: String, branch: String, subgroup: String, metric: Option[CMap[Int, Int]])
+case class KeyedHistogramRow(experimentId: String, branch: String, subgroup: String,
+                             metric: Option[CMap[String, CMap[Int, Int]]]) {
+  def collapsedMetric: Option[CMap[Int, Int]] = {
```
sunahsuh: Yeah, this was a question that came up earlier and we were told just collapse keyed histograms.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  def toHistogramRow: HistogramRow = {
+    HistogramRow(experimentId, branch, subgroup, collapsedMetric)
+  }
+}
+
+case class HistogramKey(experimentId: String, branch: String, subgroup: String)
+case class HistogramPoint(pdf: Double, count: Double, label: Option[String])
```
sunahsuh: This came from the way the graphs in Distribution Viewer were discussed as `cdf`s which isn't strictly true either, but it appears that's a commonly understood way of discussing it here. I'm not hugely opposed to changing it but it *would* require changes downstream so I'd prefer not to, unless there's a consensus about it being confusing.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -0,0 +1,114 @@
...
+                     value: Double,
+                     confidence_low: Double,
+                     confidence_high: Double,
+                     confidence_level: Double,
+                     p_value: Double)
+
+case class HistogramAnalysis(experiment_id: String,
+                             experiment_branch: String,
+                             subgroup: String,
+                             n: Long,
```
sunahsuh: > So the definition is: number of pings seen, with experiment_id, experiment_branch, and subgroup, with non-null values for n. Correct?

Yes

> Eventually we should include a probabilistic number of users along these dimensions.

Do you mean a count of client IDs? If so, I do have that as a potential item to compute at the end (also, I don't think we'd need to make it probabilistic)

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -0,0 +1,114 @@
...
+  import df.sparkSession.implicits._
+
+  def analyze(): Option[Dataset[HistogramAnalysis]] = {
+    val ds = format match {
+      case Some(d: DataFrame) => collapseKeys(d)
+      case _ => return None
+    }
+
+    val histogramAggregate = HistogramAggregator.toColumn.name("histogram_aggregate")
+    val output = ds.filter(r => r.metric != None)
```
sunahsuh: No, because this uses the Dataset api, `r` is of type `HistogramRow` at this point and r.metric is an Option

src/main/scala/com/mozilla/telemetry/experiments/analyzers/package.scala
```diff
@@ -0,0 +1,10 @@
+package com.mozilla.telemetry.experiments
+
+import scala.collection.{Map => CMap}
+
+
+package object analyzers {
+  def sumCMaps[T](l: CMap[T, Int], r: CMap[T, Int]): CMap[T, Int] = {
```
sunahsuh: At the moment, yes, but we'll need this for scalar aggregation

src/test/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzerTest.scala
sunahsuh: Similar thing as above -- we're going from case class -> dataset -> dataframe

src/main/scala/com/mozilla/telemetry/utils/package.scala
```diff
@@ -92,4 +92,8 @@ package object utils{
     println(s"Elapsed time: ${(t1 - t0)/1000000000.0} s")
     result
   }
+
+  def todayAsYYYYMMDD: String = {
```
sunahsuh: Ehhh words ;)

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
sunahsuh: > Also, do we need to check that the histogram definition is indeed available in the data?

Oh, I misunderstood that -- no, I think throwing an exception is the right thing to do here

sunahsuh:
Whew, okay, tests pass and feedback addressed

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
mreid-moz: A clarifying comment might suffice for now (and it would be good to mention the downstream consequences of changes here as well).

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
mreid-moz: nit: `r.metric.nonEmpty`?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -0,0 +1,116 @@
...
+    StructField("p_value", DoubleType, nullable = true)
+  ))
+
+  def buildHistogramPointSchema = StructType(List(
+    StructField("pdf", DoubleType),
+    StructField("count", LongType),
+    StructField("label", StringType, nullable = true)
+  ))
+
+  def buildOutputSchema = StructType(List(
```
mreid-moz: Are these schemas used for anything?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -0,0 +1,114 @@
...
+      case _ => None
+    }
+  }
+
+  def toHistogramRow: HistogramRow = {
+    HistogramRow(experimentId, branch, subgroup, collapsedMetric)
+  }
+}
+
+case class HistogramKey(experimentId: String, branch: String, subgroup: String)
```
sunahsuh: I'm torn between Scala's snakecase convention and parquet's mixed case issues :( I should probably just go with the downcase version everywhere though

sunahsuh:
RFAL

PR #243, opened 2017-06-08
Use Try.toOption instead of toInt

maurodoglio:
The Try.toOption syntax is very readable and it's less code to maintain.


fbertsch:
@maurodoglio, what about `MainSummaryView.scala`?

maurodoglio:
I rebased on top of master and I can't find any occurrences of `toInt()` left.
RFAL

PR #246, opened 2017-06-12
Fix isQuantumReady definition in main_summary

fbertsch:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -215,42 +215,57 @@ object MainSummaryView {
...
       "firefox-compact-light@mozilla.org",
       "firefox-compact-dark@mozilla.org"
     )
 
     val allowedTheme = getTheme(theme) match {
       case Some(t) => t.get(0) match {
-        case id: String => !disallowedThemes.contains(id)
-        case _ => false
+        case id: String => id match {
+            case "MISSING" => None
```
maurodoglio: What's this "MISSING"?

src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
```diff
@@ -1790,16 +1790,54 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
     )
 
-    val json3theme = parse("""{"id": "okayid"}""")
+    val json3theme = parse("""{"id": "firefox-compact-light@mozilla.org"}""")
 
-    MainSummaryView.getQuantumReady(json3e10s, json3addons, json3theme) should be (false)
+    MainSummaryView.getQuantumReady(json3e10s, json3addons, json3theme) should be (Some(false))
 
-    // not quantum ready with old-style theme
+    // not quantum-ready with old-style addon
```
maurodoglio: s/addon/theme

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
fbertsch: From the `getTheme` function. Not a super big fan, if you think there's a better refactor I'm all ears.

maurodoglio: Why doesn't getTheme return an Option? If "MISSING" is the default value of a column, we should probably set its value to theme.getOrElse("MISSING").

fbertsch:
@maurodoglio fixed the existing isQuantumReady definition, and changed all `GC_MAX_PAUSE` to `GC_MAX_PAUSE_2`. RFAL

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
maurodoglio: discussed this offline, it's not something we should address in this PR

PR #247, opened 2017-06-13
Add partition option to Generic Count view

fbertsch:
Instead of reprocessing all data every week we want to update the Quantum RC, having it partitioned means that I can update it weekly with only the most recent data, assuming the schema doesn't change. Eventually we can do this with the other counts as well.

This shouldn't affect end users as all queries are using `WHERE` and `GROUP BY`, which will still be fine with the partitions, even if two rows in separate partitions have the same values for all dimensions.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala
@@ -14,6 +14,7 @@ object GenericCountView {
   val DefaultSubmissionDateCol = "submission_date_s3"
   val DefaultHllBits = 12
   val DefaultOutputFiles = 32
+  val DefaultWriteMode = "overwrite"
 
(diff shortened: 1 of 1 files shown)


mreid-moz:
Please rebase on latest master

src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala
```diff
@@ -71,6 +72,15 @@ object GenericCountView {
       "version",
       descr = "Version of the output data. Defaults to v<from><to>",
       required = false)
+    val partition = opt[String](
```
mreid-moz: The name `partition` is a bit confusing here, considering the grouping columns/dimensions. Maybe something like `outputPrefix` or `outputPartition` would be clearer.

fbertsch:
@mreid-moz also added a write mode option. RFAL

PR #249, opened 2017-06-16
Add TLS Histograms to main_summary

fbertsch:
bug 1373645, for EKR

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -21,12 +21,18 @@ object MainSummaryView {
   def jobName: String = "main_summary"
 
   val histogramsWhitelist =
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -26,7 +26,13 @@ object MainSummaryView {
     "CYCLE_COLLECTOR_MAX_PAUSE" ::
     "INPUT_EVENT_RESPONSE_COALESCED_MS" ::
     "GHOST_WINDOWS" ::
-    "HTTP_PAGELOAD_IS_SSL" :: Nil
+    "HTTP_PAGELOAD_IS_SSL" ::
```
mreid-moz: Per IRC discussion, if everything is fine in terms of Schema Evolution, I would prefer to sort all the histograms alphabetically here so that it's easier to tell if a desired histogram is in the list.

PR #250, opened 2017-06-18
Add Quantum RC Job

fbertsch:
Phew. Originally I was hoping to use the `GenericCountView` for this, but it soon became apparent that I was trying to conform the problem to fit that solution when it clearly did not. I next had a complicated solution using crazy `DataSet.pivot`s, but realized it was too complex. This is indeed the correct solution for the problem at hand, and we should probably push these new `HyperLogLogMerge` functions to Sonotype.


src/test/scala/com/mozilla/telemetry/views/QuantumRcViewTest.scala
```diff
@@ -0,0 +1,240 @@
+package com.mozilla.telemetry
```
mreid-moz: Filename should be `QuantumRCViewTest` for consistency

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -157,3 +160,102 @@ class CollectList(inputStruct: StructType, orderCols: List[String], maxLength: O
...
+object UDFs{
+  val HllCreate = "hll_create"
+  val HllCardinality = "hll_cardinality"
+  val HllMerge = new HyperLogLogMerge
+  val FilteredHllMerge = new FilteredHyperLogLogMerge
+
+  implicit class MozUDFs(spark: SparkSession) {
+    def registerUDFs: Unit = {
+      spark.udf.register(HllCreate, hllCreate _)
+      spark.udf.register(HllCardinality, hllCardinality _)
```
fbertsch: I didn't have any need for them

src/main/scala/com/mozilla/telemetry/views/QuantumRCView.scala
```diff
@@ -0,0 +1,137 @@
...
+    "input_jank_content_over_2500" -> "input_event_response_coalesced_ms_content_above_2500",
+    "input_jank_content_over_250" -> "input_event_response_coalesced_ms_content_above_250",
+    "subsession_length" -> "subsession_length")
+
+  def aggregate(df: DataFrame): DataFrame = {
+    import df.sparkSession.implicits._
+
+    val withExperiments = df
+      .where("client_id IS NOT NULL")
+      .select(col("*"),
```
fbertsch: Spark will properly optimize this, I'd prefer to leave it as-is since it's more readable.

fbertsch:
@mreid-moz I'm not sure about deduping - I'm betting it will be expensive to dedupe 100% of our data in Spark. Presumably this shouldn't affect client counts either, only sums.

It probably would be good to do some analysis to see if this is necessary first.
Either way, RFAL.

src/main/scala/com/mozilla/telemetry/views/QuantumRCView.scala
mreid-moz: Sounds good

PR #251, opened 2017-06-19
Add scalar aggregation to experiment analysis job

sunahsuh:
Creates a few underlying base classes for generalized metric analysis, modifies the histogram analyzer to use this, and adds scalar aggregation as well.

I had a version earlier that just used parameterized classes for the scalar datasets (without the concrete type versions) but those don't seem to play well with Datasets, hence there's more boilerplate than I'd like :(


src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,86 @@
...
+      .map(toOutputSchema)
+
+    reindex(output)
+  }
+
+  private def format: Option[DataFrame] = {
+    Try(df.select(
+      col("experiment_id"),
+      col("experiment_branch").as("branch"),
+      lit("All").as("subgroup"),
```
mreid-moz: Should we add a comment here indicating that this is a placeholder value until subgroups are implemented?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+trait KeyedScalarRow[T] {
+  val experiment_id: String
+  val branch: String
+  val subgroup: String
+  val metric: Option[Map[String, T]]
+
+  def collapsedMetric: Option[Map[T, Long]] = {
+    metric match {
+      case Some(m: Map[String, T]) => Some(m.values.foldLeft(Map.empty[T, Long]) {
+        case(m: Map[T, Long], e: T @unchecked) => m + (e -> (m.getOrElse(e, 0L) + 1L))
```
mreid-moz: I don't quite understand this bit - are we counting the occurrences of each value observed for each key?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  private def addStringHistograms(l: Map[String, Long], r: Map[String, Long]): Map[String, Long] = {
+    l ++ r.map { case (k, v) => k -> (v + l.getOrElse(k, 0L)) }
+  }
+
+  override protected def reindex(aggregates: Dataset[MetricAnalysis]): Dataset[MetricAnalysis] = {
+    import aggregates.sparkSession.implicits._
+    // this is really annoying, but we need to give string scalar values indexes and they have to be
```
mreid-moz: Ew, why do we need to do this?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -2,92 +2,58 @@ package com.mozilla.telemetry.experiments.analyzers
...
-      lit("All").as("subgroup"),
-      col(name).as("metric"))
-    ) match {
-      case Success(x) => Some(x)
-      // expected failure, if the dataset doesn't include this metric (e.g. it's newly added)
-      case Failure(_: org.apache.spark.sql.AnalysisException) => None
-      case Failure(x: Throwable) => throw x
+  implicit class KeyedHistogramMetric(val metric: Option[Map[String, Map[Int, Int]]]) {
+    private def sumKeys(l: Map[Int, Int], r: Map[Int, Int]): Map[Int, Int] = {
+      l ++ r.map { case (k, v) => k -> (v + l.getOrElse(k, 0)) }
```
mreid-moz: This is very similar to the `add` function defined earlier - it may not be worthwhile for such a small amount of code, but consider adding a common util function for merging maps.

src/test/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzerTest.scala
```diff
@@ -2,35 +2,36 @@ package com.mozilla.telemetry.experiments.analyzers
...
+import org.apache.spark.sql.DataFrame
 import org.scalatest.{FlatSpec, Matchers}
 
-import scala.collection.{Map => CMap}
+import scala.collection.Map
 
 
-case class ExperimentDataset(experiment_id: String,
+case class HistogramExperimentDataset(experiment_id: String,
                              experiment_branch: String,
```
mreid-moz: nit: Please fix indentation here

src/test/scala/com/mozilla/telemetry/experiments/analyzers/ScalaAnalyzerTest.scala
```diff
@@ -0,0 +1,192 @@
+package com.mozilla.telemetry.experiments.analyzers
```
mreid-moz: Filename should be `ScalarAnalyzerTest.scala`

src/test/scala/com/mozilla/telemetry/experiments/analyzers/ScalaAnalyzerTest.scala
```diff
@@ -0,0 +1,192 @@
...
+import com.mozilla.telemetry.metrics._
+import org.apache.spark.sql.DataFrame
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.collection.Map
+
+
+case class ScalarExperimentDataset(experiment_id: String,
+                                   experiment_branch: String,
+                                   uint_scalar: Option[Int],
```
mreid-moz: Should we add a placeholder `subgroup` field here?

src/test/scala/com/mozilla/telemetry/experiments/analyzers/ScalaAnalyzerTest.scala
```diff
@@ -0,0 +1,192 @@
...
+      MetricAnalysis("experiment1", "control", "All", 3L, "keyed_string_scalar", "StringScalar",
+        stringsToPoints(List((0, "hello", 3), (1, "world", 3))), None),
+      MetricAnalysis("experiment1", "branch1", "All", 1L, "keyed_string_scalar", "StringScalar",
+        stringsToPoints(List((0, "hello", 1), (1, "world", 1))), None),
+      MetricAnalysis("experiment1", "branch2", "All", 2L, "keyed_string_scalar", "StringScalar",
+        stringsToPoints(List((0, "hello", 2), (1, "world", 2))), None)
+    )
+    assert(actual == expected)
+  }
+}
```
mreid-moz: Do we need a test that shows that we can produce a dataset containing both Scalar and Histogram results together?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+}
+
+class UintScalarAnalyzer(name: String, md: ScalarDefinition, df: DataFrame)
+  extends MetricAnalyzer[Int](name, md, df) {
+  override type PreAggregateRowType = UintScalarMapRow
+  val aggregator = UintAggregator
+
+  def collapseKeys(formatted: DataFrame): Dataset[UintScalarMapRow] = {
+    import df.sparkSession.implicits._
+    if (md.keyed) {
```
fbertsch: Nit: I would prefer to see ternary operator to get the `KeyedUintScalarRow` or `UintScalarRow` on one line, then another line to return the `Dataset` with new types. Same for other `collapseKeys` functions.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+      formatted.as[KeyedStringScalarRow].map(_.toScalarMapRow)
+    } else {
+      formatted.as[StringScalarRow].map(_.toScalarMapRow)
+    }
+  }
+
+  private def addStringHistograms(l: Map[String, Long], r: Map[String, Long]): Map[String, Long] = {
+    l ++ r.map { case (k, v) => k -> (v + l.getOrElse(k, 0L)) }
```
fbertsch: Also similar to the `add` function that mreid mentioned earlier. I wonder if there's a way to merge all these similar `map add` functions together.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+
+  override protected def reindex(aggregates: Dataset[MetricAnalysis]): Dataset[MetricAnalysis] = {
+    import aggregates.sparkSession.implicits._
+    // this is really annoying, but we need to give string scalar values indexes and they have to be
+    // consistent among all the histograms across all branches, so we: aggregate the histograms
+    // across all the branches, sort by count descending, and use that order for our index
+    val counts = aggregates.collect().map { a: MetricAnalysis =>
+      a.histogram.values.map {p: HistogramPoint => p.label.get -> p.count.toLong}.toMap[String, Long]
+    }
+    if (counts.isEmpty) return aggregates
```
fbertsch: Why not pattern match on `counts.isEmpty`, then we can eliminate the early return.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -0,0 +1,152 @@
...
+  private def addStringHistograms(l: Map[String, Long], r: Map[String, Long]): Map[String, Long] = {
+    l ++ r.map { case (k, v) => k -> (v + l.getOrElse(k, 0L)) }
+  }
+
+  override protected def reindex(aggregates: Dataset[MetricAnalysis]): Dataset[MetricAnalysis] = {
+    import aggregates.sparkSession.implicits._
+    // this is really annoying, but we need to give string scalar values indexes and they have to be
+    // consistent among all the histograms across all branches, so we: aggregate the histograms
+    // across all the branches, sort by count descending, and use that order for our index
+    val counts = aggregates.collect().map { a: MetricAnalysis =>
```
fbertsch: Any risk collecting the data here of it being too large? I don't *think* so since this is after aggregation.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,86 @@
...
+                     p_value: Double)
+
+case class MetricAnalysis(experiment_id: String,
+                          experiment_branch: String,
+                          subgroup: String,
+                          n: Long,
+                          metric_name: String,
+                          metric_type: String,
+                          histogram: Map[Long, HistogramPoint],
+                          statistic: Option[Seq[Statistic]])
```
fbertsch: Nit: should be `statistics`?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
```diff
@@ -0,0 +1,86 @@
...
+abstract class MetricAnalyzer[T](name: String, md: MetricDefinition, df: DataFrame) extends java.io.Serializable {
+  type PreAggregateRowType <: PreAggregateRow[T]
+  val aggregator: MetricAggregator[T]
+
+  import df.sparkSession.implicits._
+
+  def analyze(): Dataset[MetricAnalysis] = {
+    val ds = format match {
+      case Some(d: DataFrame) => collapseKeys(d)
+      case _ => return df.sparkSession.emptyDataset[MetricAnalysis]
```
fbertsch: Early returns make me uncomfortable. Would this work to just let an empty `Dataset` flow through the rest of the code below (i.e. make `ds` empty)?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/Aggregators.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+object BooleanAggregator extends MetricAggregator[Boolean] {
+  def finish(b: Map[Boolean, Long]): Map[Long, HistogramPoint] = {
+    val sum = b.values.sum.toDouble
+    if (sum == 0) return Map.empty[Long, HistogramPoint]
+    val f = b.getOrElse(false, 0L)
+    val t = b.getOrElse(true, 0L)
+
+    Map(0L -> HistogramPoint(f/sum, f, Some("False")), 1L -> HistogramPoint(t/sum, t, Some("True")))
```
fbertsch: in `UintAggregator` and `StringAggregator` the value is cast to a `double`, but here it is kept as a `long`. Any reason why?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/Aggregators.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+object UintAggregator extends MetricAggregator[Int] {
+  def finish(b: Map[Int, Long]): Map[Long, HistogramPoint] = {
+    val sum = b.values.sum.toDouble
+    if (sum == 0) return Map.empty[Long, HistogramPoint]
+    b.map { case (k: Int, v) => k.toLong -> HistogramPoint(v.toDouble / sum, v.toDouble, None) }
+  }
+
+  def bufferEncoder: Encoder[Map[Int, Long]] = ExpressionEncoder()
```
sunahsuh: As we discussed in person, the short answer is no, this is one of the reasons the concrete classes needed to be created :(

src/main/scala/com/mozilla/telemetry/experiments/analyzers/MetricAnalyzer.scala
sunahsuh: I'm gonna go with no since at the moment, the subgroup of "All" is an accurate value

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
sunahsuh: Yes, we're creating a histogram of scalar values

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
sunahsuh: Basically, so we can graph these consistently among all the branches :/ That is, experiment viewer needs to know how to graph the histograms we send them in a way that's comparable, so we need to give each string a consistent index

src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
sunahsuh: Agreed, and I did have it broken out earlier in the package object but unfortunately Scala doesn't have a shared `Numeric` ancestor for `Int` and `Long`, which means these would have to be separate functions anyway

src/test/scala/com/mozilla/telemetry/experiments/analyzers/ScalaAnalyzerTest.scala
sunahsuh: I don't think so -- we'll never have a "subgroup" field in the incoming dataset, it's something we'll have to generate

src/test/scala/com/mozilla/telemetry/experiments/analyzers/ScalaAnalyzerTest.scala
sunahsuh: I'm going to punt on this one for now because I think the actual problem here is that we don't have very testable view code in general and undertaking this correctly is a larger project 🏈

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
sunahsuh: If we have an issue here we'd probably also have an issue during aggregation, since there's only as many rows as there are branches.

sunahsuh:
Appreciate the reviews -- RFAL now

sunahsuh:
Forgot to link to the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1375017

PR #252, opened 2017-06-20
Add all process histograms and scalars to main_summary

fbertsch:
This does two things:
1. Adds all process histograms for any histograms in the whitelist
2. Adds all process scalars

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
+++ b/src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
@@ -4,18 +4,19 @@ import org.json4s._
 import org.json4s.jackson.JsonMethods._
 import scala.collection.mutable.{Map => MMap}
 import scala.io.Source
+import com.mozilla.telemetry.utils.MainPing
 
 case class RawHistogram(values: Map[String, Int], sum: Long)
 
 sealed abstract class HistogramDefinition extends MetricDefinition
-case class FlagHistogram(keyed: Boolean) extends HistogramDefinition
-case class BooleanHistogram(keyed: Boolean) extends HistogramDefinition
-case class CountHistogram(keyed: Boolean) extends HistogramDefinition
-case class EnumeratedHistogram(keyed: Boolean, nValues: Int) extends HistogramDefinition
-case class LinearHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
-case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int) extends HistogramDefinition
+case class FlagHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
+case class BooleanHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
+case class CountHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
+case class EnumeratedHistogram(keyed: Boolean, nValues: Int, processes: List[String]) extends HistogramDefinition
+case class LinearHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int, processes: List[String]) extends HistogramDefinition
+case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int, processes: List[String]) extends HistogramDefinition
 
-class HistogramsClass {
+class HistogramsClass extends MetricsClass {
   private val processTypes = List("content", "gpu")
   private case class HistogramLocation(path: List[String], suffix: String)
   private def generateLocations(histogramKey: String)(processType: String): HistogramLocation = {
@@ -110,6 +111,12 @@ class HistogramsClass {
               case _ => Some(false)
             }
             case ("releaseChannelCollection", JString(x)) => Some(x)
+            case ("record_in_processes", JArray(x)) => Some(x.map{ p =>
+              p match {
+                case JString(p) => Some(p)
+                case _ => None
+              }
+            }.toList.flatten)
             case _ => None
           }
         } catch {
@@ -139,23 +146,28 @@ class HistogramsClass {
         val high = v.getOrElse("high", None).asInstanceOf[Option[Int]]
         val nBuckets = v.getOrElse("n_buckets", None).asInstanceOf[Option[Int]]
 
+        val processes = getProcesses(
+          v.getOrElse("record_in_processes", Some(MainPing.ProcessTypes)).get
+          .asInstanceOf[List[String]]
+        )
+
         def addSuffixes(key: String, definition: HistogramDefinition): List[(String, HistogramDefinition)] = {
             suffixes.map(suffixType => (key + suffixType, definition))
         }
 
         (kind, nValues, high, nBuckets) match {
           case ("flag", _, _, _) =>
-            Some(addSuffixes(k, FlagHistogram(keyed)))
+            Some(addSuffixes(k, FlagHistogram(keyed, processes)))
           case ("boolean", _, _ , _) =>
-            Some(addSuffixes(k, BooleanHistogram(keyed)))
+            Some(addSuffixes(k, BooleanHistogram(keyed, processes)))
           case ("count", _, _, _) =>
-            Some(addSuffixes(k, CountHistogram(keyed)))
+            Some(addSuffixes(k, CountHistogram(keyed, processes)))
           case ("enumerated", Some(x), _, _) =>
-            Some(addSuffixes(k, EnumeratedHistogram(keyed, x)))
+            Some(addSuffixes(k, EnumeratedHistogram(keyed, x, processes)))
           case ("linear", _, Some(h), Some(n)) =>
-            Some(addSuffixes(k, LinearHistogram(keyed, low, h, n)))
+            Some(addSuffixes(k, LinearHistogram(keyed, low, h, n, processes)))
           case ("exponential", _, Some(h), Some(n)) =>
-            Some(addSuffixes(k, ExponentialHistogram(keyed, low, h, n)))
+            Some(addSuffixes(k, ExponentialHistogram(keyed, low, h, n, processes)))
           case _ =>
             None
         }
--- a/src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
+++ b/src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
@@ -13,11 +13,10 @@ class MetricsClass {
       _ match {
         case "main" => "parent" :: Nil
         case "all" => MainPing.ProcessTypes
-        case "all_childs" => MainPing.ProcessTypes.filter(_ != "parent")
-        case "all_child" => MainPing.ProcessTypes.filter(_ != "parent")
-        case "all_children" => MainPing.ProcessTypes.filter(_ != "parent")
+        case "all_child" | "all_childs" | "all_children" =>
+          MainPing.ProcessTypes.filter(_ != "parent")
         case o => o :: Nil
       }
-    }.toSet.toList
+    }.distinct
   }
 }
--- a/src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
+++ b/src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
@@ -1,0 +1,23 @@
+package com.mozilla.telemetry.metrics
+
+import com.mozilla.telemetry.utils.MainPing
+
+trait MetricDefinition {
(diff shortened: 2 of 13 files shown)


src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
```diff
@@ -1956,4 +2013,141 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
 
     spark.stop()
   }
+
+  "Process Histograms" can "be stored" in {
+    val spark = SparkSession.builder()
```
sunahsuh: Looks like tests are failing since this spark session is never shut down

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
```diff
@@ -36,6 +36,8 @@ case class Attribution(source: Option[String],
 case class Experiment(branch: Option[String])
 
 object MainPing{
+  val ProcessTypes = "parent" :: "content" :: "gpu" :: Nil
```
sunahsuh: Are process types only specific to the main ping? Or, even if it is, is it something that might make sense to put in `com.mozilla.telemetry.metrics` instead?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -654,47 +670,47 @@ object MainSummaryView {
...
       val histogramRow = MainPing.histogramsToRow(
-        histograms merge keyedHistograms,
+        MainPing.ProcessTypes.map{ p => p -> (histograms(p) merge keyedHistograms(p)) }.toMap,
         histogramDefinitions
       )
 
       Some(Row.merge(row, scalarRow, histogramRow))
     } catch {
-      case _: Exception =>
+      case e: Exception =>
```
sunahsuh: Does this variable need to be named?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -341,19 +341,35 @@ object MainSummaryView {
...
+        _ match {
+          case "parent" => "parent" -> parse(fields.getOrElse("payload.histograms", "{}").asInstanceOf[String])
+          case p => p -> payload \ "payload" \ "processes" \ p \ "histograms"
+        }
+      }.toMap
+
+      lazy val keyedHistograms = MainPing.ProcessTypes.map{
+        _ match {
+          case "parent" => "parent" -> parse(fields.getOrElse("payload.keyedHistograms", "{}").asInstanceOf[String])
+          case p => p -> payload \ "payload" \ "processes" \ p \ "histograms"
```
sunahsuh: Should the last xpath location be `"keyedHistograms"`? (If so, looks like we should add a test)

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -780,20 +796,22 @@ object MainSummaryView {
...
-          case true => StructField(Scalars.getParquetFriendlyScalarName(name, "parent"),
-                                   MapType(StringType, parquetType),
-                                   nullable = true)
-          case false => StructField(Scalars.getParquetFriendlyScalarName(name, "parent"),
-                                    parquetType,
-                                    nullable = true)
+            case UintScalar(keyed, processes) => (name, keyed, processes, IntegerType)
+            case BooleanScalar(keyed, processes) => (name, keyed, processes, BooleanType)
+            case StringScalar(keyed, processes) => (name, keyed, processes, StringType)
         }
```
sunahsuh: Looks like there a few four-space indents in this block?

src/test/resources/ShortHistograms.json
```diff
@@ -125,6 +138,7 @@
     "high": "100",
     "n_buckets": 10,
     "releaseChannelCollection": "opt-out",
-    "description": "a testing histogram; not meant to be touched"
+    "description": "a testing histogram; not meant to be touched",
+    "record_in_processes": ["main", "content", "gpu"]
```
sunahsuh: Would it make sense to add tests specifically for the `record_in_processes` additions to HistogramsTest as well?

src/test/scala/com/mozilla/telemetry/MainSummaryViewTest.scala
```diff
@@ -1956,4 +2013,141 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+    val sc = spark.sparkContext
+    import spark.implicits._
+
+    val message = RichMessage(
+      "2235",
+      Map(
+        "documentId" -> "foobar",
+        "submissionDate" -> "12345",
+        "submission" -> """{
+  "payload": {
```
sunahsuh: Is this indentation expected?

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
fbertsch: I think it is. Let's leave it here for now, and if it becomes useful to other components we can move it out. Plus there is the ambiguity between using "main" and "parent". I've solved this by just using just "parent".

fbertsch:
@sunahsuh made the changes you requested, and added some more tests for the missing stuff (and `keyedHistograms`). RFAL

sunahsuh:
LGTM

sunahsuh:
Uh oh, just read these docs while doing something unrelated and I think we have a problem:
```
record_in_processes: A list of processes the scalar is allowed to record in. Currently supported values are:

    main;
    content;
    gpu;
    all_child (record in all the child processes);
    all (record in all the processes).
```
from https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/collection/scalars.html

fbertsch:
@sunahsuh check out the scalar code, that was already taken care of :)

sunahsuh:
Ah, great, I'd scanned the Histograms code instead of Scalars, but `all` and `all_child` are also valid values for the histograms analogue as well: https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/collection/histograms.html#record-in-processes

fbertsch:
@sunahsuh fixed the scalars, added tests for scalars and histograms. RFAL

src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
```diff
@@ -111,6 +111,12 @@ class HistogramsClass {
               case _ => Some(false)
             }
             case ("releaseChannelCollection", JString(x)) => Some(x)
+            case ("record_in_processes", JArray(x)) => Some( x.map{ p =>
+              p match {
+                case JString(p) => Some(p)
+                case _ => None
```
sunahsuh: I suggest flattening out the Option[String] and calling toList here so that you end up with Option[List[String]] and can use the existing process list without doing the map later

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -0,0 +1,23 @@
...
+  val keyed: Boolean
+  val processes: List[String]
+}
+
+class Metric {
+  def getProcesses(definitionProcesses: List[String]): List[String] = {
+    definitionProcesses.map{
+      _ match {
+        case "main" => "parent"
+        case o => o
```
sunahsuh: This is a bit pedantic but technically according to the docs you can say `all` or `all_child` *along with* other values, which would break this code. More importantly, you can remove a level of matching if you flatmap with a guard on all/all_child -> List and then dedupe, which I think would be cleaner anyway.

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -0,0 +1,23 @@
+package com.mozilla.telemetry.metrics
+
+import com.mozilla.telemetry.utils.MainPing
+
+trait MetricDefinition {
```
sunahsuh: This trait isn't related to Metric -- why not keep it in the file it was in?

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -0,0 +1,23 @@
...
+  val keyed: Boolean
+  val processes: List[String]
+}
+
+class Metric {
```
sunahsuh: `Metric` feels like the wrong word semantically, if it's supposed to be a base for `HistogramsClass` and `ScalarsClass`. Also, getProcesses is basically a static method -- maybe it would be better suited for a package object.

src/main/scala/com/mozilla/telemetry/utils/MainPing.scala
sunahsuh: This is a larger "HistogramsClass needs a bit of love" thing but there *is* already a `processTypes` value in `HistogramsClass`, albeit without the parent -- I don't think this is within the scope of this PR but it seems like the right thing to do would be to refactor `HistogramsClass` to be able to use the full `processTypes` list and remove the one in `HistogramsClass`. Possibly worth submitting a bug for.

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -0,0 +1,23 @@
...
+  def getProcesses(definitionProcesses: List[String]): List[String] = {
+    definitionProcesses.map{
+      _ match {
+        case "main" => "parent"
+        case o => o
+      }
+    } match {
+      case List("all") => MainPing.ProcessTypes
+      case List("all_childs") => MainPing.ProcessTypes.filter(_ != "parent")
```
sunahsuh: The docs call this `all_child` instead of `all_childs`

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
fbertsch: My rationale was that we are defining `ScalarDefinition` and `ScalarClass` in the same file, as well as `HistogramDefinition` and `HistogramClass` in the same file, so `MetricDefinition` and `MetricClass` should be as well. In which case, the files names are: `Histogram`, `Scalar`, `Metric`

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
fbertsch: I think I would prefer it to stay in a superclass for two reasons:
1. There is a lot of duplicate code in `HistogramsClass` and `ScalarsClass`. We should start to integrate that into a superclass moving forward.
2. This function shouldn't be used by anyone *but* those - I don't really want any other code reading the `Histograms.json` or `Scalars.yaml` files, since they may not properly parse them.

I did change the name to `MetricsClass` to better match the subclasses.

fbertsch:
@sunahsuh RFAL

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -7,17 +7,17 @@ trait MetricDefinition {
...
-    definitionProcesses.map{
+    definitionProcesses.flatMap{
       _ match {
-        case "main" => "parent"
-        case o => o
+        case "main" => "parent" :: Nil
+        case "all" => MainPing.ProcessTypes
+        case "all_childs" => MainPing.ProcessTypes.filter(_ != "parent")
+        case "all_child" => MainPing.ProcessTypes.filter(_ != "parent")
+        case "all_children" => MainPing.ProcessTypes.filter(_ != "parent")
```
sunahsuh: You can turn these three lines into `case "all_childs" | "all_child" | "all_children" => MainPing.ProcessTypes.filter(_ != "parent")`

src/main/scala/com/mozilla/telemetry/metrics/Metric.scala
```diff
@@ -0,0 +1,23 @@
...
+    definitionProcesses.flatMap{
+      _ match {
+        case "main" => "parent" :: Nil
+        case "all" => MainPing.ProcessTypes
+        case "all_childs" => MainPing.ProcessTypes.filter(_ != "parent")
+        case "all_child" => MainPing.ProcessTypes.filter(_ != "parent")
+        case "all_children" => MainPing.ProcessTypes.filter(_ != "parent")
+        case o => o :: Nil
+      }
+    }.toSet.toList
```
sunahsuh: Why not use `.distinct`?

fbertsch:
@sunahsuh those were good suggestions :) RFAL!

sunahsuh:
:shipit:

PR #254, opened 2017-06-22
Bug 1375560 - Dedupe experiments view

fbertsch:
This also includes a generic `dedupe()` function available for
all DataFrames, so we can easily call: `df.dedupe()` to remove
any duplicates.


src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -258,4 +259,13 @@ object UDFs{
       spark.udf.register(HllCardinality, hllCardinality _)
     }
   }
+
+  implicit class Deduper(df: DataFrame) {
```
sunahsuh: Can we put this somewhere else? This isn't a UDF -- I think renaming this MozDataFrame or some other more general name and designating it as a place to put our own implicit DataFrame utilities would be rather useful

fbertsch:
@sunahsuh I missed the native implementation of dedupe :) Fixed!

PR #258, opened 2017-06-28
Bug 1309598 - Remove Loop metrics from main_summary & client_count datasets

SamPenrose:
There was a single test for NULL. No code specific to client_count.


mreid-moz:
@SamPenrose Can you take a look at the conflicts with master?

PR #262, opened 2017-07-07
Remove submission_date col from main_summary

fbertsch:
We should have added this to the bug squash :)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -484,10 +484,6 @@ object MainSummaryView {
           case JString(x) => x
           case _ => null
         },
-        fields.getOrElse("submissionDate", None) match {
-          case x: String => x
-          case _ => return None // required
-        },
         weaveConfigured.orNull,
         weaveDesktop.orNull,
         weaveMobile.orNull,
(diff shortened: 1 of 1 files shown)


SamPenrose:
Would it make sense to rename the existing column to "submission_date" to match analyst expectations?

fbertsch:
@SamPenrose probably not, we have `submission_date_s3` uniform across most datasets now. If we want to change this one we would need to change them all, which means actually renaming every file in every dataset to have that partitioned name.

mreid-moz:
I think this change requires some research and a communications plan - we want to minimize breakage and surprises. Research to figure out what (if anything) we know will be impacted, and comms to give folks a heads-up that this change is happening.

@fbertsch thoughts?

fbertsch:
@mreid-moz definitely a good idea. I was also starting to question the efficiency difference in `submission_date` and `submission_date_s3`; parquet stores some stats about columns, presumably some `min` and `max` would be part of that. Given those, after reading the metadata for each parquet file in other `submission_date_s3` partitions, it wouldn't read the data itself (i.e. predicate pushdown).

mreid-moz:
@fbertsch Let's test the efficiency! If there's minimal (or no) detectable difference, I don't think this is really worth the potential breakage.

fbertsch:
Create bug 1387617 to investigate.

PR #263, opened 2017-07-10
Bug 1379621 - Fix failing analysis job

fbertsch:
Also refactors a bit, moving to s3 buckets and taking locations from the views themselves.


fbertsch:
Note: A todo is to change all `experiment_aggregates` to `experiment_analysis`. I'll save this for a later date because it will require moving all of the files in s3, and I'm just not feeling that motivated on a Monday morning.

sunahsuh:
@fbertsch can you take a look at the failure?

PR #266, opened 2017-07-12
Bug 1378804 - Add Process Metrics to Experiment Analysis View

fbertsch:
I included a bit of the reverted code, since I needed it for testing. I stripped out everything to do with the derived scalars, which we can talk about re-adding down the line (with some changes).

Look at the commit for a summary of changes. Kept Histograms.definitions backwards compatible with LongitudinalView.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
+++ b/src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
@@ -9,14 +9,16 @@ import com.mozilla.telemetry.utils.MainPing
 case class RawHistogram(values: Map[String, Int], sum: Long)
 
 sealed abstract class HistogramDefinition extends MetricDefinition
-case class FlagHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
-case class BooleanHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
-case class CountHistogram(keyed: Boolean, processes: List[String]) extends HistogramDefinition
-case class EnumeratedHistogram(keyed: Boolean, nValues: Int, processes: List[String]) extends HistogramDefinition
-case class LinearHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int, processes: List[String]) extends HistogramDefinition
-case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int, processes: List[String]) extends HistogramDefinition
+case class FlagHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class BooleanHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class CountHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class EnumeratedHistogram(keyed: Boolean, originalName: String, nValues: Int, process: Option[String] = None) extends HistogramDefinition
+case class LinearHistogram(keyed: Boolean, originalName: String, low: Int, high: Int, nBuckets: Int, process: Option[String] = None) extends HistogramDefinition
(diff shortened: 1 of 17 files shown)


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -205,7 +205,7 @@ object LongitudinalView {
        8 workers and 20 executors, 320 partitions provide a good compromise. */
 
     val histogramDefinitions = Histograms.definitions(includeOptin)
-    val scalarDefinitions = Scalars.definitions(includeOptin)
+    val scalarDefinitions = Scalars.definitions(includeOptin).filter(_._2.process == Some("parent"))
```
mreid-moz: Please add a comment here describing the set of scalar definitions we're filtering/keeping

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -708,7 +707,10 @@ object MainSummaryView {
 
       Some(Row.merge(row, scalarRow, histogramRow))
     } catch {
-      case _: Exception =>
+      case e: Exception =>
```
mreid-moz: What are the possible error cases here? Can we be more selective about what we're catching?

fbertsch: This was there previously. I'm not sure what the class of Exceptions that can result are, but I'm a bit hesitant ti change it because doing so could cause the job fail (i.e. it raises an uncaught exception, we add that one, another is raised by other improper data, this continues ad infintum). Let's fix this in a follow-up.

src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
```diff
@@ -9,14 +9,16 @@ import com.mozilla.telemetry.utils.MainPing
...
-case class ExponentialHistogram(keyed: Boolean, low: Int, high: Int, nBuckets: Int, processes: List[String]) extends HistogramDefinition
+case class FlagHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class BooleanHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class CountHistogram(keyed: Boolean, originalName: String, process: Option[String] = None) extends HistogramDefinition
+case class EnumeratedHistogram(keyed: Boolean, originalName: String, nValues: Int, process: Option[String] = None) extends HistogramDefinition
+case class LinearHistogram(keyed: Boolean, originalName: String, low: Int, high: Int, nBuckets: Int, process: Option[String] = None) extends HistogramDefinition
+case class ExponentialHistogram(keyed: Boolean, originalName: String, low: Int, high: Int, nBuckets: Int, process: Option[String] = None) extends HistogramDefinition
 
 class HistogramsClass extends MetricsClass {
+  def HistogramPrefix = "histogram"
```
fbertsch: In `ScalarsClass`, it is `ScalarPrefix`. I prefer having capital constants.

src/main/scala/com/mozilla/telemetry/metrics/Histogram.scala
```diff
@@ -151,23 +157,30 @@ class HistogramsClass extends MetricsClass {
           .asInstanceOf[List[String]]
         )
 
-        def addSuffixes(key: String, definition: HistogramDefinition): List[(String, HistogramDefinition)] = {
-            suffixes.map(suffixType => (key + suffixType, definition))
```
fbertsch: Done and done.

fbertsch:
@sunahsuh My hope is that once we get the wide-dataframes issue fixed, we can make longitudinal directory from main_summary using GenericLongitudinal.. Then we can not only wipe out all that code, but also have the schemas match exactly (except, ya know, as arrays).

PR #267, opened 2017-07-13
Add a .dockerignore file

mreid-moz:
If there's a bunch of stuff in your working directory, it all gets
sent to Docker when building the image, which causes it to take a
long time (particularly if you've built the assembly jar).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git
+.idea
+docs
+lib_managed
+target


wlach:
Are you sure about this? I don't see any directive to copy files from the checkout here:

https://github.com/mozilla/telemetry-batch-view/blob/d541518f6b06c9e39dba4333571fc602da1bb796/Dockerfile

We do mount the project directory as a volume inside the docker container when running the tests, but AFAIK that shouldn't be terribly slow? And does that even use .dockerignore?

mreid-moz:
it doesn't copy the files, it's part of the "Sending build context to Docker daemon <some size>" step when building the image.

Before this change, I saw something like `Sending build context to Docker daemon 2.816 GB`, and after, it was down to `Sending build context to Docker daemon 8.767 MB`

wlach:
Crazy! You learn something new every day. In that case, LGTM :)

PR #269, opened 2017-07-19
Bug 1307187 - Add process scalars to longitudinal

fbertsch:
Was in the neighborhood anyways.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -821,25 +821,28 @@ object LongitudinalView {
     implicit val formats = DefaultFormats
 
     // Get a list of the scalars in the ping.
-    val scalarList = payloads.map{ case (x) =>
-      // Get the payload of the Heka frame, which contains the payload for the
-      // main ping.
+    val scalarsList = payloads.map{ case (x) =>
       val payload = parse(x.getOrElse("payload", return).asInstanceOf[String])
-      // If we can't find the "scalars" section for a ping, don't throw away all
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -821,25 +821,26 @@ object LongitudinalView {
...
+      } toMap
     }
 
+    val byProcessScalarList = MainPing.ProcessTypes.map{ process =>
+      process -> scalarsList.map(_(process))
+    } toMap
+
     for {
       (key, definition) <- scalarDefinitions.toList
+      payloads = byProcessScalarList(definition.process.get)
```
harterrt: Using `payloads` here confused me. If I understand correctly, this variable stores a list of lists of scalar json, right? I conflated this variable with the 'payloads' from above. Maybe `scalar_payloads` would be a better name?

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -848,17 +849,28 @@ object LongitudinalView {
...
-      }
+      MainPing.ProcessTypes.map{ process =>
+        (payload \ "processes" \ process \ "keyedScalars").toOption match {
+          case Some(scalars) => process -> scalars.extract[Map[String, Map[String, AnyVal]]]
+          case _ => process -> Map[String, Map[String, AnyVal]]()
+        }
+      } toMap
     }
 
+    // this allows us to avoid iterating
```
harterrt: This comment was helpful. Maybe include it above as well.

fbertsch:
@harterrt RFAL

PR #271, opened 2017-07-20
Add metadata (client and ping count) row to experiment aggregates

sunahsuh:
This adds a row to the experiment aggregates output which includes ping and client count as "Statistics" objects for each branch/subgroup

Required for mozilla/experiments-viewer#155

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentAnalyzer.scala
@@ -1,0 +1,37 @@
+package com.mozilla.telemetry.experiments.analyzers
+
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.{col, lit, count, sum, countDistinct}
+
+
+case class ExperimentMetadata(experiment_id: String,
+                              branch: String,
+                              subgroup: String,
+                              client_count: Long,
+                              ping_count: Long)
+
+object ExperimentAnalyzer {
+  def getExperimentMetadata(df: DataFrame): Dataset[MetricAnalysis] = {
+    import df.sparkSession.implicits._
+
+    val counts = df
+      .select(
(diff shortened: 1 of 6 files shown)


src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentAnalyzer.scala
```diff
@@ -0,0 +1,38 @@
...
+  def getExperimentMetadata(df: DataFrame): Dataset[MetricAnalysis] = {
+    import df.sparkSession.implicits._
+
+    val counts = df
+      .select(
+        col("experiment_id"),
+        col("experiment_branch").as("branch"),
+        lit("All").as("subgroup"),
+        col("client_id"))
+      .groupBy(col("experiment_id"), col("branch"), col("subgroup"), col("client_id"))
```
fbertsch: I'm not clear on why there's two `groupBy`s. Couldn't this be removed and line 26 just be `.agg(countDistinct(col("client_id")).alias("client_count"), count("*").alias("ping_count")`*. Are we gaining something here that I'm missing?

*note could also use approx_distinct, but probably unnecessary.

sunahsuh: Heh, basically, I was looking at the wrong doc page to see which aggregations were available (https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.sql.RelationalGroupedDataset instead of sql.functions) 😛

But yeah, if we are running this job with enough clients that we have to use approx_distinct I think we'll run into other issues first

sunahsuh:
RFAL -- note I made the output of `getExperimentMetrics` a Dataset since keeping the output typed makes the tests a bit nicer and keeps up from having to cast at runtime.

Also, please keep this PR open so @robhudson can make any changes he needs to on his side first before we merge.

sunahsuh:
Okay, I'm just going to merge now since robhudson would prefer to have output to work against anyway

PR #272, opened 2017-07-21
Add logging to ExperimentAnalysisView

sunahsuh:
Per our discussion -- adds logging so we have a slightly better idea of where things are failing in the experiment aggregate job

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
@@ -10,6 +10,8 @@ object ExperimentAnalysisView {
   def schemaVersion: String = "v1"
   def jobName: String = "experiment_analysis"
 
+  private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
+
   // Configuration for command line arguments
   class Conf(args: Array[String]) extends ScallopConf(args) {
     // TODO: change to s3 bucket/keys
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -89,11 +99,15 @@ object ExperimentAnalysisView {
     val experimentData = data.where(col("experiment_id") === experiment)
 
     val metrics = metricList.map {
-      case (name: String, hd: HistogramDefinition) =>
-        new HistogramAnalyzer(name, hd, experimentData).analyze()
-      case (name: String, sd: ScalarDefinition) =>
-        ScalarAnalyzer.getAnalyzer(name, sd, experimentData).analyze()
-      case _ => throw new UnsupportedOperationException("Unsupported metric definition type")
+      case (name: String, md: MetricDefinition) =>
+        logger.info(s"Aggregating metric $name")
```
fbertsch: Unfortunately I don't think this will do what we want :( these values will all be logged, and it's not until line 51 that spark will actually do the work. We need the logging within Spark itself to let us know which values are actually being aggregated.

sunahsuh: Doh, right -- so, it looks like ParquetReadSupport logs the column names and schema each time it reads from disk anyway, so changing the log level seems like it'll suffice for our purposes (also, given the DAG didn't fail over the weekend we unfortunately seem to have an intermittent failure :()

Alternately, given the actual size of the output I don't think we'd have an issue if we wanted to collect the results sequentially as we go along, but that's probably unnecessary given we can get what we need from the parquet INFO-level logs.

sunahsuh:
Okay, I removed the useless per-metric log but I think the rest of it is still useful at least while we're in dev/debugging mode

PR #274, opened 2017-08-02
Fix spark mode on experiment jobs

sunahsuh:
Welp. I'm gonna guess at least a few of the problems with the experiment jobs are due to hard-coding local spark mode.


acmiyaguchi:
I'm going to rerun the experiments job for 20170724 on airflow.

PR #275, opened 2017-08-02
Add the "isWebextension" info to the AMODatabase

Dexterp37:
This also updates the AddonRecommender to make sure that
information bit is also reported into the addon mapping file
used to produce the results for the collaborative filter.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/ml/AMODatabase.scala
+++ b/src/main/scala/com/mozilla/telemetry/ml/AMODatabase.scala
@@ -1,7 +1,7 @@
 package com.mozilla.telemetry.ml
 
 import java.nio.charset.StandardCharsets
-import java.nio.file.{Files, Paths}
+import java.nio.file.{Files, Path, Paths}
 
 import org.json4s._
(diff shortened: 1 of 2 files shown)


Dexterp37:
I've just finished testing this in a cluster, the addons file is correctly generated and the recommender seems to pick it up well.

maurodoglio:
@Dexterp37 do you mind to split this commit according to the set of changes that you mentioned in the commit message? As a new reviewer on this codebase I find hard to understand what each part is doing.

src/main/scala/com/mozilla/telemetry/ml/AMODatabase.scala
```diff
@@ -109,11 +120,39 @@ final object AMODatabase {
...
     */
   def contains(addonId: String): Boolean = {
     addonDB.contains(addonId)
   }
+
+  /**
+    * Get the path to the local cache file.
+    * @return A Path object representing the location of the addons cache file.
+    */
+  def getLocalCachePath() : Path = {
```
maurodoglio: Please use this function everywhere in this file vs Paths.get

PR #278, opened 2017-08-03
Bug 1387107 - Stop and start spark session between experiment runs

fbertsch:
The job was failing intermittently. We've seen this before on
consecutive spark jobs, for example, main_summary.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
@@ -24,33 +24,41 @@ object ExperimentAnalysisView {
     verify()
   }
 
+  def getSpark: SparkSession = {
+    val spark = getOrCreateSparkSession(jobName)
+    spark.sparkContext.setLogLevel("INFO")
+    spark
+  }
+
   def main(args: Array[String]) {
     val conf = new Conf(args)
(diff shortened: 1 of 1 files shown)


fbertsch:
@sunahsuh currently running on a cluster, I'm hopeful it will finish about the same time the build does.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -24,33 +24,41 @@ object ExperimentAnalysisView {
...
     spark.sparkContext.setLogLevel("INFO")
+    spark
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args)
+    var spark = getSpark
 
-    val data = spark.read.parquet(conf.inputLocation())
+    var data = spark.read.parquet(conf.inputLocation())
```
sunahsuh: Looks like if we rename just this initial instance of `spark` and `data` (and the call to `getExperiments()` then we can avoid mutable vars

fbertsch:
@sunahsuh RFAL

sunahsuh:
Fingers crossed!

fbertsch:
@sunahsuh ran on a cluster, took <1 hour, which seems to be what we'd expect. Let's kick this off!

PR #281, opened 2017-08-08
Use mozilla/spark-hyperloglog from Sonatype

fbertsch:


fbertsch:
@mreid-moz good point, I think so!

PR #283, opened 2017-08-16
Add HBase view for the addon recommender

Dexterp37:
This PR adds a scala job that scans the main summary and feeds addon plus some other data into an HBase instance.

**Do not review yet**.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
@@ -1,0 +1,144 @@
+package com.mozilla.telemetry.views
+
+import com.mozilla.telemetry.utils._
+import org.apache.hadoop.hbase.client.HBaseAdmin
+import org.apache.hadoop.hbase.io.compress.Compression
+import org.apache.hadoop.hbase.util.RegionSplitter
+import org.apache.hadoop.hbase.{HColumnDescriptor, HTableDescriptor}
+import org.apache.spark.sql.expressions.Window
+import org.apache.spark.sql.functions.{desc, row_number}
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.joda.time.{DateTime, Days, format}
+import org.rogach.scallop._
+import unicredit.spark.hbase._
+
+object HBaseAddonRecommenderView {
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val from = opt[String]("from", descr = "From submission date", required = false)
+    val to = opt[String]("to", descr = "To submission date", required = false)
+    val hbaseMaster = opt[String]("hbase-master", descr = "IP address of the HBase master", required = true)
+    verify()
+  }
+
+  private[views] val tableName = "addon_recommender_view"
+  private[views] val columnFamily = "cf"
+  private[views] val column = "payload"
+
+  private[views] def createHBaseTable(useCompression: Boolean = true)(implicit config: HBaseConfig): Unit = {
+    val admin = new HBaseAdmin(config.get)
+    if (!admin.tableExists(tableName)) {
+      val splits = new RegionSplitter.HexStringSplit().split(1000)
+      val td = new HTableDescriptor(tableName)
+      val cd = new HColumnDescriptor(columnFamily)
+      cd.setTimeToLive(60*60*24*90)  // 90 days retention
+      cd.setMaxVersions(1) // Keep a single version of each document
+      if (useCompression) {
+        cd.setCompressionType(Compression.Algorithm.SNAPPY)
+      }
+      td.addFamily(cd)
+      admin.createTable(td, splits)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  private[views] val tableName = "addon_recommender_view"
+  private[views] val columnFamily = "cf"
+  private[views] val column = "payload"
+
+  private[views] def createHBaseTable(useCompression: Boolean = true)(implicit config: HBaseConfig): Unit = {
+    val admin = new HBaseAdmin(config.get)
+    if (!admin.tableExists(tableName)) {
+      /*
+         About 200 GB of compressed data is generated per day. As we want to store data for 90 days we expect to
```
maurodoglio: Please delete this comment as it is obsolete. We should probably tune the region split value to reflect the kind of small dataset that we are storing.

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -0,0 +1,154 @@
...
+
+        // Get the data for the desired date.
+        val data = datasetForDate(currentDateString)
+
+        // Only get the most recent entry (by subsession start date rather than submission
+        // date) for each client.
+        val clientsData: DataFrame = data
+            .select(
+              $"client_id",
+              $"document_id",
```
maurodoglio: We don't need document_id anymore, do we?

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -0,0 +1,154 @@
...
+              $"scalar_parent_browser_engagement_unique_domains_count",
+              $"active_addons",
+              row_number()
+                .over(Window.partitionBy("client_id")
+                  .orderBy(desc("subsession_start_date")))
+                .alias("clientid_rank"))
+          .where($"clientid_rank" === 1)
+
+        // Convert the DataFrame to JSON and get an RDD out of it.
+        val subset = data.select("client_id", "document_id", "subsession_start_date")
```
maurodoglio: s/data/clientsData on this line and the one below.

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -0,0 +1,154 @@
...
+                                   "locale",
+                                   "os",
+                                   "places_bookmarks_count",
+                                   "scalar_parent_browser_engagement_tab_open_event_count",
+                                   "scalar_parent_browser_engagement_total_uri_count",
+                                   "scalar_parent_browser_engagement_unique_domains_count",
+                                   "active_addons").toJSON.rdd
+
+        // Build an RDD containing (key, (document, timestamp)) tuples: one per client.
+        val rdd = subset.rdd.zip(jsonData).flatMap{ case (row, json) =>
```
maurodoglio: I'm wondering if there's a better way to obtain key/value tuples out of the rdd now that we only have client_id as the key

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -0,0 +1,154 @@
...
+          val clientId = row.getString(0)
+          val startDate = row.getString(2)
```
maurodoglio: you can probably get rid of startDate & co now that we only keep one value per client. timestamp will be added automatically by hbase, so we don't need to handle that.

src/test/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderViewTest.scala
```diff
@@ -0,0 +1,103 @@
...
+
+import org.apache.hadoop.hbase.client.HBaseAdmin
+import org.apache.spark.sql.SparkSession
+import org.joda.time.DateTime
+import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
+import org.scalatest.tagobjects.Slow
+import unicredit.spark.hbase._
+
+case class TestMainSummaryPing(client_id: Option[String],
+                               document_id: Option[String],
```
maurodoglio: you can get rid of document_id

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
Dexterp37: I removed it and it turns out that the code was implicitly performing data validation and discarding entries with invalid dates in the `subsession_start_date` field. I changed the code to keep the validation part and dropped adding it to the HBase.

PR #284, opened 2017-08-17
Bug 1388867 - Add heavy_users dataset

fbertsch:
Did this a bit backwards, I'll be creating the job that calculates and uploads the cutoffs after this one.

Note: I originally wrote this using DataFrames, then decided to make it as typed as possible with Datasets. Feel free to comment on code quality of that sort.


fbertsch:
Refactoring this.

PR #285, opened 2017-08-18
Bug 1388867 - Add heavy_users dataset

fbertsch:
Took a stab at using only Typed Datsets, feel free to comment on readability/code quality.


src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  /*
+   * For the first 27 days, we can't actually calculate
+   * whether they are a heavy user, we can only sum the
+   * active_ticks. For those, we will just boostrap the
+   * table, and store NULL for heavy_user and
+   * prev_year_heavy_user.
+   */
+  def getBootstrapStatus(ds: Dataset[HeavyUsersRow], date: DateTime): Boolean = {
```
sunahsuh: It's not entirely clear from this name whether True means "yes we should bootstrap" or "no we do not need to bootstrap"

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+      .map(addCutoffs(date, bootstrap, cutoff, prevYearCutoff))
+      .as[HeavyUsersRow]
+
+    (withCutoffs, cutoff)
+  }
+
+  private def getCutoffs(spark: SparkSession, bucket: String): Map[String, Double] = {
+    val cutoffsPath = s"s3://$bucket/$CutoffsDirectory"
+    spark.read.csv(cutoffsPath)
+      .collect()
```
sunahsuh: If we're opening this to collect it immediately, perhaps we should read it directly?

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+      case false => Some(
+        ds.toDF().stat.approxQuantile("active_ticks_period", Array(0.9), 0.0001).head)
+    }
+  }
+
+  /*
+   * Given bootstrap status and cutoffs, get a finalized HeavyUsersRow for the client's data
+   */
+  private def addCutoffs(date: String, bootstrap: Boolean, cutoff: Option[Double], prevYearCutoff: Option[Double])(r: UserActiveTicks): HeavyUsersRow = {
+    val hus = (c: Option[Double], v: Long) => for{ e <- c } yield v >= e
```
sunahsuh: Okay, so I see where the "compare cutoffs" function in the tests comes from, and why you wanted to use a one-liner, but since this is 1. repeated and 2. counter to the pattern elsewhere, maybe we can just stick it in a utility function as a match/case

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+    val heavyUser = if(bootstrap) None else hus(cutoff, r.active_ticks_period)
```
sunahsuh: isn't the if statement redundant here and in the line below?

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+   * of the closest date.
+   */
+  def getCutoff(date: DateTime, cutoffs: Map[String, Double]): Option[Double] = {
+    cutoffs.size match {
+      case 0 => None
+      case _ =>
+        val closest = cutoffs.keys.map(k => (k, fmt.parseDateTime(k)))
+          .map{ case (k, k_date) => (k, if(k_date < date) (k_date to date).millis else if (k_date > date) (date to k_date).millis else 0)}
+          .minBy(_._2)._1
+        Some(cutoffs(closest))
```
sunahsuh: On day 28 (first day of non-bootstrap status) we'll start using day 1 as the previous year's cutoff -- seems like that'd be a bit counter to expectations?

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+      .map(e => e.getAs[String](0) -> e.getAs[String](1).toDouble)
+      .toMap
+  }
+
+  private def storeCutoffs(spark: SparkSession, cutoffs: Map[String, Double], date: String, bucket: String): Unit = {
+    import spark.implicits._
+    val cutoffsPath = s"s3://$bucket/$CutoffsDirectory"
+    val current = cutoffs.toList.toDF()
+    current.write.mode("overwrite").csv(cutoffsPath)
+    current.write.csv(s"$cutoffsPath-$date")
```
sunahsuh: any reason not to make this also an overwrite (in case we re-run a date?)

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+ * a total of active_ticks for that day.
+ *
+ * Info on this is in bug 1388867.
+ * The metabug for the data engineering effort for heavy users
+ * is bug 1388732.
+ **/
+object HeavyUsersView {
+
+  val CutoffsDirectory = "heavy_users_cutoffs"
+  val CutoffsFile = "cutoffs.csv"
```
sunahsuh: This is unused, as far as I can tell?

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
sunahsuh: Also, as it stands this would write a new directory into the root of the bucket (e.g. telemetry-parquet). Did you mean to use filenames for this?

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
fbertsch: Good catch! Actually Saptarshi has already calculated the cutoffs through early 07/2016, but only for MWF. As such we decided to use the closest day as the cutoff. TLDR: the list is seeded with old dates.

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+    val subtractDay = fmt.print(datetime - TimeWindow)
+
+    val prevData: Dataset[HeavyUsersRow] = existingData
+      .filter(_.submission_date_s3 == prevDay)
+    val subtractData: Dataset[HeavyUsersRow] = existingData
+      .filter(_.submission_date_s3 == subtractDay)
+
+    // Aggregate today's main_summary data
+    val aggregated: Dataset[UserActiveTicks] = ds
+      .filter(r => r.submission_date_s3 == date && Option(r.active_ticks).getOrElse(0: Long) > 0)
```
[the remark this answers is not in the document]
fbertsch: Yes Saptarshi ran comparisons and we decided this is "good enough" - especially when considering the pingsender moving forward.

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
fbertsch: Yeah, I considered that, but honestly it seems easier to read/write using just Spark. Consider what I have versus e.g. https://stackoverflow.com/questions/40132148/read-properties-file-located-in-s3-in-scala :(

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -0,0 +1,298 @@
...
+      .groupByKey(r => (r.client_id, r.sample_id, r.profile_creation_date))
+      .agg(sumLong[MainSummaryRow](_.active_ticks))
+      .map(r => UserActiveTicks(r._1._1, r._1._2, r._1._3, r._2, r._2))
```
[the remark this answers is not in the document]
fbertsch: re: `3. ...you could get rid of the UserActiveTicks case class`, I was initially going to do this, but the downside is then Spark is carrying around these extra values (`date`, `heavy_user`, `prev_year_heavy_user`) for no reason. It's a performance-reading tradeoff, though unfortunately I don't know how much of a performance gain it gives.

fbertsch:
@sunahsuh RFAL, sorry I did a force commit rather than adding an extra commit with my changes :(

fbertsch:
@sunahsuh also, I'm not really happy with my solution to the duplicate `submission_date_s3` columns (which is to just drop the one in the output). I would prefer to do all of it using Datasets, but I think it would require yet _another_ case class :/

PR #286, opened 2017-08-21
Added selection of scalar firstPaint value over SimpleMeasurement.

vgutierrez9:
Bug # TBA - Added code to choose the scalar version of firstPaint, when present in the payload, over the SimpleMeasurements version.

Includes tests. but we still need to investigate whether SimpleMeasurements firstPaint shows in every ping or just the first subsession ping, and if the scalar is behaving the same way.

Attn: @sunahsuh


src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1554,4 +1554,89 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
 
     selectedData should be (111)
   }
+
+  "Scalar implementations of Simple Measurements activeTicks and firstPaint" can "be properly selected" in {
```
sunahsuh: You can probably remove the `activeTicks and firstPaint` now :)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1554,4 +1554,89 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+      None);
+
+    val summary = MainSummaryView.messageToRow(messageBothPresent, scalarsDef, histogramDefs)
+
+    val selectedActiveTicks =
+      applySchema(summary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs)).getAs[Int]("active_ticks")
+    selectedActiveTicks should be (888)
+
+    val selectedFirstPaint =
+      applySchema(summary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs)).getAs[Int]("first_paint")
```
sunahsuh: I'd create a new val that holds `applySchema(summary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs))` and use that in your test assertions instead of doing this twice

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1554,4 +1554,89 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+    selectedFirstPaint should be (999)
+  }
+
+
+  "Simple Measurements activeTicks and firstPaint" can "be properly selected in the absence of a scalar value" in {
```
sunahsuh: Same comments as above, although I think something like `"Migrated scalar values" can "fall back to simple measurements values"` might be a better description

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1554,4 +1554,89 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+              }
+            }
+          }
+        }"""),
+      None);
+
+    val summary = MainSummaryView.messageToRow(messageSMPresent, scalarsDef, histogramDefs)
+
+    val selectedActiveTicks =
+      applySchema(summary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs)).getAs[Int]("active_ticks")
```
sunahsuh: Same comment as above

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -656,7 +656,14 @@ object MainSummaryView {
 
         // bug 1353114 - payload.simpleMeasurements.*
         asInt(simpleMeasures \ "main"),
-        asInt(simpleMeasures \ "firstPaint"),
+
+        // bug TBA - use scalar version when available.
```
sunahsuh: I would just remove the `bug TBA` here

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
vgutierrez9: Just to confirm, I can delete the active-ticks specific test and keep the test that covers both?

sunahsuh: Ah, no, I meant from the test name :)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1513,33 +1516,39 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
 
     val messageSMPresent = RichMessage(
       "1234",
       Map(
         "documentId" -> "foo",
         "submissionDate" -> "1234",
-        "payload.simpleMeasurements" -> """{"activeTicks": 111}""",
+        "payload.simpleMeasurements" ->
+          """{"activeTicks": 111,
+              "firstPaint": 222}""".stripMargin,
```
sunahsuh: stripMargin doesn't do anything here (https://www.scala-lang.org/api/current/scala/collection/immutable/StringLike.html#stripMargin:String)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1548,10 +1557,13 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
-    val selectedData =
-      applySchema(summary.get, MainSummaryView.buildSchema(activeTicksDef, histogramDefs)).getAs[Int]("active_ticks")
+    val messageSummary = MainSummaryView.messageToRow(messageSMPresent, scalarsDef, histogramDefs)
+    val appliedSummary = applySchema(messageSummary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs))
+
+    val selectedActiveTicks = appliedSummary.getAs[Int]("active_ticks")
+    selectedActiveTicks should be (111)
 
-    selectedData should be (111)
+    val selectedData = appliedSummary.getAs[Int]("first_paint")
```
sunahsuh: please use a more descriptive variable name (or you can choose not to use a variable at all and do `appliedSummary.getAs[Int]("first_paint") should be (222)` directly)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1513,33 +1516,39 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
-    selectedData should be (999)
+    val selectedActiveTicks = appliedSummary.getAs[Int]("active_ticks")
+    selectedActiveTicks should be (888)
+
+    val selectedFirstPaint = appliedSummary.getAs[Int]("first_paint")
+    selectedFirstPaint should be (999)
   }
 
-  "Simple Measurements activeTicks" can "be properly selected in the absence of a scalar value" in {
-    val scalarUrlMockActiveTicks = (a: String, b: String) => Source.fromFile("src/test/resources/ScalarsActiveTicks.yaml")
```
sunahsuh: You should also delete this old file (`git rm <filename>`)

vgutierrez9:
RFAL :)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1513,33 +1516,39 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+          """{"activeTicks": 111,
+              "firstPaint": 222}""",
         "submission" ->
           """{
           "payload": {
             "processes": {
               "parent": {
                 "scalars": {
-                  "browser.engagement.active_ticks": null
+                  "browser.engagement.active_ticks": null,
```
sunahsuh: More realistic here would be not having these scalar keys at all, right?

vgutierrez9:
RFAL (:

vgutierrez9:
Added the test case for when the scalar definition is not present.
RFAL 😸

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1548,10 +1555,23 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+    val messageSummary = MainSummaryView.messageToRow(messageSMPresent, scalarsDef, histogramDefs)
+    val appliedSummary = applySchema(messageSummary.get, MainSummaryView.buildSchema(scalarsDef, histogramDefs))
+
+    val selectedActiveTicks = appliedSummary.getAs[Int]("active_ticks")
+    selectedActiveTicks should be(111)
+
+    val selectedFirstPaint = appliedSummary.getAs[Int]("first_paint")
+    selectedFirstPaint should be(222)
+
+    // If the scalar's definition does not exist, the simple measurement value should be selected.
```
sunahsuh: This should be a new test, i think

vgutierrez9: I agree, will do!

vgutierrez9:
RFAL 💃

sunahsuh:
You'll need to rebase this to master now to update it before merging

PR #287, opened 2017-08-22
Bug 1391307 - Add flag to main_summary to include all histograms

fbertsch:
Hopefully we can use this soon :)


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -58,6 +58,7 @@ object MainSummaryView {
     val limit = opt[Int]("limit", descr = "Maximum number of files to read from S3", required = false)
     val channel = opt[String]("channel", descr = "Only process data from the given channel", required = false)
     val appVersion = opt[String]("version", descr = "Only process data from the given app version", required = false)
+    val allHistograms = opt[Boolean](descr = "Flag to use all histograms", required = false)
```
mreid-moz: One nitpicky thing - all the other scallop-arg constructors I looked at in this repo have a string name as their first argument, and this one doesn't. I checked and it works as expected (`--all-histograms` populates `conf.allHistograms`), but it seems like a good idea to keep 'em all consistent.

PR #289, opened 2017-08-23
Bug 1393002: Collect each metric result out of spark to reduce number of simultane…

sunahsuh:
…ous task results


sunahsuh:
This fixed OOM errors I was able to reproduce on a test cluster

sunahsuh:
Hold on the review, I have to fix a test

sunahsuh:
Yeah, I agree -- feels like it shouldn't be this hard for not that much data. I suspect the permutation tests will be a point that might trigger a few more issues, so that might be a good time to reconsider the whole structure.

PR #290, opened 2017-08-23
Bug 1381641 - Implement a sync view that flattens sync engine data

thomcc:
Ideally I'd like for my patch for [Bug 1390320](https://bugzilla.mozilla.org/show_bug.cgi?id=1390320) to land first, but I'm getting it up without waiting, since it is not a small change (although I think the github diff viewer is making it look bigger than it is) and it wouldn't really surprise me that much if it were a little controversial. Once that's available, I'll probably change SyncFlatView's loop body around a bit to take advantage of it.

Since the github diff doesn't seem to capture this well, most of the code in SyncPingConversion.scala previously lived in SyncView.scala.

This also addresses a very small issue ([Bug 1389233](https://bugzilla.mozilla.org/show_bug.cgi?id=1389233)), where android and iOS will write outgoing arrays as an object. This is a bug in them that is being addressed, but we'd like the (historical) data. I did it here since this code changes the outgoing handling anyway, so it is easy to do as part of this patch.

Somewhat rambling issue summary and background:

`sync_summary` has a good amount of data and has been useful for ad-hoc analysis, but has essentially been unqueryable in redash. Every query times out or exausts its available resources in one way or another. This is, essentially, because of a `cross join unnest` required since we store the engine sync data as an `engines` array in the sync row. That's more or less what this view addresses, by flattening that array out.

Additionally, frequently the parquet files written out by SyncView, are unreadable, due to, AFAICT, [this bug in parquet](https://issues.apache.org/jira/browse/PARQUET-511) (although I'm not 100% confident that's the bug, the error log is almost identical, but it seems... surprising, that we'd hit the limit). Regardless of the cause, reading a parquet file written by the SyncView seems to have a decent chance of failing in practice.

Anyway, I initially wanted to write this as a python rollup view. Due to performance issues around UDFs implemented in python (Even a trivial one caused a single day of data to take several hours), we wouldn't have as rich of data as this code provides. But even ignoring that, the headaches associated with the busted parquet files (as well as just the extremely large size of the data) made this really painful. After struggling with this for a while, I ended up asking @acmiyaguchi if he thought I could just land it as another sync view, and he gave me some indication along the lines of "well, if that's the best place for it..." (although I'm aware in an ideal world it probably is not).

So I prototyped this in a zeppelin notebook based on a modified version of the SyncView code, ran it across a little over three months, and ended up with something in redash that we can actually query. Even better, most of the queries complete very quickly.

Regarding replacing the current SyncView code vs adding a new sync view: We do have code that depends on the current view (not much), and the data the code I'm landing generates is a bit more annoying to work with. It also doesn't include everything we want in every situation, just the stuff we want to be able to directly query in redash. That said, if having two views into the sync ping data in this repo is completely abhorrent, we could probably include more data here, and change some other stuff on our end to make this doable.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
@@ -1,0 +1,593 @@
+package com.mozilla.telemetry.utils
+
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types._
+import org.joda.time.DateTime
+import org.json4s.{DefaultFormats, JValue}
+import org.json4s.JsonAST.{JArray, JInt, JObject, JString}
+import java.util.UUID
+
+// Common conversion code for SyncView and SyncFlatView. Schema for sync pings described here:
+// https://dxr.mozilla.org/mozilla-central/source/services/sync/tests/unit/sync_ping_schema.json
+object SyncPingConversion {
+  /*
+   * The type definitions for the rows we create.
+   */
+  private val failureType = StructType(List(
+    // failures are probably *too* flexible in the schema, but all current errors have a "name" and a second field
+    // that is a string or an int. To keep things simple and small here, we just define a string "value" field and
+    // convert ints to the string.
+    StructField("name", StringType, nullable = false),
+    StructField("value", StringType, nullable = true)
+  ))
+
+  // The record of incoming sync-records.
+  private val incomingType = StructType(List(
+    StructField("applied", LongType, nullable = false),
+    StructField("failed", LongType, nullable = false),
+    StructField("new_failed", LongType, nullable = false),
+    StructField("reconciled", LongType, nullable = false)
+  ))
+
+  // Outgoing records.
+  private val outgoingType = StructType(List(
+    StructField("sent", LongType, nullable = false),
+    StructField("failed", LongType, nullable = false)
+  ))
+
+  // Entries in devices array
+  private val deviceType = StructType(List(
+    StructField("id", StringType, nullable = false),
+    StructField("version", StringType, nullable = false),
+    StructField("os", StringType, nullable = false)
+  ))
+
+  // Data about a single validation problem found
+  private val validationProblemType = StructType(List(
+    StructField("name", StringType, nullable = false),
+    StructField("count", LongType, nullable = false)
+  ))
+
+  // Data about a validation run on an engine
+  private val validationType = StructType(List(
+    // Validator version, optional per spec, but we fill in 0 where it was missing.
+    StructField("version", LongType, nullable = false),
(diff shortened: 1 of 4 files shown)


src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
```diff
@@ -0,0 +1,593 @@
...
+  private def multiSyncPayloadToNestedRow(ping: JValue, syncs: List[JValue]): List[Row] = {
+    syncs.flatMap(sync => singleSyncPayloadToNestedRow(ping, sync))
+  }
+
+  // Convert an "old style" ping that records a single Sync to a (nested) row
+  private def singleSyncPayloadToNestedRow(ping: JValue, sync: JValue): Option[Row] = {
+    val application = ping \ "application"
+    val payload = ping \ "payload"
+
+    def stringFromSyncOrPayload(s: String): String = {
```
acmiyaguchi: Do you happen to know what part of the sync schema triggers this behavior between `uid` and `deviceID`?

src/main/scala/com/mozilla/telemetry/views/SyncFlatView.scala
```diff
@@ -0,0 +1,140 @@
...
+      case _ => DateTime.now.minusDays(1)
+    }
+    val from = conf.from.get match {
+      case Some(f) => fmt.parseDateTime(f)
+      case _ => DateTime.now.minusDays(1)
+    }
+
+    // Set up Spark
+    val sparkConf = new SparkConf().setAppName(jobName)
+    sparkConf.setMaster(sparkConf.get("spark.master", "local[*]"))
```
acmiyaguchi: I don't know if we set `spark.master` in our production setting. Have you been seeing speed-up while adding more machines to the cluster, just to verify that this is working as expected?

thomcc: I haven't done this in my local runs, mostly since zeppelin doesn't let you change everything about the configuration at runtime. This is here solely because it is that way in the other sync view does it, if you think I should remove it, I will.

src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
thomcc: We transitioned from having the deviceID and UID per-sync to per-ping. UID also didn't exist in even older versions of the schema. It's probably safe to discard these, though.

src/main/scala/com/mozilla/telemetry/views/SyncFlatView.scala
acmiyaguchi: I think it's a good thing to be aware of, but if it's set properly from the airflow side of things it should pose no problems.

PR #291, opened 2017-08-25
Filter the data to prevent OOM in HBaseAddonRecommenderView

Dexterp37:
This patch changes the job to create a shortlist of client_id
and subsession start date retaining only the most recent tuple
per client. The list is then used to filter the subset of data
before it is pushed to HBase.


Dexterp37:
Mauro, please ignore this for now: I've created the PR to run the tests.

Dexterp37:
@maurodoglio this is RFAL :) I tested it on a 5 machines cluster and it doesn't OOM anymore.

src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala
```diff
@@ -74,12 +87,17 @@ object HBaseAddonRecommenderView {
...
               $"scalar_parent_browser_engagement_unique_domains_count",
-              $"active_addons",
-              row_number()
-                .over(Window.partitionBy("client_id")
-                  .orderBy(desc("subsession_start_date")))
-                .alias("clientid_rank"))
-          .where($"clientid_rank" === 1)
+              $"active_addons")
+
+        // Due to a bug [SPARK-12837] in Spark < 2.2.0, we need to disable broadcasting
```
maurodoglio: EMR 5.8.0 comes with Spark 2.2.0, it would be great to verify that it works on that version. If that's the case, then we can remove this config change and tell the airflow job to use emr 5.8.0.

Dexterp37: Thanks for the great tip, I didn't know Spark 2.2.0 was in EMR 5.8.0. On that EMR it works perfectly without this configuration change. I'll change the Airflow configuration and eventually drop this PR once we see it running on Airflow.

Dexterp37:
The Airflow job was changed to use EMR 5.8.0 and everything works fine. We don't need this PR anymore, closing it off in favor of mozilla/telemetry-airflow#163 - Thanks Mauro!

PR #293, opened 2017-08-31
Add crashes to experiment analysis view

fbertsch:
Note that the data for this isn't available yet :/ see https://bugzilla.mozilla.org/show_bug.cgi?id=1388351#c3.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/CrashAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/CrashAnalyzer.scala
@@ -1,0 +1,76 @@
+package com.mozilla.telemetry.experiments.analyzers
+
+import org.apache.spark.sql.{Column, DataFrame}
+import org.apache.spark.sql.functions.{col, lit, sum}
+import org.apache.spark.sql.types._
+
+object CrashAnalyzer {
+
+  val CrashTypes = 
+    "main_crashes"             ::
+    "content_crashes"          ::
+    "gpu_crashes"              ::
+    "plugin_crashes"           ::
+    "gmplugin_crashes"         ::
+    "content_shutdown_crashes" :: Nil
+
+  val ComplexCrashTypes: Map[String, Column] = Map(
+    "main_plus_content_crashes"                -> (sum("main_crashes") + sum("content_crashes")),
+    "content_minus_shutdown_crashes"           -> (sum("content_crashes") - sum("content_shutdown_crashes")),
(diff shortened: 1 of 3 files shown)


src/main/scala/com/mozilla/telemetry/experiments/analyzers/CrashAnalyzer.scala
```diff
@@ -0,0 +1,76 @@
...
+    "subsession_count" :: Nil
+
+  def makeRateName(name: String): String = name.dropRight(2) + "_rate"
+
+  def makeTitle(name: String): String = name.split("_").map(_.capitalize).mkString(" ")
+
+  def getExperimentCrashes(errorAggregates: DataFrame): Seq[MetricAnalysis] = {
+    import errorAggregates.sparkSession.implicits._
+
+    val aggregates = Extras.map{ c => sum(c).as(c) } ++
```
sunahsuh: Would be nice to get the client count in here too as a denominator (we could pass it in to getExperimentCrashes from the exp metadata, or we could merge the hlls which would include crashy clients that we don't have main pings from..)

fbertsch:
@sunahsuh RFAL. A bit nervous about that code that gets the min submission date for a given experiment (since it has no testing), I suppose I should run this on a cluster before we merge.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -54,8 +54,14 @@ object ExperimentAnalysisView {
 
       val spark = getSpark
       val experimentsSummary = spark.read.parquet(conf.inputLocation())
-      val errorAggregates = spark.read.parquet(s"s3://${conf.errorAggregatesBucket()}/$errorAggregatesSuffix")
+
+      val minDate = experimentsSummary.agg(min("submission_date_s3")).first.getAs[String](0)
```
sunahsuh: This would get the min date for all of experimentsSummary, not the particular experiment, which kind of defeats the purpose (given experimentSummary is older than errorAggregates) -- seems like we should either move the experiment id filtering on experimentsSummary to this main function, or move the minDate/errorAggregates filter to `getExperimentMetrics`

sunahsuh:
Oh, also, weirdness re: CI, bleh

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
fbertsch: Ah right, looks like I moved the experiment filtering after I added this bit.

fbertsch:
@sunahsuh RFAL

PR #295, opened 2017-09-04
Add a temporary list of disabled addon ids to main_summary

Dexterp37:
This field is required for the TAAR project and should not
be used for other purposes. See bug 1390814.
This makes use of the soon to be deprecated "addonDetails"
field that contains the list of the user's addons. It will
then populate the "disabled_addon_ids" column with the ids
of the addons that are in "addonDetails" but not in "activeAddons".
This field will go away after addon recommendation for FF57
are no longer needed.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
+++ b/src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
@@ -165,6 +165,57 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
     actual should be (expected)
   }
 
+  "MainSummary legacy addons" can "be summarized" in {
+    val message = RichMessage(
+      "1234",
+      Map(
+        "documentId" -> "foo",
+        "submissionDate" -> "1234",
+        "submission" -> """{
+          "payload": {
+            "addonDetails": {
+              "XPI": {
+                "some-disabled-addon-id": {
+                  "dont-care": "about-this-data",
+                  "we-discard-this": 11
+                },
+                "active-addon-id": {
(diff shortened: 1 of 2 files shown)


src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -165,6 +165,61 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
...
+      "document_id" -> "foo",
+      "disabled_addons_ids" -> List("some-disabled-addon-id")
+    )
+    val actual = applySchema(summary.get, MainSummaryView.buildSchema(scalarDefs, histogramDefs))
+      .getValuesMap(expected.keys.toList)
+    for ((f, v) <- expected) {
+      withClue(s"$f:") { actual.get(f) should be (Some(v)) }
+      actual.get(f) should be (Some(v))
+    }
+    actual should be (expected)
```
sunahsuh: I'm a bit confused about these checks -- shouldn't just this line be enough to check?

Dexterp37: Good point!

Dexterp37:
So, this field is generally ok to use (I changed the comments in the source to match this), just let's not count on it in the future. We can keep the field in the main_summary around, but we might need to slightly change the ping structure to something better than |addonDetails|.

Dexterp37:
Hey @sunahsuh , this is RFAL!

sunahsuh:
Awesome, lgtm!

PR #296, opened 2017-09-06
Permutations UDF

sunahsuh:
The first step in doing permutations tests in the experiments analysis job: this PR adds a method to generate a column with permutations given a count of clients in each branch.

A few implementation notes:
- To save space (especially since I'd like to add the column once and cache the result once per experiment) I'm generating `Byte`s, where the number maps to an index of alphabetically ordered branch names. This also means that we're limited to 128 branches which, from an experimental design perspective, should probably not happen :)
- The tests run pretty fast on my local machine so I'm pretty sure this will be reasonably performant
- I cribbed some of the implementation from PlanOut's [WeightedChoice class](https://github.com/facebook/planout/blob/f779f7863461741d087a457bb167e385b1b6dc2e/alpha/ruby/lib/plan_out/op_random.rb#L53)
- The tests are deterministic so they shouldn't fail randomly unless we change the hashes/implementation
- "Fun" thing I found out (by which I mean, something I spent way too long trying to debug) -- it seems like if you have a case class that's not accessible in the scope where a `SparkSession` is created (say, for example, you create a local case class within a method and use a class-level `SparkSession`), implicit conversions won't work.


src/test/scala/com/mozilla/telemetry/experiments/PermutationsTest.scala
```diff
@@ -0,0 +1,24 @@
...
+    val second = Permutations.weightedGenerator(cutoffs, "hello", 100)("world")
+
+    first should equal (second)
+  }
+
+  "Generated permutations" should "conform to weights" in {
+    val numTrials = 1000000
+    val output = Permutations.weightedGenerator(cutoffs, "hello", numTrials)("world")
+    val counts = output.groupBy(identity).mapValues(_.size)
+    counts(0) / numTrials.toDouble should be > 0.25 * 0.98
```
fbertsch: nit: Use [check numbers against range](http://www.scalatest.org/user_guide/using_matchers#checkingNumbersAgainstARange), here and in `ExperimentAnalysisViewTest`

sunahsuh:
> a question about how this will be used in practice. Given that each client, for each iteration N, is in one experiment branch, will you then be aggregating the histograms for each iteration across each branch? E.g. in permutation 1, groupBy(branch), where branch is the one given by weightedChoice. Then compare the values between branches?

Essentially, yeah -- we'd probably want to generate that array of permutations once since my intuition would be that it'd be a fairly expensive operation, and try to cache the result (although we've run into those issues with caching the experiment dataset so we might want to scrap that and generate on the fly instead.)

sunahsuh:
Thanks for the suggestions -- RFAL!

PR #297, opened 2017-09-06
Heavy users updates

fbertsch:
1. Joining on Sample ID
2. Removing `date`s data before running
3. Using consistent hashing as the partitioning scheme

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/ConsistentPartitioner.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/ConsistentPartitioner.scala
@@ -1,0 +1,108 @@
+package com.mozilla.telemetry.utils
+
+import org.apache.spark.sql.{DataFrame, Row}
+import org.apache.spark.sql.functions.udf
+import org.apache.spark.sql.expressions.UserDefinedFunction
+import scala.math.{floor, log, pow}
+
+/**
+ * A small utility class to enable Consistent Hashing across
+ * a set of partitions.
+ *
+ * Best used in multiples of two, else the partitions will
+ * not be equally sized. If the hash does not result in
+ * well distributed data, the partitions will also not
+ * be equally sized.
+ */
+object ConsistentPartitioner {
+
+  /*
+   * Get the point in [0, 1] that represents this partition.
+   *
+   * To give an idea of the sequence that results, below are
+   * the first 10 inputs and outputs. Note that at all powers
+   * of 2, the range [0, 1] is evenly divided between all partitions.
+   *
+   * For example: With N=2 (so n={0, 1}), you have two partitions.
+   * One in the range [0.0, 0.5] and one in the range [0.5, 1.0].
+   *
+   * Say we then want to add some more partitions, because these
+   * are becoming too large. If we now set N=4, n={0, 1, 2, 3}.
+   * Using the chart below, we can see that n=0 and n=1 have the
+   * same values - 1.0 and 0.5, respectively. n=2 then adds
+   * 0.25, and n=3 adds 0.75. Our partition values are now
+   * {0.25, 0.5, 0.75, 1.0}.
+   *
+   * n=2 then splits the region [0.0, 0.5] into two
+   * where values in [0.0, 0.25] are now shuffled into
+   * partition 2, and values [0.25, 0.5] are shuffled
+   * partition 1. n=3 splits [0.5, 1.0] into two as well
+   * and any values in [0.5, 0.75] will be shuffled into
+   * partition 3.
+   *
+   * Note that: before adding more partitions, values in [0.0, 0.5]
+   * were all shuffled into partition 1. After adding more partitions,
+   * partition 1 has values in [0.25, 0.5]; this means that to join
+   * with that data, partition 1 still only needs to check partition 1.
+   *
+   * Indeed, for any consistently partitioned dataset P1 with M partitions
+   * joining a consistenly partitioned dataset P2 with N partitions:
+   * if M >= N, each partition in P1 only needs to check one partition in P2
+   * if M < N, each partition in P1 will need to check at most N/2 in P2
+   *  (if M%2 == 0 and N%2 == 0)
+   *
+   *
+   * | n  | output |
+   * --------------
+   * | 0  | 1.0    |
+   * | 1  | 0.5    |
+   * | 2  | 0.25   |
+   * | 3  | 0.75   |
+   * | 4  | 0.125  |
+   * | 5  | 0.375  |
+   * | 6  | 0.625  |
+   * | 7  | 0.875  |
+   * | 8  | 0.0625 |
+   * | 9  | 0.1875 |
+   */
+  def getValue(n: Int): Double = {
+    if(n == 0){
+      return 1.0
+    }
+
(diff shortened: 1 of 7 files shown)


fbertsch:
@sunahsuh, good eye! Yes with hashes it won't do what my end goal is. My plan is to let this run as-is, for now. With this partitioning strategy in place, I can empirically test new query plans until I have it more performant. In that sense, this is no worse than any other partitioning strategies.

The added benefit is ranges of sample_ids will, in general, be faster (which is usually how this will be used, until I have bloom filters).

re: deleting prefixes, I made a util function to do that, and update my code and experiments summary to use it. RFAL!

fbertsch:
@sunahsuh one last thing, re: partitioning schema not encoded in path. Yes, but parquet stores metadata about columns, and Spark can do predicate pushdown to ensure it's only reading relevant data. For example, Parquet would store the min/max, Spark would see that the range it's looking for does not overlap with [min, max], and thus it wouldn't read the parquet file at all.

src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -252,6 +257,12 @@ object HeavyUsersView {
     current.write.mode("overwrite").csv(s"$cutoffsPath-$date")
   }
 
+  private def deleteDate(bucket: String, date: String): Unit = {
+    implicit val s3 = S3.at(Region.US_WEST_2)
```
sunahsuh: Do we still need this now?

fbertsch:
@sunahsuh I found differently - I see statistics enabled for any numeric types. As such I switched the type of `sample_id` to `Integer`. Predicate pushdown can still be enabled in joins in Presto, they will just have to specify `where sample_id = 42` (i.e. the join won't be enough). Predicate pushdown will still apply, and will definitely be a huge benefit.

There are alternative query plans that would benefit from this arrangement. For example, partitioning the grouped `Main Summary` dataframe using the same scheme; using a broadcast join, for example, on each partition could result in partition pruning before the join step even takes place. If that fails, one could even read in each partition as a separate dataset and broadcast join each of those. What I'm getting at is there are a variety of strategies that would benefit from this partitioning scheme, and over time I want to fully optimize the joins to take advantage of it.

Also removed the leftover code pieces from the aws removal!

sunahsuh:
> I found differently - I see statistics enabled for any numeric types.

Oh yeah, I was definitely using the wrong command on parquet-tools :) I see the stats outputs now.

So yeah, I definitely agree with you on the advantages of your approach generally but I don't think you've addressed my main concern, which is why the consistent partitioning scheme would be better than using the RangePartitioner -- I don't see why it wouldn't give us all the benefits you describe without us having to write our own code 🤔❓❓

fbertsch:
@sunahsuh two points on improvement over RangePartitioner:

RangePartitioner will divide the partition column (weighted) into equivalent ranges. Since we have equal sized partitions for each `sample_id`, this is roughly the same as partitioning by `sample_id / N`, which leads to the problem that as we add/remove a partition, it moves data out of every single partition. With the Consistent Hashing partitioner, it only affects one partition.

We can even partition data before joins to take advantage of this property. For example, if the data on disk is partitioned into N partitions using consistent hashing, we can also using consistent hashing for a dataset in memory into M partitions (i.e. that spark already read in). Then we can guarantee that, if M > N, for each partition in M we only have to check one partition for data in N. If M < N there are still benefits, but we can't give any guarantees. In both cases, there is less data to be joined on each partition than if we used RangePartitioner (unless, of course, M=1). If M = N, then both RangePartitioner and Consistent Hashing Partitioner behave the same.

Also, I do think we can switch `sample_id` in `main_summary` to an int, and use predicate pushdown on joins! See https://stackoverflow.com/questions/31476335/hive-1-1-0-alter-table-partition-type-from-int-to-string.

sunahsuh:
>  which leads to the problem that as we add/remove a partition, it moves data out of every single partition. With the Consistent Hashing partitioner, it only affects one partition.

Unless I'm reading the code wrong, this partitioner implementation is essentially equivalent to range partitioning, especially since sample_id/100 is being used directly as the "hash", and if we increase the number of partitions to N + 1, then the first partition will have less sample_ids moved "out" than partition N. For true consistent partitioning it looks like we'd need to add more virtual bucket points in the ring.

In any case, even with true consistent partitioning, here's what we'd see:
- We increase the number of partitions by 1 (probably our most common case). With consistent partitioning, an equivalent number of sample_ids from each partition get moved to the new partition. With range partitioning, some sample_ids would be moved from each partition, but more from the last few than the first few.
- In the case where we're selecting out one particular sample id and joining it across days, we'd see no functional difference.
- In the case where we're joining the entirety of two days across a partition number increase, we would still have the case where any partition from before the increase would have its data spread across two partitions after the increase in *both cases.*

> For example, if the data on disk is partitioned into N partitions using consistent hashing, we can also using consistent hashing for a dataset in memory into M partitions (i.e. that spark already read in). Then we can guarantee that, if M > N, for each partition in M we only have to check one partition for data in N.

Intuitively, I don't see how that could be true -- there's at least one partition M that has to look across possibly 2+ partitions, no? And if that case, that'd be much worse since the workload in the join would be unevenly distributed and we'd have to wait for that one uneven task. With the range partitioner, you'd have to look across at most 2 partitions, for M > N, and 1 partition if M = cardinality of the hash column.

sunahsuh:
Okay, these replies are becoming novel-length -- let's jump into vidyo and check each other's assumptions/hash this out? (PUN INTENDED)

src/main/scala/com/mozilla/telemetry/utils/ConsistentPartitioner.scala
```diff
@@ -0,0 +1,82 @@
...
+   */
+  def getPartitionRing(numPartitions: Int): Seq[Double] = {
+    (0 to (numPartitions - 1)).map(getValue _).sorted
+  }
+
+  /*
+   * Given a sorted list of partition values in [0, 1], return a function
+   * that takes a sample value and outputs the partition that sample belongs in.
+   */
+  def getPartioningFunc(numPartitions: Int, hashFunc: String => Double, defaultPartition: Int = 0): UserDefinedFunction = {
```
sunahsuh: typo in `getPartioningFunc`

src/main/scala/com/mozilla/telemetry/utils/ConsistentPartitioner.scala
```diff
@@ -0,0 +1,82 @@
...
+   * | 0  | 1.0    |
+   * | 1  | 0.5    |
+   * | 2  | 0.25   |
+   * | 3  | 0.75   |
+   * | 4  | 0.125  |
+   * | 5  | 0.375  |
+   * | 6  | 0.625  |
+   * | 7  | 0.875  |
+   * | 8  | 0.0625 |
+   * | 9  | 0.1875 |
```
sunahsuh: Given the many different "interpretations" of Consistent Hashing across the interwebs, I think it makes sense to expand on this comment. A few example partition rings for different numPartitions would probably be useful, and something about one partition being divided in half for each increase in numPartitions.

PR #298, opened 2017-09-06
Bug 1397416 - Add histogram uptake to main summary

mythmon:
@fbertsch r?


fbertsch:
Hmm, the build is failing because the histogram is being filtered out in [0]. @mythmon, I'll take a look at this tomorrow morning.

[0] https://github.com/mythmon/telemetry-batch-view/blob/3bd1830a30b339a148b072e8403ac57f3dd46569/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L829

fbertsch:
Figured out what the problem is. We do not include any Categorical histograms - that will have to be fixed before we can merge this :/

https://bugzilla.mozilla.org/show_bug.cgi?id=1397913

fbertsch:
Figured out what the problem is. We do not include any Categorical histograms - that will have to be fixed before we can merge this :/

https://bugzilla.mozilla.org/show_bug.cgi?id=1397913

fbertsch:
Once we merge #302, the build should pass.

PR #299, opened 2017-09-07
Bug 1381653 - Add SANDBOX_REJECTED_SYSCALLS keyed histogram to Main Summary

jld:
This patch hasn't been tested; I'm hoping that's not necessary for a simple change like this.


fbertsch:
@jld, the build is failing because we initially did not want to include Count Histograms (they should really be scalars).

Is there a bug out for eventually making this probe a scalar? If there is, I'm okay with adding it as-is, because we can eventually remove it once it is a scalar. All we need to do is add a case CountHistograms and a comment about why we're including them, exactly the same as we did for FlagHistograms. See [0].

[0] https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala#L834

fbertsch:
@jld, does this still need to be added?

fbertsch:
Closing for now. Chat in bug 1381653 isn't clear on which direction this PR will go.

PR #300, opened 2017-09-07
Update normandy recipe eligibility in ExperimentSummaryView

sunahsuh:
This PR updates the experiment summary job to:
- Add a list of excluded recipes (first use is to exclude the screenshots release rollout)
- Include opt-out studies per [bug 1395362](https://bugzilla.mozilla.org/show_bug.cgi?id=1395362)

Unfortunately, opt-out studies have the slug in a different field than pref rollouts so that adds a bit of not-so-pretty munging.

Briefly, I've removed the "action" filter from the normandy url that we're pulling down so we're not getting just pref-flip experiments, normalized the slug/name fields, added opt-out studies to the allowed actions filter and added an excluded experiments list.

### DIFF
--- a/src/test/resources/normandy_api_result.json
+++ b/src/test/resources/normandy_api_result.json
@@ -201,5 +201,142 @@
       },
       "comment": "looks good to me"
     }
+  },
+  {
+    "id": 225,
+    "last_updated": "2017-08-30T20:17:59.565943Z",
+    "name": "Opt-out Nightly Enrollment Test - Nothing Burger",
+    "enabled": false,
+    "is_approved": false,
+    "revision_id": "8a013fc441e6396836ac7222cae63756c7818b32be4850902a431dbfa838132c",
+    "action": "opt-out-study",
+    "arguments": {
+      "name": "nightly-nothing-burger-1",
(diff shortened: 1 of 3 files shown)


src/test/resources/normandy_api_result.json
```diff
@@ -201,5 +201,142 @@
...
+        "first_name": "",
+        "last_name": "",
+        "email": "mgrimes@mozilla.com"
+      },
+      "approved": true,
+      "approver": {
+        "id": 11,
+        "first_name": "Kamyar",
+        "last_name": "Ardekani",
+        "email": "kardekani@mozilla.com"
```
fbertsch: Please anonymize these emails in the test files.

sunahsuh: This is pulled down from a public API, the one whose URI is in the code..

sunahsuh:
> Question: Why use def for the class vars, instead of val?

Probably just a holdover from jobName/schemaVersion using def (which is consistent in nearly all our jobs), and that the lines followed those lines. It doesn't make a practical difference but each access does execute each time so we should probably just change all the defs that should be vals in all the classes unless there's a compelling reason not to do so.

PR #303, opened 2017-09-14
Fix the MainSummaryView to correctly find addonDetails

Dexterp37:
Turns out I was looking up the field in the wrong data structure: `payload` is not the ping payload, but rather the message payload. I've changed this accordingly and I'm re-generating the dataset in a test bucket.


Dexterp37:
Thanks Sunah! Looks like this is working now, I saved a test parquet file in s3://telemetry-test-bucket/main_summary/v4/submission_date_s3=20170912/ and cross checked.

PR #307, opened 2017-09-25
Bug 1402747 - Fix missing OS field

fbertsch:
I did confirm my added test raises the same error as I saw in production.

This change will take OS from the first ping for a client it finds
it from. It also applies the change to client_id and normalized_
channel, just in case.

This change also makes OS optional. Why is that?
1. Top-level OS field comes env/system/os/name [0]
2. env/system/os/name can be missing [1]

[0] https://github.com/mozilla-services/lua_sandbox_extensions/blob
/master/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L213
[1] https://github.com/mozilla-services/mozilla-pipeline-schemas/
blob/master/schemas/telemetry/main/main.4.schema.json#L778

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -483,7 +483,7 @@ object LongitudinalView {
     val builder = SchemaBuilder
       .record("Submission").fields()
         .name("client_id").`type`().stringType().noDefault()
-        .name("os").`type`().stringType().noDefault()
+        .name("os").`type`().optional().stringType()
         .name("normalized_channel").`type`().stringType().noDefault()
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -483,7 +483,7 @@ object LongitudinalView {
     val builder = SchemaBuilder
       .record("Submission").fields()
         .name("client_id").`type`().stringType().noDefault()
-        .name("os").`type`().stringType().noDefault()
+        .name("os").`type`().optional().stringType().noDefault()
```
mreid-moz: I think you need to leave off the `noDefault` when using `optional()`

PR #309, opened 2017-09-26
Add select only used columns, add permutations column and persist res…

sunahsuh:
…ulting dataset

I ran a couple different trials to tune this and using a 10-node cluster this dropped the job's runtime significantly.

One note: I reduced the number of permutations drastically -- just *generating* 5000 increased the job time drastically (largely because of storage, I think, and not computation) and we probably only need 100 to get a decent confidence interval for the p-value in most cases. If we need more for specific metrics, we can generate them on the fly, but I think we can probably start with 100 and work our way up if we need to.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
@@ -6,6 +6,7 @@ import com.mozilla.telemetry.metrics._
 import com.mozilla.telemetry.utils.getOrCreateSparkSession
 import org.apache.spark.sql.functions.{col, min, udf}
 import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.storage.StorageLevel
 import org.rogach.scallop.ScallopConf
 
 import scala.util.{Failure, Success, Try}
@@ -15,8 +16,25 @@ object ExperimentAnalysisView {
   def errorAggregatesPath = "error_aggregates/v2"
   def jobName = "experiment_analysis"
   def schemaVersion = "v1"
+  // This gives us ~120MB partitions with the columns we have now. We should tune this as we add more columns.
+  def rowsPerPartition = 25000
 
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
+
+  // scalar_ and histogram_ columns are automatically added
+  private val usedColumns = List(
+    "client_id",
+    "experiment_id",
+    "experiment_branch"
+  )
+
+  implicit class ExperimentDataFrame(df: DataFrame) {
+    def selectUsedColumns: DataFrame = {
+      val columns = (usedColumns
+      ++ (df.schema.map(_.name).filter { s: String => s.startsWith("histogram_") || s.startsWith("scalar_") }))
+      df.select(columns.head, columns.tail: _*)
+    }
+  }
 
   // Configuration for command line arguments
   class Conf(args: Array[String]) extends ScallopConf(args) {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -18,6 +19,21 @@ object ExperimentAnalysisView {
...
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
 
+  // scalar_ and histogram_ columns are automatically added
+  private val usedColumns = List(
+    "client_id",
+    "experiment_id",
+    "experiment_branch"
+  )
+
+  implicit class ExperimentDataFrame(df: DataFrame) {
```
fbertsch: I've been leaning away from using implicits [0], becuase they make it hard to follow the code. Consider changing this, but not necessary if you decide against it.

[0] https://github.com/databricks/scala-style-guide#implicits

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -55,6 +71,8 @@ object ExperimentAnalysisView {
       logger.info(s"Aggregating pings for experiment $e")
 
       val spark = getSpark
+      spark.sqlContext.sql(s"set spark.sql.shuffle.partitions=${spark.sparkContext.defaultParallelism}")
```
fbertsch: Any reason not to push this to the `getOrCreateSparkSession` function?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -117,35 +135,58 @@ object ExperimentAnalysisView {
...
 
+  // Adds a column for permutations, repartitions if warranted, and persists the result for quicker access
+  def getPersistedDataFrame(experimentsSummary: DataFrame, metadata: Array[MetricAnalysis], experiment: String): DataFrame = {
+    val topLevelMetadata = metadata.filter(_.subgroup == "All")
+    val totalPings = topLevelMetadata.flatMap(_.statistics.get.filter(_.name == "Total Pings").map(_.value)).sum
+    val branchCounts = topLevelMetadata.map {
+      r => r.experiment_branch ->
+        (r.statistics.get.collectFirst {case s: Statistic if s.name == "Total Clients" => s} match {
+          case Some(s: Statistic) => s.value.toLong
+          case _ => 0L
```
fbertsch: Is this case expected? Should we raise an error instead?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -117,35 +135,58 @@ object ExperimentAnalysisView {
...
+        })
+    }.toMap
+
+    val withPermutations = addPermutations(experimentsSummary.selectUsedColumns, branchCounts, experiment)
+    (totalPings / 10000 match {
```
fbertsch: nit: magic number `100000`

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -117,35 +135,58 @@ object ExperimentAnalysisView {
...
         }
     }
 
-    val metadata = ExperimentAnalyzer.getExperimentMetadata(experimentsSummary).collect()
     val crashes = CrashAnalyzer.getExperimentCrashes(errorAggregates)
     (metadata ++ metrics ++ crashes).toList
   }
 
+  // Adds a column for permutations, repartitions if warranted, and persists the result for quicker access
+  def getPersistedDataFrame(experimentsSummary: DataFrame, metadata: Array[MetricAnalysis], experiment: String): DataFrame = {
```
fbertsch: Consider renaming to include the full scope of this functinon: e.g. `addPermutationsAndPersist`

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
sunahsuh: Yeah, I definitely understand and generally agree with the reasoning, but this is scoped inside the object where it's used and it's, I hope, pretty obvious what it does.

sunahsuh:
RFAL!

PR #310, opened 2017-09-26
Validate experiments data before aggregating

sunahsuh:
- Adds a method to histogram definitions returning its valid bucket values
- Validates experiments data before aggregation to filter out negative values and validate incoming buckets

Closes mozilla/experiments-viewer#225 and mozilla/experiments-viewer#260


src/main/scala/com/mozilla/telemetry/experiments/analyzers/HistogramAnalyzer.scala
```diff
@@ -28,6 +28,14 @@ class HistogramAnalyzer(name: String, hd: HistogramDefinition, df: DataFrame)
   extends MetricAnalyzer[Int](name, hd, df) {
   override type PreAggregateRowType = PreAggHistogramRow
   override val aggregator = UintAggregator
+  val buckets = hd.getBuckets
+  def validateRow(row: PreAggregateRowType): Boolean = {
```
fbertsch: Nit: A comment about what kind of validation this is doing would be helpful

PR #314, opened 2017-09-28
Sense for previous data in heavy_users

fbertsch:


src/main/scala/com/mozilla/telemetry/views/HeavyUsersView.scala
```diff
@@ -239,6 +242,16 @@ object HeavyUsersView {
     (withCutoffs, cutoff)
   }
 
+  def senseExistingData(existingData: Dataset[HeavyUsersRow], date: String, firstRun: Boolean): Unit = {
+    if(!firstRun) {
+      val lastDateRow = existingData.select(max(col("submission_date_s3"))).collect()
+      val lastDate = fmt.parseDateTime(lastDateRow(0).getString(0))
```
sunahsuh: Not something to fix in this PR, but I just noticed we're using both github's nscala_time and joda.time in the repo for similar purposes. Looks like joda.time is more widely used, so we should migrate nscala_time uses and remove it from the build.

fbertsch:
@sunahsuh unfortunately I did mark a day complete in Airflow, but it still failed to run.

I do agree with the sentiment, but I really need to get this running now and not fight with the tooling. Sometime in q4 I'll try and use Airflow to solve this problem instead.

PR #316, opened 2017-09-29
Bug 1381840 - Add 1 day Firefox retention view using HLL

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1381840)

This PR is used in conjunction with [mozilla/python_mozetl #133](https://github.com/mozilla/python_mozetl/pull/133) to generate a 1 day Firefox retention dataset.

The mozetl script will preprocess the data in the following stages:
1. Convert `new_profile_ping_parquet` to a subset of `main_summary_v4`
2. Join `new_profile` with `main_summary` and select the subset for the retention period
3. Build a select expression for cleaning and derived columns and apply the transformation
4. Write the cleaned intermediate data into HDFS

This job will carry out the aggregation of the final data, and store the data to the final output location.

As opposed to previous versions of `retention_dev` which use the `GenericCountView`, `retention_dev_v3` uses the HLL UDFs directly. The GenericCountView was immensely useful for creating the first two revisions of this dataset, but deciding on a good generic API for passing in aggregate functions via the command line is hard.

It would be nice to have access to the `spark-hyperloglog` library via `py4j`, but there are [problems with calling `hll_cardinality`](https://bugzilla.mozilla.org/show_bug.cgi?id=1305087) that I have verified for myself. Now that pyspark can be installed via pip, it would be attractive to extend `spark-hyperloglog` with a pyspark binding. 

Finally, the entire job wasn't written directly as a scala view because the cleaning subroutines and structure of the `daily_retention` job is heavily influenced by the third revision of the `churn` code. The testing harness used in `churn` can be reused to validate pre-aggregate transformations of the intermediate data.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/RetentionView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/RetentionView.scala
@@ -1,0 +1,89 @@
+package com.mozilla.telemetry.views
+
+
+import com.mozilla.telemetry.utils.UDFs._
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.functions._
+import org.rogach.scallop._
+
+
+object RetentionView {
+  class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Run date for this job", required = true)
+    val input = opt[String]("input", descr = "Source for parquet data", required = true)
+    val bucket = opt[String]("bucket", descr = "output bucket", required = true)
+    val prefix = opt[String]("prefix",
+      descr = "output prefix",
+      required = false,
+      default = Some("retention/v1")
+    )
+    val hllBits = opt[Int](
+      "hll-bits",
+      descr = "Number of bits to use for hll. 13 bits is 8192 bytes with an error of 0.0115. Defaults to 13.",
+      required = false,
+      default = Some(13)
+    )
+    verify()
+  }
+
+  val dimensions: List[String] = List(
+    "subsession_start",
+    "profile_creation",
+    "days_since_creation",
+    "channel",
+    "app_version",
+    "geo",
+    "distribution_id",
+    "is_funnelcake",
+    "source",
+    "medium",
+    "content",
+    "sync_usage",
+    "is_active"
+  )
+
+  val metrics: List[String] = List(
+    "usage_hours",
+    "sum_squared_usage_hours",
+    "total_uri_count",
+    "unique_domains_count"
+  )
+
+
+  def transform(dataframe: DataFrame, hllBits: Int): DataFrame = {
+    val expr = List(s"hll_create(client_id, $hllBits) as hll") ++ dimensions ++ metrics
+
+    dataframe.selectExpr(expr:_*)
+      .groupBy(dimensions.head, dimensions.tail:_*)
+      .agg(
+        HllMerge(col("hll")).as("hll"),
+        sum("usage_hours").as("usage_hours"),
+        sum("sum_squared_usage_hours").as("sum_squared_usage_hours"),
+        sum("total_uri_count").as("total_uri_count"),
+        avg("unique_domains_count").as("unique_domains_count")
+      )
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args)
+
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/RetentionView.scala
```diff
@@ -0,0 +1,81 @@
...
+        sum("usage_hours").as("usage_hours"),
+        sum("sum_squared_usage_hours").as("sum_squared_usage_hours"),
+        sum("total_uri_count").as("total_uri_count"),
+        avg("unique_domains_count").as("unique_domains_count")
+      )
+
+    result
+      .write
+      .mode("overwrite")
+      .parquet(s"s3://${conf.bucket()}/${conf.prefix()}/start_date=${conf.date()})")
```
mreid-moz: There's an extra `)` near the end, please remove it:
```
.parquet(s"s3://${conf.bucket()}/${conf.prefix()}/start_date=${conf.date()})")
                                                                           ^ here

**Comment by mreid-moz** [comment]
@fbertsch do you have any suggestions for how to incorporate arbitrary aggregates into `GenericCountView`?

**Comment by fbertsch** [comment]
Arbitrary aggregations could be [added as a prop map](https://github.com/scallop/scallop/wiki/Option-types-and-definitions#property-arguments). The call would then look like `--aggregations sum=usage_hours,sum_squared_usage_hours,total_uri_count avg=unique_domains_count`. You could then use [Scala reflection](https://docs.scala-lang.org/overviews/reflection/overview.html#accessing-and-invoking-members-of-runtime-types) to get the function that is the key of the `--arguments` call.

An alternative would be a specific argument per aggregation type e.g. `--avg unique_domains_count`, but that could get unwieldy beyond two or three.

Re: Pyspark hyperloglog, Saptarshi worked on this back in the fall but couldn't get it fully working: see https://bugzilla.mozilla.org/show_bug.cgi?id=1305087

**Comment by acmiyaguchi** [comment]
@mreid-moz RFAL

File: src/main/scala/com/mozilla/telemetry/views/RetentionView.scala [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+      .agg(
+        HllMerge(col("hll")).as("hll"),
+        sum("usage_hours").as("usage_hours"),
+        sum("sum_squared_usage_hours").as("sum_squared_usage_hours"),
+        sum("total_uri_count").as("total_uri_count"),
+        avg("unique_domains_count").as("unique_domains_count")
+      )
+  }
+
+  def main(args: Array[String]) {
```
mreid-moz: Please add an argument for the number of hll bits, with documentation similar to [GenericCountView here](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala#L42-L46)

acmiyaguchi:
@mreid-moz RFAL

src/test/scala/com/mozilla/telemetry/views/RetentionViewTest.scala
```diff
@@ -0,0 +1,106 @@
...
+    sample.copy(client_id = "1_dup", channel = "release"),
+    sample.copy(client_id = "1_dup", channel = "release", unique_domains_count = 0),
+    sample.copy(client_id = "2", channel = "release"),
+    sample.copy(client_id = "3", channel = "release"),
+    sample.copy(client_id = "4", channel = "beta"),
+    sample.copy(client_id = "5_dup", channel = "beta"),
+    sample.copy(client_id = "5_dup", channel = "beta"),
+    sample.copy(client_id = "5_dup", channel = "beta"),
+    sample.copy(client_id = "5_dup", channel = "beta")
+
```
mreid-moz: nit: remove blank line

mreid-moz:
Note that we have [Bug 1405066](https://bugzilla.mozilla.org/show_bug.cgi?id=1405066) to track adding support for aggregations to `GenericCountView`.

acmiyaguchi:
Rebased and squashed, waiting for tests to pass.

mreid-moz:
Looks good!

PR #318, opened 2017-10-05
Move date deletion to after sensing in heavy_users

fbertsch:


fbertsch:
This won't work as-is, removing this PR for now.

PR #319, opened 2017-10-06
Add e10sMultiProcess to main_summary

fbertsch:


fbertsch:
Whoever gets to it first can r :)

PR #323, opened 2017-10-16
Bug 1390707 - Add user_prefs entry

mreid-moz:
Add an item to the user_prefs struct for 'extensions.legacy.enabled'.


mreid-moz:
The CI build failed with `java.lang.OutOfMemoryError: Metaspace`. It appears to be unrelated to this change, but we should fix that.

See [Bug 1408432](https://bugzilla.mozilla.org/show_bug.cgi?id=1408432)

mreid-moz:
Closed in favour of #338

PR #324, opened 2017-10-18
Bug 1409861 - Handle sync pings that have os information at top level

thomcc:
I think this also slightly cleans up the sync ping code, but it's still pretty gnarly (sorry).


acmiyaguchi:
Looks like intermittent failure on Travis, I'm kicking off the tests again.

PR #325, opened 2017-10-19
Bug 1410152 - Handle android sync telemetry submitting wrong document version

thomcc:
See [bug 1410152 in bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1410152) for description.


thomcc:
It's not actually needed, since Android never submit's event telemetry (and now it reports the correct version -- so if it started submitting event telemetry it would do the right thing), but it's probably better to be consistent (so I've done that).

PR #328, opened 2017-10-31
[Bug 1375018] Telemetry batch view add maven local repo to replace sonatype

haroldwoo:
- Telemetry batch view add maven local repo to replace sonatype
- Add circleci build to automatically publish jar to local maven repo on commits to master (this will require extra setup on the github side)


fbertsch:
@haroldwoo looks like the build is failing, can you check it out?

haroldwoo:
yeah its the tests are too large.  The PR i have opts out of tests, but the
initial run was some default that circleci uses.

Your build has exceeded the memory limit of 4G on 1 container. The results
of this build are likely invalid. We have taken a snapshot of the memory
usage at the time, which you can find in a build artifact named
`memory-usage.txt`. The RSS column in this file shows the amount of memory
used by each process, measured in kilobytes.

build.sbt
```diff
@@ -1,6 +1,6 @@
 javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint")
 
-resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"
+resolvers += "S3 local maven snapshots" at "s3://net-mozaws-data-us-west-2-ops-mavenrepo/snapshots"
```
fbertsch: Can we put the s3 location in a variable and use it both here and on line 76?

README.md
```diff
@@ -83,3 +84,4 @@ System.setProperty("spark.sql.warehouse.dir", "file:///C:/somereal-dir/spark-war
 
 See [SPARK-10528](https://issues.apache.org/jira/browse/SPARK-10528). Run "winutils chmod 777 /tmp/hive" from a privileged prompt to make it work.
 
+Any commits to master should also trigger a circleci build that will do the sbt publishing for you to our local maven repo in s3.
```
fbertsch: Can you update the section above that says to use `sbt assembly`?

PR #333, opened 2017-11-01
Excluded concluded experiments from ExperimentSummaryView

sunahsuh:
While we wait for mozilla/normandy#804, this is a workaround that should allow us to exclude definitely concluded experiments while staying backfill-friendly by checking the recipe's last updated date. This way, we might get some false positives (concluded experiments that were, for some reason, updated after being disabled) but we'll never have false negatives.

This should significantly speed up both jobs since we have quite a few concluded experiments at this point.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
@@ -1,5 +1,7 @@
 package com.mozilla.telemetry.views
 
+import java.util.Date
+
 import com.mozilla.telemetry.utils.{deletePrefix, getOrCreateSparkSession}
 import org.apache.spark.sql.functions._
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
```diff
@@ -125,10 +131,19 @@ object ExperimentSummaryView {
     }
   }
 
+  def shouldProcessExperiment(r: NormandyRecipe, date: Date) = {
+    experimentsQualifyingAction.contains(r.action) &&
+    !excludedExperiments.contains(r.arguments.slug.get) &&
```
fbertsch: nit: should this and the following line be indented once more?

sunahsuh:
> I especially like the alphabetical reordering of the imports :)

Hah, I think IntelliJ must have done that on its own

PR #335, opened 2017-11-02
Revert "Bug 1375018 replace sonatype with local s3 maven repo"

sunahsuh:
We need to update the path to the jar on all of our airflow bash scripts before we re-revert this


fbertsch:
This was fixed with #336.

PR #336, opened 2017-11-02
fixing sbt assembly target and badge

haroldwoo:
fixes sbt assembly target to target/scala-2.11
fixes badge in readme


fbertsch:
@haroldwoo did you verify that this will work with our current Airflow shell scripts?

PR #339, opened 2017-11-06
Bug 1344259: Support for content and addon events

sunahsuh:
- Adds event extraction from content and dynamic processes in main summary (note: we're sticking the process name into the map_values field due to the schema evolution bug for nested objects)
- Moves the process name into a top-level field in the events table
- Not related to the bug, but since I was looking at it: adds the "experiments" block to the events table

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/Events.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/Events.scala
@@ -70,14 +70,16 @@ object Event {
       case _ => None
     }
   }
+
+  def fromList(event: List[Any], process: String): Option[Event] = {
+    fromList(event).map(e => e.copy(
+      mapValues = Some(Map("telemetry_process" -> process) ++ e.mapValues.getOrElse(Map()))))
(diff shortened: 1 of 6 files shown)


src/main/scala/com/mozilla/telemetry/utils/Events.scala
```diff
@@ -70,14 +70,15 @@ object Event {
       case _ => None
     }
   }
+
+  def fromList(event: List[Any], process: String): Option[Event] = {
+    fromList(event).map(e => e.copy(mapValues = Some(Map("process" -> process) ++ e.mapValues.getOrElse(Map()))))
```
mreid-moz: Do we need to check if `"process"` is already in the  map? Should we use a name that's less likely to be used in the map, or maybe reserve a prefix for system-added keys?

sunahsuh: I changed it to "telemetry_process", which should be less prone to overwriting

PR #340, opened 2017-11-06
Bug 1411945 - remove E10sExperiment

relud:
r? @mreid-moz 

as far as I can tell, these two files are the only ones in this repo with any reference to E10sExperiment, and there were not tests associated with it.

Is there anything else I need to do to remove this, or is removing these files enough?


relud:
resolves [Bug 1411945](https://bugzilla.mozilla.org/show_bug.cgi?id=1411945)

PR #341, opened 2017-11-06
Confidence intervals

sunahsuh:
This includes implementations of bootstrapped confidence intervals for both histograms and scalars but only turns them on for scalars in the job since the current histogram implementation is too slow to be pragmatically useful.

Briefly, the metric analyzer is responsible for providing resampled aggregates to the DescriptiveStatistics class, which merely runs the statistic on each aggregate, sorts the results and uses the simple percentile method to return the confidence interval.

The histogram resampler uses the built-in spark sample-with-replacement method, which is effective but slow.

The scalar resampler, on the otherhand, distributes resampling into separate spark tasks for each node and therefore is able to run much faster since we can resample directly from the aggregated histogram. With the histogram CIs turned off and the scalar CIs turned on, it looks like the jobs take ~4x, which is fine since the job now takes ~2 hours to run.

I've added an "isCategorical" attribute to each metric definition for numeric vs categorical data, and am skipping bootstrapping/summary statistics for categorical metric types since those aren't particularly meaningful.

PS I am so tired of looking at this code

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
@@ -90,9 +90,50 @@ case class LongScalarMapRow(experiment_id: String, branch: String, subgroup: Str
 case class StringScalarMapRow(experiment_id: String, branch: String, subgroup: String,
                               metric: Option[Map[String, Long]]) extends PreAggregateRow[String]
 
+trait AggregateResampler {
+  type PreAggregateRowType
 
-class BooleanScalarAnalyzer(name: String, md: ScalarDefinition, df: DataFrame)
-  extends MetricAnalyzer[Boolean](name, md, df) {
+  // This implementation reaggregates directly from the aggregated histogram, which works when we expect one sample
+  // per ping, such as when we're resampling scalar aggregates
+  def resample(ds: Dataset[PreAggregateRowType],
+               aggregations: List[MetricAnalysis],
+               iterations: Int = 1000): Map[MetricKey, List[MetricAnalysis]] = {
+    aggregations.map { a =>
+      val total = a.histogram.values.map(_.count).sum
+      // We could use the pdf field directly to cumulatively sum but we'd likely run into weird floating point addition
+      // issues
+      val cdf =  a.histogram
+        .map { case (k, v) => k -> v.count }
+        .toList
+        .scanLeft(0L, 0.0) { case((_: Long, c: Double), (k: Long, v: Double)) => (k, c + v)}
+        .tail
+        .map { case (k, v) => (k, v / total) }
+      (a, total, cdf)
+    }.map { case (m, total, cdf) =>
+      val aggs = ds.sparkSession.sparkContext.parallelize(1 to iterations).map { x: Int =>
+        val agg = resampleCdf(cdf, x, total.toInt)
+        m.copy(histogram = agg.map {case (k, v) => k -> HistogramPoint(v / total, v.toDouble, None)})
+      }.collect.toList
+      m.metricKey -> aggs
+    }.toMap
+  }
+
+  private def resampleCdf(cumulativeWeights: List[(Long, Double)], seed: Int, n: Int): Map[Long, Long] = {
+    val rng = new scala.util.Random(seed)
+    (0 to n).foldLeft(Map.empty[Long, Long]) { (agg, _) =>
+      val rand = rng.nextDouble()
+      val sample = cumulativeWeights.collectFirst { case(k, c) if rand <= c => k } match {
+        case Some(k) => k
+        case _ => throw new Exception("Something very strange happened in constructing the cdf")
+      }
+      agg + (sample -> (agg.getOrElse(sample, 0L) + 1L))
+    }
+  }
+}
+
+
+class BooleanScalarAnalyzer(name: String, md: ScalarDefinition, df: DataFrame, bootstrap: Boolean = false)
+  extends MetricAnalyzer[Boolean](name, md, df, bootstrap) with AggregateResampler {
   override type PreAggregateRowType = BooleanScalarMapRow
   val aggregator = BooleanAggregator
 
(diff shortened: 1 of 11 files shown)


src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -90,9 +90,50 @@ case class LongScalarMapRow(experiment_id: String, branch: String, subgroup: Str
...
+  // per ping, such as when we're resampling scalar aggregates
+  def resample(ds: Dataset[PreAggregateRowType],
+               aggregations: List[MetricAnalysis],
+               iterations: Int = 1000): Map[MetricKey, List[MetricAnalysis]] = {
+    aggregations.map { a =>
+      val total = a.histogram.values.map(_.count).sum
+      // We could use the pdf field directly to cumulatively sum but we'd likely run into weird floating point addition
+      // issues
+      val cdf =  a.histogram
+        .map {case (k, v) => k -> v.count }
```
fbertsch: nit: space before case

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -90,9 +90,50 @@ case class LongScalarMapRow(experiment_id: String, branch: String, subgroup: Str
...
+      m.metricKey -> aggs
+    }.toMap
+  }
+
+  private def resampleCdf(cumulativeWeights: List[(Long, Double)], seed: Int, n: Int): Map[Long, Long] = {
+    val rng = new scala.util.Random(seed)
+    (0 to n).foldLeft(Map.empty[Long, Long]) { (agg, _) =>
+      val rand = rng.nextDouble()
+      val sample = cumulativeWeights.collectFirst { case(k, c) if rand <= c => k } match {
+        case Some(b) => b
```
fbertsch: nit: s/b/k

fbertsch: Could also just call `.get` instead of pattern matching, since that would bubble up an exception if there was no value. Downside would be harder to trace an error though.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
```diff
@@ -90,9 +90,50 @@ case class LongScalarMapRow(experiment_id: String, branch: String, subgroup: Str
...
+      // issues
+      val cdf =  a.histogram
+        .map {case (k, v) => k -> v.count }
+        .toList
+        .scanLeft(0L, 0.0) { case((_: Long, c: Double), (k: Long, v: Double)) => (k, c + v)}
+        .tail
+        .map { case (k, v) => (k, v / total) }
+      (a, total, cdf)
+    }.map { case (m, total, cdf) =>
+      val aggs = ds.sparkSession.sparkContext.parallelize(1 to iterations).map { x: Int =>
```
fbertsch: An option to increase efficiency is to parallelize just the number of `spark.default.parallelism`, and use a `flatMap` and have each task just return `(spark.default.parallelism / iterations) + 1` `MetricAnalysis` objects. No change needed here.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -79,6 +82,7 @@ object ExperimentAnalysisView {
 
       val experimentsSummary = spark.read.option("mergeSchema", "true").parquet(conf.inputLocation())
         .where(col("experiment_id") === e)
+        .where(col("submission_date_s3") <= date)
```
fbertsch: Should this have been there awhile ago?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -139,58 +143,33 @@ object ExperimentAnalysisView {
...
 
     val metrics = metricList.flatMap {
       case (name: String, md: MetricDefinition) =>
         md match {
           case hd: HistogramDefinition =>
-            new HistogramAnalyzer(name, hd, persisted).analyze()
+            new HistogramAnalyzer(name, hd, persisted, bootstrapHistograms).analyze()
           case sd: ScalarDefinition =>
-            ScalarAnalyzer.getAnalyzer(name, sd, persisted).analyze()
+            ScalarAnalyzer.getAnalyzer(name, sd, persisted, true).analyze()
```
fbertsch: Should this also be a flag, except a flat to turn them off?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
sunahsuh: hmm, one of the things I wanted was to be able to preserve determinism across runs, and that would affect the result in a way that'd be very hard to trace back. Also, in our nightly runs parallelism is currently 640, which is close enough to the default 1000 that I suspect it wouldn't make *that* big of a difference?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
sunahsuh: Yes, I should have noted that (but totally forgot I put that in there) -- I caught that while I was doing this PR and fixed it. It should definitely be separate commit though, at least, so we can revert selectively if necessary.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
sunahsuh: Yeah, I think in this case even if I used `.get` it'd be best practice to catch the `NoSuchElementException` and throw an Exception with more context anyway

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ScalarAnalyzer.scala
fbertsch: Yup, I agree that the difference here isn't substantial enough to make the change.

mreid-moz:
🎉 🎉 🎉 🎉 🎉 🎉

PR #343, opened 2017-11-08
Add appleModelId to main_summary.

sapohl:


src/main/scala/com/mozilla/telemetry/utils/package.scala
```diff
@@ -11,6 +11,7 @@ package object utils{
   private val specialCases = Map(
     "submission_url" -> "submissionURL",
     "memory_mb" -> "memoryMB",
+    "apple_model_id" -> "appleModelId",
```
fbertsch: This isn't required for adding to `main_summary`.

sapohl:
Trying. ;-) First time using GitHub on this machine and I still need to set things up properly (2FA is not liking it right now). Will squash as soon as possible. Thanks for the review!

sapohl:
@fbertsch: I think that did it and is ready to be merged. Thanks again!

PR #344, opened 2017-11-08
Bug 1414093 - Ensure empty problem data arrays are treated as missing

thomcc:
iOS and Android are both guilty of this, and, among other issues, it leads to graphs like https://sql.telemetry.mozilla.org/queries/48563#131003, saying iOS and android fail every time.

It would be possible to handle elsewhere instead, but is simplest to handle here, since this field is already nullable and there's a well defined meaning for it being null.


thomcc:
> So this solves the problem where an empty array is not null?

Yes.

> Is this a patch that needs to be retroactively applied to the dataset?

Yes, I was going to request a backfill for this (as well as a number of other smaller issues we've fixed since the last backfill) after it was merged.

PR #347, opened 2017-11-13
Bug 1412875 - Add doctype as a parameter to MainSummaryView

acmiyaguchi:
I've tested this using the following command:

```
spark-submit \
--master yarn \
--deploy-mode client \
--class com.mozilla.telemetry.views.MainSummaryView \
telemetry-batch-view-1.1.jar \
--bucket telemetry-test-bucket \
--from 20171109 \
--to 20171109 -\
-channel nightly \
--doc-type first_shutdown
```


acmiyaguchi:
The `jobName` needs to be changed for the first-shutdown dataset, but changing the jobName in `main_summary` might not be the best solution.

acmiyaguchi:
I changed the output tablename to be ``${doctype}_summary`.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -90,11 +90,13 @@ object MainSummaryView {
     val channel = opt[String]("channel", descr = "Only process data from the given channel", required = false)
     val appVersion = opt[String]("version", descr = "Only process data from the given app version", required = false)
     val allHistograms = opt[Boolean]("all-histograms", descr = "Flag to use all histograms", required = false)
+    val docType = opt[String]("doc-type", descr = "DocType of pings conforming to main ping schema", required=false, default=Some("main"))
     verify()
   }
 
   def main(args: Array[String]) {
     val conf = new Conf(args) // parse command line arguments
+    val outputPrefix = s"${conf.docType()}_summary"
```
mreid-moz: nit: please move this down and use `filterDocType` so we don't need to call `conf.docType()` twice.

PR #348, opened 2017-11-13
Bug 1416915: Add browser.search.widget.inNavBar to userPrefsList

sunahsuh:
First one to r+ wins dzeber's undying gratitude

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
+++ b/src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
@@ -496,7 +496,7 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
         |}
       """.stripMargin)
     MainSummaryView.getOldUserPrefs(json1 \ "environment" \ "settings" \ "userPrefs") should be (None)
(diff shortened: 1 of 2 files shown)


src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -529,7 +529,7 @@ class MainSummaryViewTest extends FlatSpec with Matchers{
         |}
       """.stripMargin)
     MainSummaryView.getOldUserPrefs(json3 \ "environment" \ "settings" \ "userPrefs") should be (Some(Row(2, true)))
-    MainSummaryView.getUserPrefs(json3 \ "environment" \ "settings" \ "userPrefs", userPrefs) should be (Row(2, true, null))
+    MainSummaryView.getUserPrefs(json3 \ "environment" \ "settings" \ "userPrefs", userPrefs) should be (Row(2, true, null, null))
```
relud: this seems like it would be the appropriate place to test with `browser.search.widget.inNavBar` present in `userPrefs`. `extensions.legacy.enabled` isn't being tested anywhere either, so I see how this follows form for that, but I'm in favor of more testing.

relud: I changed my mind

PR #350, opened 2017-11-16
Limit bootstrapping to experiments with less than 25M pings for now

sunahsuh:
Turns bootstrapped confidence intervals back on for experiments with < 25M pings


src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -17,6 +17,7 @@ object ExperimentAnalysisView {
   def schemaVersion = "v1"
   // This gives us ~120MB partitions with the columns we have now. We should tune this as we add more columns.
   def rowsPerPartition = 25000
+  val bootstrapLimit = 25000000 // Empirically tested to take ~1 hour with this many pings
```
fbertsch: Should this be a configuration option?

sunahsuh: I think given you can override bootstrapping entirely with the bootstrapScalars option, this isn't necessary. Also, my hope is that this will be a temporary fix anyway and we can improve performance enough to not have to limit by ping count.

PR #351, opened 2017-11-16
Fix the docType filter

mreid-moz:
Subtle bug with the new docType filtering where the `case` variable was shadowing and causing all docTypes to match.


mreid-moz:
Please don't merge this yet - still testing.

mreid-moz:
Ok, my test looks good now.

PR #352, opened 2017-11-17
Fix AMODatabase not being shared with worker nodes

Dexterp37:
This is fixed by turning AMODatabase into a class and making the database value loaded lazily. This also simplifies the testing code: we use a different AMODatabase instance for each test.


src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -44,6 +44,7 @@ private case class ItemFactors(id: Int, features: Array[Float])
 object AddonRecommender {
   implicit val formats = Serialization.formats(NoTypeHints)
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getName)
+  private lazy val amoDB = new AMODatabase()
```
fbertsch: Could you retrieve the `addonDB: Map[String, AMOAddonInfo]` here instead? That is surely serializable, and wouldn't require loading the db on every worker node.

Dexterp37: Ah, do you mean I should:

- leave `AMODatabase` as a `final object`;
- make it return the `Map[String, AMOAddonInfo]`;
- store that `Map` in `AddonRecommender` and broadcast it with `sc.broadcast(..)`?

fbertsch: Unfortunately that wouldn't work, if we want to broadcast the `Map[String, AMOAddonInfo]` then it actually needs to be a Spark Dataset (Spark SQL doesn't support broadcast variables). This isn't a bad idea, since the broadcast join threshold is 10mb by default, and this is double that.

The alternative is just use the `Map[String, AMOAddonInfo]` directly (not through the `AMODatabase`, and not as a `lazy val`). This should make the function inside `flatMap` to be computed as a closure, and the map would then only be computed once on the master node and serialized to each worker.

Basically, it's up to you - if you're happy with recreating the AMODatabase on each worker node, and the performance hit is acceptable, you can keep it as-is.

Dexterp37:
Ah, I think I got what you mean now :) I've changed the PR to do that!

src/main/scala/com/mozilla/telemetry/ml/AMODatabase.scala
```diff
@@ -107,9 +106,9 @@ final object AMODatabase {
     * Initialize this object by downloading the AMO addons.
     * @param apiURI The AMO api URI to use to fetch the data.
     */
-  def init(apiURI: String = defaultAMORequestURI): Unit = {
+  def init(apiURI: String = defaultAMORequestURI): Map[String, AMOAddonInfo] = {
```
fbertsch: Consider changing the name of this function now to e.g. `getAddonMap`.

src/main/scala/com/mozilla/telemetry/ml/AMODatabase.scala
```diff
@@ -155,7 +154,7 @@ final object AMODatabase {
     * @param addonId The GUID for the desired addon.
     * @return True if the addon database contains the addon, False otherwise.
     */
-  def contains(addonId: String): Boolean = {
+  def contains(addonId: String, addonDB: Map[String, AMOAddonInfo]): Boolean = {
```
fbertsch: Can this function be removed entirely, in favor of calling `addonDB.contains` directly?

Dexterp37:
Yup :-) I spawned a cluster this morning and published to `telemetry-test-bucket`!

PR #353, opened 2017-11-21
Bug 1412875 - Add a FirstShutdownView

acmiyaguchi:
This creates a copy of MainSummaryView as a FirstShutdownView. This supports the creation of a first_shutdown dataset without affecting accidentally interrupting service to `main_summary_V4`. 

* defaults docType to `first_shutdown`
* renames misleading variable `payload` to `submission`
* uses `submission \ "payload"` instead of `parse(fields.get("payload.<subfield>", "{}") ...)` due to missing extraction in the DWL (see [comment #5](https://bugzilla.mozilla.org/show_bug.cgi?id=1412875#c5))
* adds simple tests for addonDetails (`submission.payload.addonDetails`), histograms (parent vs content), and experiments (DWL extracted `environment.experiments`)

This could also be implemented by adding `first_shutdown` to the ingestion filter, but it might be more worthwhile to refactor the moztelemetry Dataset API to create a logical view of the submission (see [bug 1419116](https://bugzilla.mozilla.org/show_bug.cgi?id=1419116)).


src/main/scala/com/mozilla/telemetry/views/FirstShutdownView.scala
```diff
@@ -0,0 +1,1120 @@
...
+import org.json4s.jackson.JsonMethods.parse
+import org.json4s.{DefaultFormats, JValue}
+import org.rogach.scallop._
+
+import scala.util.{Success, Try}
+
+object FirstShutdownView {
+
+  def schemaVersion: String = "v4"
+  def jobName: String = "main_summary"
```
mreid-moz: s/main/first_shutdown/

src/main/scala/com/mozilla/telemetry/views/FirstShutdownView.scala
```diff
@@ -0,0 +1,1120 @@
...
+
+  val histogramsWhitelist =
```
mreid-moz: Let's just reference this histograms list from `MainSummaryView` so they don't get out of sync.

src/main/scala/com/mozilla/telemetry/views/FirstShutdownView.scala
```diff
@@ -0,0 +1,1120 @@
...
+  // Prefs where we only record whether a pref has been set should
+  // use StringUserPref as we observe a value of "<user-set>".
+  //
+  // Supported pref data types are:
+  //   nsIPrefBranch.PREF_STRING -> StringUserPref
+  //   nsIPrefBranch.PREF_BOOL -> BooleanUserPref
+  //   nsIPrefBranch.PREF_INT -> IntegerUserPref
+  // See the `_getPrefData()` function in TelemetryEnvironment.jsm
+  // for reference: https://mzl.la/2zo7kyK
+  val userPrefsList =
```
mreid-moz: Similarly, let's reference this from `MainSummaryView` directly.

acmiyaguchi:
MainSummaryView would need to be a class, and re-assembling the message is difficult from the RichMessage interface. It would probably be easier if messageToRow were working on the raw json instead of heka frames.

@mreid-moz RFAL

src/test/scala/com/mozilla/telemetry/views/FirstShutdownViewTest.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  val defaultSchema = MainSummaryView.buildSchema(userPrefs, scalarDefs, histogramDefs)
+
+  val defaultMessageToRow = (m: Message) =>
+    FirstShutdownView.messageToRow(m, userPrefs, scalarDefs, histogramDefs)
+
+  // Apply the given schema to the given potentially-generic Row.
+  def applySchema(row: Row, schema: StructType): Row = new GenericRowWithSchema(row.toSeq.toArray, schema)
+
+  "MainSummary plugin counts" can "be summarized" in {
```
mreid-moz: nit: s/MainSummary/FirstShutdown/ in test descriptions throughout this file.

fbertsch:
@acmiyaguchi note that I made a similar extension for `error_aggregates`, and ended up keeping the original as an object and calling it's methods directly. I do think that would be easier than having an entirely new object where we may have to replicate changes to `main_summary`.

How long do we plan on keeping this dataset for?

fbertsch:
@acmiyaguchi that change was here: https://github.com/mozilla/telemetry-streaming/commit/736f52ca649efba0b4fec5ab0b54ce039fda0971

acmiyaguchi:
@fbertsch This commit boils down to the differences in the ping schemas due to the field extraction in the hindsight filter.  I'm using most of the methods from the MainSummaryView. Unfortunately, nothing that can be done about `messageToRow` without refactoring the method since its the method that jsonifies and maps the fields in the heka frame to the columns in the rdd. 

There are some follow-up bugs ([1419116](https://bugzilla.mozilla.org/show_bug.cgi?id=1419116), [1420981](https://bugzilla.mozilla.org/show_bug.cgi?id=1420981)) for maintaining a logical view of the ping and for removing this view. I want to enable first-shutdown for Firefox 58 or 59 if it seems like a reasonable thing and have all the related code here disappear.

acmiyaguchi:
@mreid-moz This adds a few more tests that should cover most of the differences between the pings.

PR #356, opened 2017-11-27
Fix null pointer exception from specific client

sunahsuh:
This fixes the ExperimentAnalysisView failures over the weekend -- basically, we have a ping or two where the inner map in a keyed histogram is null, for whatever reason, which was an unhandled issue. I've only been able to find two instances of this underlying issue in the main_summary in the column that was throwing this error (`histogram_parent_fx_urlbar_selected_result_index_by_type`) since 11/24 (when the job started failing with this error.) Since the incidence is small, I decided to just discard those pings.


sunahsuh:
Makes sense -- merging this as-is to fix this for tonight and I'll give more though to how to make this actually alert us if the incidence increases

PR #358, opened 2017-11-29
Bug 1421697 - Add experiments to Crash Summary

fbertsch:


fbertsch:
@sunahsuh I was going to remove them but figured let's keep it backwards compatible for now. This is a low-use dataset so I don't think it should cause issues. We can remove those fields once we're sure no one is using them.

PR #360, opened 2017-12-08
Experiment error aggregates

fbertsch:
Tested, this works with the updated experiment_error_aggregates.


fbertsch:
@sunahsuh note that on a 4-node cluster this ran in ~10 minutes, so it shouldn't affect the performance of this job at all.

PR #361, opened 2017-12-19
Limit histograms and scalars to a default list in Experiment Analysis…

sunahsuh:
… View

This should at least get us running. Ran a test on one experiment and verified the output had the expected metrics.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
@@ -21,17 +21,38 @@ object ExperimentAnalysisView {
 
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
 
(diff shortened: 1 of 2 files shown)


sunahsuh:
@mreid-moz Looks like Frank's on PTO today so if you have a chance to review this, t'would be most apprciated :)

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -21,17 +21,38 @@ object ExperimentAnalysisView {
 
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getSimpleName)
 
-  // scalar_ and histogram_ columns are automatically added
   private val usedColumns = List(
```
fbertsch: nit: consider renaming this to `metricWhitelist` or `columnWhitelist`, to parallel the `whitelist` usage in `MainSummaryView`.

PR #362, opened 2018-01-04
Bug 1404715 - Longitudinal Fixes

sunahsuh:
This PR does two things:
1. Removes the hack from Bug 1321424 which was a quick fix to prevent duplicate longitudinal runs, which we haven't run into since the Dockerflow migration. Instead the behavior was changed to delete any previously existing data before the run, which is closer to the overwrite logic we're using for main summary and other jobs recently
2. Sets the number of shuffle/output partitions to defaultParallelism (960 in the case of the 30 node cluster we're currently using), which will hopefully fix the errors for the last two runs, which are probably OOMs due to sorting in too-large partitions, and will potentially also speed up the job enough to finish under the timeout threshhold

Tested this on nightly pings for 20171230

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -17,6 +17,7 @@ import com.mozilla.telemetry.avro
 import com.mozilla.telemetry.heka.{Dataset, Message}
 import com.mozilla.telemetry.metrics._
 import com.mozilla.telemetry.utils._
+import com.mozilla.telemetry.utils.deletePrefix
 
 protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int = 1000) extends Iterator[List[Map[String, Any]]] {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -155,98 +156,97 @@ object LongitudinalView {
...
+              JString(startDate) <- (info \ "subsessionStartDate").toOption
+              JInt(counter) <- (info \ "profileSubsessionCounter").toOption
+            } yield ((clientId, startDate, counter.toInt),
+              fields + ("payload" -> compact(render(payload))) - "submission")
+        }
+        .repartitionAndSortWithinPartitions(new ClientIdPartitioner(numPartitions))
+        .map { case (key, value) => (key._1, value) }
 
-    /* One file per partition is generated at the end of the job. We want to have
+      /* One file per partition is generated at the end of the job. We want to have
```
acmiyaguchi: This comment is out of date now.

PR #363, opened 2018-01-08
Bug 1399625 - reduce main_summary parquet file size

relud:
to reduce individual file size

I added `document_id` to the columns we reparititon by, because if only increase the number for repartition then there's no effect, and I _think_ `document_id` is supposed to be unique per ping.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -189,8 +189,10 @@ object MainSummaryView {
         val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$currentDateString"
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
-        // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        // Repartition the dataframe by document_id before saving.
+        // This results in 4 buckets that are distinct from sample_id, so that
+        // there are 4 files per sample_id partition.
+        val partitioned = records.repartition(4, records.col("document_id"))
 
         // Then write to S3 using the given fields as path name partitions. Overwrites
         // existing data.


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -190,7 +190,7 @@ object MainSummaryView {
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
         // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        val partitioned = records.repartition(4, records.col("sample_id"), records.col("document_id"))
```
relud: on replacing `100` with `4`: it appears that we get up to that number of files _per partition_, based on hashing the columns provided into buckets to redistribute. so maybe we should drop `sample_id` in this line and just repartition by `document_id`, relying on line 197 to split by sample_id.

fbertsch:
@relud have you confirmed that this writes out more files? In my experience this will not actually increase the number of files. See my comment here for more info on how to do this: https://github.com/mozilla/telemetry-streaming/pull/69/files#r139155089

relud:
this is how I tested:
<pre>
Using Scala version 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112)
Type in expressions to have them evaluated.
Type :help for more information.

scala> import java.util.UUID.randomUUID
import java.util.UUID.randomUUID

scala> import scala.sys.process._
import scala.sys.process._

scala> val df = (1 to 200000).map { e => (e%100, randomUUID().toString) }.toList.toDF("id", "val")

scala> df.repartition(100, df.col("sample_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/08 22:13:23 WARN TaskSetManager: Stage 2 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
100
res0: Int = 0

scala> df.repartition(2, df.col("document_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/08 22:15:03 WARN TaskSetManager: Stage 4 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
200
res1: Int = 0

scala> df.repartition(4, df.col("document_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/08 22:17:27 WARN TaskSetManager: Stage 8 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala>  "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
400
res2: Int = 0

scala> val df = (1 to 200000).map { e => (e%200, randomUUID().toString) }.toList.toDF("id", "val")

scala> df.repartition(100, df.col("sample_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/08 22:18:20 WARN TaskSetManager: Stage 2 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
200
res3: Int = 0

scala> df.repartition(2, df.col("document_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/08 22:20:04 WARN TaskSetManager: Stage 4 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
400
res4: Int = 0
</pre>

I believe what happens with using `repartition` is that it makes buckets and fills them based on a hashed value of the columns provided. Then `partition` splits those buckets across directories, resulting in up to buckets*partitions number of files. But that's just the max, because if each bucket maps to a single partition, you get 1*partitions number of files (res0 and res3). Meanwhile if each bucket is independent of partition and present for every partition, you get the max number of files (res1, res2, and res4).

relud:
the second answer here is what i'm basing my understanding on: https://stackoverflow.com/questions/40416357/spark-sql-difference-between-df-repartition-and-dataframewriter-partitionby

relud:
When i leave out the repartitions columns completely, my test fails to distribute data into buckets that are distinct from sample_id:
<pre>
scala> df.repartition(2, df.col("document_id")).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
[Stage 0:>                                                          (0 + 0) / 4]18/01/10 18:30:31 WARN TaskSetManager: Stage 0 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.
[Stage 1:>                                                          (0 + 2) / 2]SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
200
res2: Int = 0

scala> df.repartition(2).write.partitionBy("sample_id").mode("overwrite").parquet("/tmp/test-output")
18/01/10 18:31:05 WARN TaskSetManager: Stage 2 contains a task of very large size (3961 KB). The maximum recommended task size is 100 KB.

scala> "find /tmp/test-output -type f -name *.parquet" #| "wc -l" !
warning: there was one feature warning; re-run with -feature for details
100
res4: Int = 0
</pre>

I think this is because using `repartition` without columns is just counting off messages into buckets in order, but `sample_id` is also generated in order, so they line up. I don't know if this would be an issue in `main_summary`, but I'm definitely concerned it could be.

relud:
I added a small comment to the code to document why it is repartitioning by `document_id`, if we leave that in.

PR #365, opened 2018-01-13
Restore behaviour prior to #363

mreid-moz:
I believe this is causing the recent failures of main_summary - repartitioning to only 4 partitions seems likely to be blowing up a few of the workers.


mreid-moz:
Thanks @relud! May turn out that this is not the issue, but it's an easy first thing to try.

PR #366, opened 2018-01-16
Bug 1399625 - repartition main_summary for more files

relud:
do not merge unless prod data test run succeeds

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -12,6 +12,7 @@ import com.mozilla.telemetry.utils.{Addon, Attribution, Events,
 import com.mozilla.telemetry.utils.{BooleanUserPref, IntegerUserPref, UserPref}
 import org.json4s.{JValue, DefaultFormats}
 import com.mozilla.telemetry.metrics._
+import java.util.zip.CRC32
 
 import scala.util.{Success, Try}
 
@@ -92,6 +93,7 @@ object MainSummaryView {
     val appVersion = opt[String]("version", descr = "Only process data from the given app version", required = false)
     val allHistograms = opt[Boolean]("all-histograms", descr = "Flag to use all histograms", required = false)
     val docType = opt[String]("doc-type", descr = "DocType of pings conforming to main ping schema", required=false, default=Some("main"))
+    val filesPerPartition = opt[Int]("files-per-partition", descr = "Number of files to create in every partition in S3", required = false, default=Some(4))
     verify()
   }
 
@@ -189,8 +191,18 @@ object MainSummaryView {
         val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$currentDateString"
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
-        // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        // Add temporary part_id column to control repartition
+        val filesPerPartition = conf.filesPerPartition()
+        val partitionId = (clientId: String, sampleId: Int) => {
+          val crc = new CRC32
+          crc.update(if (clientId == null) "".getBytes else clientId.getBytes)
+          sampleId * filesPerPartition + crc.getValue % filesPerPartition
+        }
+        sqlContext.udf.register("partid", partitionId)
+        val recordsWithPartId = records.selectExpr("*", "partid(client_id, sample_id) as part_id")
+
+        // Repartition the dataframe before saving
+        val partitioned = recordsWithPartId.repartition(recordsWithPartId.col("part_id")).drop("part_id")
 
         // Then write to S3 using the given fields as path name partitions. Overwrites
         // existing data.


relud:
test run: https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-1ZDJNSKZNDTAI

relud:
rake 2: https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-1PKRC0XB8Z4O4

relud:
take 3: https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-4QZMBH0ORRV3

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -189,8 +190,14 @@ object MainSummaryView {
         val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$currentDateString"
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
-        // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        // Add temporary part_id column to control repartition
+        val filesPerSampleId = 4
```
fbertsch: Please move this to a constant; even better, to a configuration parameter.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -189,8 +190,14 @@ object MainSummaryView {
...
+        val partitionId = (s: Int) => s * filesPerSampleId + Random.nextInt(filesPerSampleId)
```
fbertsch: I would prefer `partitionId` to be reproducible for a given `client_id`, so that the same clients are contained within the same files for each day.

relud: fixed, replaced `Random.nextInt()` with @mreid-moz 's `CRC32` modulus

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: fixed

relud:
take 4: https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-2LPZY5124NUBO

relud:
test run with client_id CRC32 for bucketing: [j-249UDFZDJKFEF](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-249UDFZDJKFEF)

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -106,6 +108,7 @@ object MainSummaryView {
       case Some(f) => fmt.parseDateTime(f)
       case _ => DateTime.now.minusDays(1)
     }
+    val filesPerPartition = conf.filesPerPartition()
```
relud: I keep getting `Job aborted due to stage failure: Task not serializable: java.io.NotSerializableException: com.mozilla.telemetry.views.MainSummaryView$Conf` which I think is due to this line, but I have no idea how to make it stop trying to serialize Conf

fbertsch: @relud can you confirm that by commenting out this line and replacing it with `val filesPerPartition = 4`?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -189,8 +192,17 @@ object MainSummaryView {
         val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$currentDateString"
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
-        // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        // Add temporary part_id column to control repartition
+        val partitionId = (clientId: String, sampleId: Int) => {
+          val crc = new CRC32
+          crc.update(if (clientId == null) "".getBytes else clientId.getBytes)
+          sampleId * filesPerPartition + crc.getValue % filesPerPartition
```
mreid-moz: I don't think this will map clientIds in a given partition back to the same sampleId... I believe you need something like:

`sampleId + ((crc.getValue % filesPerPartition) * 100)`

We should factor this function out and add test coverage for it. Since (I think...) we need the magic `100` anyways, it might be easier to leave out the `sampleId` argument and go back to just specifying `clientId` and `modulus` (which would let us test that we can reproduce the `sampleId` calculation as well).

relud: given a `sampleId` of `10`, `sampleId * filesPerPartition + crc.getValue % filesPerPartition` will yield values `40`, `41`, `42`, `43` where `sampleId + ((crc.getValue % filesPerPartition) * 100)` will yield values `10`, `110`, `210`, `310`.

`crc.getValue % filesPerPartition` is consistent for a given client id, and only produces integers in `[0,1,2,3]`. both methods will produce up to 4 `partitionId` values per `sampleId`. If that weren't true then [j-4QZMBH0ORRV3](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-4QZMBH0ORRV3) which succeeded, but used `Random.nextInt()` in place of `crc.getValue` would have more than 396 parquet files (`aws s3 ls 's3://telemetry-test-bucket/main_summary/v4/submission_date_s3=20180109/' --recursive | grep parquet | wc -l`).

that said, this 396 files business is v odd, i'm going to need to look into it.

mreid-moz: Right, it wasn't clear to me that everything with a `partitionId` of `40`, say, would map back to `sampleId`=`10`. Thinking about it in more detail I see that it's filling the partitionId space evenly and won't have any dupes with your approach either. Sorry for the noise :)

Still though, factoring this out and adding some tests would still be good.

relud: I found out why we get 396 files. from the [repartition(column) docs](https://spark.apache.org/docs/2.1.0/api/scala/index.html#org.apache.spark.sql.Dataset@repartition(partitionExprs:org.apache.spark.sql.Column*):org.apache.spark.sql.Dataset[T])

> Returns a new Dataset partitioned by the given partitioning expressions, using spark.sql.shuffle.partitions as number of partitions. The resulting Dataset is hash partitioned.

which means that it doesn't create a partition for each unique value of (column). it hashes (column) into a number of partitions equal to spark.sql.shuffle.partitions, and that hashing is getting 4 collisions.

if I instead do `repartition(sampleIdModulus*filesPerPartition, df.col("part_id"))`, I only get one collision:

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: I tested that and it didn't help, now i'm testing if it's the crc thing, because before crc & conf option this worked.

relud: when i put `val filesPerPartition = 4` here, it causes the exception, but when I put it right before the `val partitionId = (clientId: String, sampleId: Int) => {` line, it works. testing to see if the same applies to `val filesPerPartition = conf.filesPerPartition()`

relud: apparently i just needed to be putting this in the right spot. it's working now.

relud:
successful test run: [j-IB9A02LKIMLZ](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-IB9A02LKIMLZ)

fbertsch:
@relud, a few things to check on this successful run:
- That the data is the same
- Whether there were `null` sample_ids in that run. If not, you may need to deal with that in your code
- That the size of the files are relatively uniform (would be surprised if this *wasn't* the case)

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -189,8 +191,18 @@ object MainSummaryView {
         val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$currentDateString"
         val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
 
-        // Repartition the dataframe by sample_id before saving.
-        val partitioned = records.repartition(100, records.col("sample_id"))
+        // Add temporary part_id column to control repartition
+        val filesPerPartition = conf.filesPerPartition()
+        val partitionId = (clientId: String, sampleId: Int) => {
```
mreid-moz: Please move this function out of the `main` function as a top-level function in `MainSummaryView` and add some tests in `MainSummaryViewTest`.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -189,8 +191,18 @@ object MainSummaryView {
...
+          val crc = new CRC32
+          crc.update(if (clientId == null) "".getBytes else clientId.getBytes)
+          sampleId * filesPerPartition + crc.getValue % filesPerPartition
+        }
+        sqlContext.udf.register("partid", partitionId)
+        val recordsWithPartId = records.selectExpr("*", "partid(client_id, sample_id) as part_id")
+
+        // Repartition the dataframe before saving
+        val partitioned = recordsWithPartId.repartition(recordsWithPartId.col("part_id")).drop("part_id")
```
mreid-moz: Per [previous comment](https://github.com/mozilla/telemetry-batch-view/pull/366#discussion_r162136316), do we need to add the desired number of partitions as the first argument to `repartition` here to make sure we aren't limited by the default `spark.sql.shuffle.partitions`?

PR #367, opened 2018-01-22
Bug 1432282 - Add fields from sync-ping/payload/os to the sync event telemetry table

thomcc:
[Relevant bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1432282). We plan on adding event telemetry for additional platforms soon, and so we're getting this done in advance.

I don't know if I need to bump `SyncEventView.schemaVersion` to `"v2"` for this, please let me know either way.

Thanks.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/views/SyncEventViewTest.scala
+++ b/src/test/scala/com/mozilla/telemetry/views/SyncEventViewTest.scala
@@ -147,6 +147,47 @@ class SyncEventViewTest extends FlatSpec with Matchers{
       |}
     """.stripMargin)
 
+  def sync_payload_with_os = parse(
+    """
+      |{
+      |  "type": "sync",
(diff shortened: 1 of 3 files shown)


src/test/scala/com/mozilla/telemetry/views/SyncEventViewTest.scala
```diff
@@ -194,4 +239,40 @@ class SyncEventViewTest extends FlatSpec with Matchers{
...
+      val rdd = sc.parallelize(row)
+
+      val dataframe = spark.createDataFrame(rdd, SyncEventConverter.syncEventSchema)
+
+      // verify the contents.
+      dataframe.count() should be (1)
+      val checkRow = dataframe.first()
+
+      val payload = sync_payload \ "payload"
+      // not present in this ping (common, need to make sure this works)
```
sunahsuh: Looks like a copy-paste comment here

PR #369, opened 2018-01-31
Add MWU to experiment statistics

robhudson:


robhudson:
The tests will fail until https://github.com/mozilla/moztelemetry/pull/17 is merged.

src/main/scala/com/mozilla/telemetry/experiments/statistics/ComparativeStatistics.scala
```diff
@@ -1,6 +1,7 @@
 package com.mozilla.telemetry.experiments.statistics
 
 import com.mozilla.telemetry.experiments.analyzers.{HistogramPoint, MetricAnalysis, Statistic}
+import com.mozilla.telemetry.stats
```
sunahsuh: Proabably not in scope for this review, but I'd suggest changing the package name in moztelemetry to `com.mozilla.telemetry.statistics` for consistency (since this one is in statistics, and I'm fairly certain it's older :))

robhudson: Should I do that now before this lands?

sunahsuh: I suppose now would probably be a good time since we'll be doing a version bump soon

PR #370, opened 2018-02-06
Bug 1434991 - add bucketed() udf to GenericCountView

relud:
the buckets are specified like `--bucket-splits 0 5 10 100`

`bucketed(value)` returns the next split bigger than value, or the largest split + 1 for values bigger than all splits.

The test changes show how telemetry-airflow can use this to generate a `bucketed_uri_count` field.


src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala
```diff
@@ -98,6 +102,15 @@ object GenericCountView {
     }
   }
 
+  private def getBucket(splits: List[Int], value: Float): Int = {
+    for (i <- splits) {
+      if (value <= i) {
+        return i
+      }
+    }
+    return splits(splits.length-1)+1
```
mreid-moz: style nit - the last expression in a function is the return value, so you don't need `return` here. Also, you can simplify to `splits.last + 1`

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
```diff
@@ -36,6 +37,7 @@ object Submission{
     "os" -> List("Windows", "Darwin"),
     "os_version" -> List("1.0", "1.1"),
     "devtools_toolbox_opened_count" -> List(0, 42),
+    "scalar_parent_browser_engagement_total_uri_count" -> List(0,3,8,19,45,50,60,80,150,200,250,251,255),
```
mreid-moz: nit: please add spaces after commas in the list

relud:
oh cool, that is a way better way to do `bucketed()`. I had to use `int('0')` instead of `lit(0)` because lit doesn't seem to be available in the correct context.

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
```diff
@@ -153,6 +161,15 @@ class GenericCountTest extends FlatSpec with Matchers with BeforeAndAfterAll {
     distributionIdCount.length should be (4)
   }
 
+  "Only bucketed values" should "be in bucketed_uri_count" in {
+    val bucketedUriCountCount = aggregates
+      .groupBy("bucketed_uri_count")
+      .agg(countDistinct($"bucketed_uri_count"))
+      .collect()
+
+    bucketedUriCountCount.length should be (7)
```
fbertsch: Please verify that the output histogram values are correct.

src/main/scala/com/mozilla/telemetry/views/GenericCountView.scala
```diff
@@ -98,6 +98,12 @@ object GenericCountView {
     }
   }
 
+  private def getBucket(value: Float, splits: Seq[Int]): Int = {
```
fbertsch: you can actually probably remove the brackets around `splits.collectFirst` altogether, since it's just one expression.

relud: fixed

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
relud: fixed

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
```diff
@@ -148,6 +156,21 @@ class GenericCountTest extends FlatSpec with Matchers with BeforeAndAfterAll {
...
+  "Only bucketed values" should "be in bucketed_uri_count" in {
+    val bucketedUriCountCount = aggregates
+      .groupBy("bucketed_uri_count")
+      .agg(countDistinct($"bucketed_uri_count"))
+      .collect()
+
+    val expect = List(0, 5, 20, 50, 100, 250, 251)
+
+    bucketedUriCountCount.foreach { x =>
+      expect should contain (x(0))
```
fbertsch: Please check that the counts for each of these is correct, e.g. 20 should have a count of 2 (8 and 19).

relud:
@fbertsch could you take one more look at it, I had to rework the test to use distinct values of `scalar_parent_browser_engagement_total_uri_count`

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
```diff
@@ -93,18 +97,23 @@ class GenericCountTest extends FlatSpec with Matchers with BeforeAndAfterAll {
     "app_version" ::
     "e10s_enabled" ::
     "os" ::
-    "os_version" :: Nil
+    "os_version" ::
+    "scalar_parent_browser_engagement_total_uri_count" :: Nil
+
+  private val splits = "array(int('0'), int('5'), int('20'), int('50'), int('100'), int('250'))"
```
mreid-moz: Can you use `array(0, 5, 20, 50, 100, 250)`?

relud: that doesn't work. it tries to use the numbers as column references.

relud:
will do

relud:
moved to `udfs.scala`

PR #371, opened 2018-02-07
Bug 1431458 - add util that validates test main_summary output against prod

relud:
initially went with matching count and set of `document_id`s

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
@@ -1,0 +1,117 @@
+package com.mozilla.telemetry.utils
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.rogach.scallop._
+
+object DatasetComparator {
+  def jobName: String = "dataset_comparator"
+
+  // Configuration for command line arguments
+  class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Date to compare", required = true)
+    val dateField = opt[String]("date-field", descr = "Name of date partition field", required = false, default = Some("submission_date_s3"))
+    val dataPath = opt[String]("data-path", descr = "Path in buckets to data", required = false, default = Some("main_summary/v4"))
+    val failWithoutMatch = opt[Boolean]("fail-without-match", descr = "If true, exit 1 if data does not match", required = false, default = Some(true))
+    val originalBucket = opt[String]("original-bucket", descr = "Source bucket for original data", required = false, default = Some("telemetry-parquet"))
+    val originalViewname = opt[String]("original-viewname", descr = "View to pull original data from. Cannot be used with --original-bucket", required = false)
+    val select = opt[String]("select", descr = "Comma separated list of columns that will be directly compared with DataFrame.except", required = false, default = Some("document_id"))
+    val testBucket = opt[String]("test-bucket", descr = "Source bucket for test data", required = false, default = Some("telemetry-test-bucket"))
+    val testViewname = opt[String]("test-viewname", descr = "View to pull test data from. Cannot be used with --test-bucket", required = false)
+    val where = opt[String]("where", descr = "Filter statement for the data to compare; e.g. \"sample_id = '0'\"", required = false)
+
+    conflicts(originalViewname, List(originalBucket))
+    conflicts(testViewname, List(testBucket))
+    verify()
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args) // parse command line arguments
+    // Set up Spark
+    val spark = getOrCreateSparkSession(jobName)
+    val valid = validate(spark, conf)
+    spark.stop()
+    if (!valid && conf.failWithoutMatch()) {
+      System.exit(1)
+    }
+  }
+
+  def getDataFrame(spark: SparkSession, view: ScallopOption[String], bucket: ScallopOption[String], dataPath: String, dateField: String, date: String, where: ScallopOption[String]): DataFrame = {
+    val df = view.get match {
+      case Some(t) => spark.sql(s"SELECT * FROM $t").where(s"$dateField = '$date'")
+      case _ => spark.read.parquet(s"s3://${bucket()}/$dataPath/$dateField=$date")
+    }
+    where.get match {
+      case Some(clause) => df.where(clause)
+      case None => df
+    }
+  }
+
+  def validate(spark: SparkSession, conf: Conf): Boolean = {
+    val dataPath = conf.dataPath()
+    val date = conf.date()
+    val dateField = conf.dateField()
+    val select = conf.select()
+
+    println("=======================================================================================")
+    println(s"BEGINNING JOB $jobName FOR $date")
+
+    val original = getDataFrame(spark, conf.originalViewname, conf.originalBucket, dataPath, dateField, date, conf.where)
+    val test = getDataFrame(spark, conf.testViewname, conf.testBucket, dataPath, dateField, date, conf.where)
+
+    var same = true
+
+    // compare row counts
+    val originalCount = original.count
+    val testCount = test.count
+    if (originalCount != testCount) {
+      println(s"Wrong number of rows. Expected $originalCount. Got $testCount")
+      same = false
+    }
+
+    // compare null count in each column
+    original.columns.foreach{ col =>
+      if (test.columns contains col) {
+        val originalNulls = original.filter(s"$col is null").count
+        val testNulls = test.filter(s"$col is null").count
+        if (originalNulls != testNulls) {
+          println(s"Wrong number of null values in '$col'. Expected $originalNulls. Got $testNulls")
+          same = false
+        }
+      } else {
+        println(s"Missing column '$col'.")
+        same = false
+      }
+    }
+
+    // check for extra columns
+    test.columns.foreach { col =>
+      if (!(original.columns contains col)) {
+        println(s"Extra column '$col'.")
+        same = false
+      }
+    }
+
+    // compare comparable columns
+    val originalColumns = original.select(select)
+    val testColumns = test.select(select)
+
+    val missing = originalColumns.except(testColumns).count
+    if (missing > 0) {
+      println(s"$missing missing value(s)")
+      same = false
+    }
+
+    val extra = testColumns.except(originalColumns).count
+    if (extra > 0) {
+      println(s"$extra extra value(s)")
+      same = false
+    }
+
+    val status = if (same) "SUCCESS" else "FAILURE"
+
+    println(s"$status ON JOB $jobName FOR $date")
+    println("=======================================================================================")
+
+    same
+  }
+}
--- a/src/test/scala/com/mozilla/telemetry/utils/DatasetComparatorTest.scala
+++ b/src/test/scala/com/mozilla/telemetry/utils/DatasetComparatorTest.scala
@@ -1,0 +1,77 @@
+package com.mozilla.telemetry.utils
+
+import org.apache.spark.sql.SparkSession
+import org.scalatest.FlatSpec
+
+case class MSRow(
+  document_id:Int,
+  submission_date_s3:String,
+  client_id:Option[String] = None
+)
+
+class DatasetComparatorTest extends FlatSpec {
+  private val spark = SparkSession
+    .builder()
+    .appName("DatasetComparatorTest")
+    .master("local[*]")
+    .getOrCreate()
+
+  import spark.implicits._
+
+  private val date = "test"
+
+  def getConf(original: String, test: String, where: String = "", select: String = "document_id", dateField: String = "submission_date_s3"): DatasetComparator.Conf = {
+    val args =
+      "--date" :: date ::
+      "--date-field" :: dateField ::
+      "--original-viewname" :: original ::
+      "--select" :: select ::
+      "--test-viewname" :: test ::
+      {if (where == "") Nil else "--where" :: where :: Nil}
+    new DatasetComparator.Conf(args.toArray)
+  }
+
+  (0 to 10).map{e => MSRow(e,date)}.toDF.createOrReplaceTempView("zeroToTen")
+  (0 to 9).map{e => MSRow(e,date)}.toDF.createOrReplaceTempView("zeroToNine")
+  (0 to 11).map{e => MSRow(e,date)}.toDF.createOrReplaceTempView("zeroToEleven")
+  (1 to 11).map{e => MSRow(e,date)}.toDF.createOrReplaceTempView("oneToEleven")
+  (0 to 10).map{e => (0,date)}.toDF("document_id", "submission_date_s3").createOrReplaceTempView("ZeroNulls")
+  (0 to 10).map{e => if (e % 2 == 0) MSRow(0,date,Some("c")) else MSRow(0,date)}.toDF.createOrReplaceTempView("FiveNulls")
+  (1 to 11).map{e => if (e % 2 == 0) MSRow(0,date,Some("c")) else MSRow(0,date)}.toDF.createOrReplaceTempView("SixNulls")
+
+  "Simple frame" must "equal itself" in {
+    assert(DatasetComparator.validate(spark, getConf("zeroToTen", "zeroToTen")))
+  }
+
+  it must "not equal one less row" in {
+    assert(!DatasetComparator.validate(spark, getConf("zeroToTen", "zeroToNine")))
+  }
+
+  it must "equal one less row with where clause" in {
+    assert(DatasetComparator.validate(spark, getConf("zeroToTen", "zeroToNine", "document_id < 10")))
+  }
+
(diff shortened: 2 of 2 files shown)


src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,94 @@
...
+    // compare row counts
+    val originalCount = original.count
+    val testCount = test.selectExpr("COUNT(*)").collect().last(0)
+    if (original.count != test.count) {
+      println(s"Date $currentDateString contains wrong number of rows. Expected $originalCount. Got $testCount")
+      same = false
+    }
+
+    // compare comparable columns
+    val originalColumns = original.selectExpr("document_id")
```
acmiyaguchi: This can also just be `original.select`

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  def equal(currentDateString: String, original: DataFrame, test: DataFrame): Boolean = {
+    var same = true
+
+    // compare row counts
+    val originalCount = original.count
+    val testCount = test.selectExpr("COUNT(*)").collect().last(0)
+    if (original.count != test.count) {
```
acmiyaguchi: re-use the above variables

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    }
+  }
+
+  def equal(currentDateString: String, original: DataFrame, test: DataFrame): Boolean = {
+    var same = true
+
+    // compare row counts
+    val originalCount = original.count
+    val testCount = test.selectExpr("COUNT(*)").collect().last(0)
```
acmiyaguchi: Why not `test.count`?

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,94 @@
...
+
+      if (!equal(currentDateString, original, test)) {
+        fail = true
+      }
+
+      println(s"COMPLETED JOB $jobName FOR $currentDateString")
+      println("=======================================================================================")
+    }
+
+    spark.sparkContext.stop()
```
acmiyaguchi: `spark.stop` also deals with all the relevant clean up.

relud:
apologies, i forgot to re-read the code after the tests were working, and I submitted the code in an incomplete state. I have corrected the review and added a check for nulls, and a relevant test.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.joda.time.{DateTime, format}
+import org.rogach.scallop._
+
+object MainSummaryValidate {
+  def schemaVersion: String = "v4"
+  def jobName: String = "main_summary_validate"
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Submission date to compare", required = false)
```
acmiyaguchi: This should be a required field

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args) // parse command line arguments
+    val datePattern = "yyyyMMdd"
+    val fmt = format.DateTimeFormat.forPattern(datePattern)
+    val date = conf.date.get match {
+      case Some(f) => fmt.parseDateTime(f).toString(datePattern)
+      case _ => DateTime.now.minusDays(1).toString(datePattern)
+    }
```
acmiyaguchi: This setup can be removed

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,93 @@
...
+    if (!equal(date, original, test)) {
+      fail = true
+    }
+
+    println(s"COMPLETED JOB $jobName FOR $date")
+    println("=======================================================================================")
+
+    spark.stop()
+
+    if (fail) {
```
acmiyaguchi: This variable could be replaced with the actual condition

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
relud: fixed

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
relud: fixed

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
relud: fixed

relud:
@acmiyaguchi i added checks for null count in each column and mismatched columns.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,97 @@
...
+    val testBucket = conf.testBucket()
+    val originalBucket = conf.originalBucket()
+
+    println("=======================================================================================")
+    println(s"BEGINNING JOB $jobName FOR $date")
+
+    val original = spark.read.parquet(s"s3://$originalBucket/main_summary/$schemaVersion/submission_date_s3=$date")
+    val test = spark.read.parquet(s"s3://$testBucket/main_summary/$schemaVersion/submission_date_s3=$date")
+
+    if (!equal(date, original, test)) {
```
acmiyaguchi: You could probably remove the date from this function, since all of the text will be wrapped in a box.

relud: fixed

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,97 @@
...
+import org.joda.time.{DateTime, format}
+import org.rogach.scallop._
+
+object MainSummaryValidate {
+  def schemaVersion: String = "v4"
+  def jobName: String = "main_summary_validate"
+
+  // Configuration for command line arguments
+  private class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Submission date to compare", required = true)
```
acmiyaguchi: One last comment, this job would also be useful as a single sample_id or channel. I think passing in --channel=nightly is a common way of reducing the size of the main_summary job for validation.

relud: added option `--where` and modified tests for improved coverage and ability to test `--where`

src/test/scala/com/mozilla/telemetry/utils/MainSummaryValidateTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  import spark.implicits._
+
+  private val date = "test"
+
+  def getConf(original: String, test: String, where: String = ""): MainSummaryValidate.Conf = {
+    val args =
+      "--date" :: date ::
+      "--original-viewname" :: original ::
+      "--test-viewname" :: test ::
+      {if (where == "") Nil else "--where" :: where :: Nil}
```
relud: there is probably a more idiomatic way to express this, but i don't know it (yet)

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+    val date = conf.date()
+    val where = conf.where()
+
+    println("=======================================================================================")
+    println(s"BEGINNING JOB $jobName FOR $date")
+
+    val original = conf.originalViewname.get match {
+      case Some(t) => spark.sql(s"SELECT * FROM $t").where(s"submission_date_s3 = '$date' and $where")
+      case _ => spark.read.parquet(s"s3://${conf.originalBucket()}/main_summary/$schemaVersion/submission_date_s3=$date").where(where)
+    }
```
acmiyaguchi: I would actually do something like the following: 

```scala
val originalDF = where.get match {
  case Some(clause) => original.where(clause)
  case None => original
}
```

where original is the dataframe for main_summary. 

The argument is a Option[String], which is composed of two cases Some(x) and None. This means you don't need the default value of Some("TRUE") in the argument definition.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    val status = if (same) "SUCCESS" else "FAILURE"
+
+    println(s"$status ON JOB $jobName FOR $date")
+    println("=======================================================================================")
+
+    same
+  }
+
+  def main(args: Array[String]) {
```
acmiyaguchi: Can this be pushed up toward the top of the file?

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+  def schemaVersion: String = "v4"
+  def jobName: String = "main_summary_validate"
+
+  // Configuration for command line arguments
+  class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Submission date to compare", required = true)
+    val originalBucket = opt[String]("original-bucket", descr = "Source bucket for original main_summary data", required = false, default = Some("telemetry-parquet"))
+    val originalViewname = opt[String]("original-viewname", descr = "View to pull origianl data from. Cannot be used with --original-bucket", required = false)
+    val testBucket = opt[String]("test-bucket", descr = "Source bucket for test main_summary data", required = false, default = Some("telemetry-test-bucket"))
+    val testViewname = opt[String]("test-viewname", descr = "View to pull test data from. Cannot be used with --test-bucket", required = false)
```
acmiyaguchi: The test data is probably not going to be registered with hive, unfortunately.

relud: right, this is basically just for unit tests, like GenericCountViewTest does.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.rogach.scallop._
+
+object MainSummaryValidate {
+  def schemaVersion: String = "v4"
+  def jobName: String = "main_summary_validate"
+
+  // Configuration for command line arguments
+  class Conf(args: Array[String]) extends ScallopConf(args) {
+    val date = opt[String]("date", descr = "Submission date to compare", required = true)
+    val originalBucket = opt[String]("original-bucket", descr = "Source bucket for original main_summary data", required = false, default = Some("telemetry-parquet"))
```
fbertsch: You can use `requiredOne` and `conflicts` to prevent the use of both this and `originalViewname`. See https://github.com/mozilla/telemetry-streaming/blob/master/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala#L83

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+  def validate(spark: SparkSession, conf: Conf): Boolean = {
+    val date = conf.date()
+    val where = conf.where()
+
+    println("=======================================================================================")
+    println(s"BEGINNING JOB $jobName FOR $date")
+
+    val original = conf.originalViewname.get match {
+      case Some(t) => spark.sql(s"SELECT * FROM $t").where(s"submission_date_s3 = '$date' and $where")
+      case _ => spark.read.parquet(s"s3://${conf.originalBucket()}/main_summary/$schemaVersion/submission_date_s3=$date").where(where)
```
fbertsch: This is main_summary specific, this could be a config

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+    }
+    val test = conf.testViewname.get match {
```
fbertsch: This logic and the logic above (for `var original`) are the same, consider refactoring them together.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+          same = false
+        }
+      } else {
+        println(s"Missing column '$col'.")
+        same = false
+      }
+    }
+
+    // check for extra columns
+    test.columns.foreach { col =>
```
fbertsch: Should this test for missing columns as well?

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+      if (!(original.columns contains col)) {
+        println(s"Extra column '$col'.")
+        same = false
+      }
+    }
+
+    // compare comparable columns
+    val originalColumns = original.select("document_id")
```
fbertsch: If these were configurable, it would generalize the ability to directly compare columns. It would then allow multiple, and let you know which ones were different and how different they were.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
```diff
@@ -0,0 +1,103 @@
...
+    same
+  }
+
+  def main(args: Array[String]) {
+    val conf = new Conf(args) // parse command line arguments
+    // Set up Spark
+    val spark = getOrCreateSparkSession(jobName)
+    val valid = validate(spark, conf)
+    spark.stop()
+    if (!valid) {
```
fbertsch: It might be good to have a config option of whether this will fail on failed validation. It might be a use case to have the job succeed, just report differences.

src/main/scala/com/mozilla/telemetry/utils/MainSummaryValidate.scala
relud: that happens a couple lines above during the null comparison check

relud:
Addressed reviews

src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
```diff
@@ -0,0 +1,119 @@
+package com.mozilla.telemetry.utils
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.rogach.scallop._
+
+object DatasetComparator {
+  def jobName: String = "main_summary_validate"
```
acmiyaguchi: The name of the job has changed.

src/test/scala/com/mozilla/telemetry/utils/DatasetComparatorTest.scala
```diff
@@ -0,0 +1,77 @@
...
+case class MSRow(
+  document_id:Int,
+  submission_date_s3:String,
+  client_id:Option[String] = None
+)
+
+class DatasetComparatorTest extends FlatSpec {
+  private val spark = SparkSession
+    .builder()
+    .appName("ClientCountViewTest")
```
acmiyaguchi: Name should reflect the job.

src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+
+    // compare comparable columns
+    val originalColumns = original.select(select)
+    val testColumns = test.select(select)
+
+    val missing = originalColumns.except(testColumns).count
+    if (missing > 0) {
+      println(s"$missing missing row(s)")
```
fbertsch: Nit: change this verbiage, the rows aren't missing, just different.

relud: I think that missing is the correct verbiage here. This shows distinct row values found in `original` and not `test`. The next check does distinct row values found in `test` and not `original`, and refers to them as extra rows. If only one of them is hit, then it definitely indicates missing or extra values.

That said, I think perhaps `values` would be more correct here, since it's a set operation, and only counts the number of distinct different values.

PR #372, opened 2018-02-09
Bug 1431198 - Add environment/system/sec fields to MainSummary

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1431198)

Adds the `environment/system/sec` fields to the top level schema. This has also been tested with data in telemetry-test-bucket for 20180201.


acmiyaguchi:
Grr, broken tests are from the FirstShutdownView being out of sync with MainSummary. It's no longer needed, so I'll tack the change along with this one.

PR #373, opened 2018-02-12
Add write-mode to generic longitudinal

fbertsch:


fbertsch:
It is intentional, thanks for the r+!

PR #374, opened 2018-02-12
WIP: Switch from bootstrap to jackknife statistics

sunahsuh:
Tests will likely fail, but this is the current block jackknife implementation for verification on the mathy parts --

Each metric row now contains a block_id, which is assigned via CRC32 + modulo much like sample_id. Aggregation of metrics is now broken into two stages:
- Each block_id is aggregated by itself, and this result is persisted in memory, and the true aggregate is created from this
- The drop-k (where k = (total pings / number of blocks) jackknife pseudosamples are created by
  - creating a copy of each block for every number from 0 to <num_blocks> (the pseudosample number) _except the block's original id_
  - the block copies are then aggregated and grouped by each psuedosample number
  - There's an explanatory comment on why it was done this way (I also ran into issues with task serialization when doing this a more straighforward way.)

The `DescriptiveStatistics` class's `confidenceInterval` method now, if a dataset of psuedosamples is passed in, will run the summary stat on each psuedosample, get the mean and standard deviation of the result from the pseudosamples, and calculate the 99% confidence interval of the normally-distributed means, which gives us the 99% confidence interval for the point estimate itself.

At one point, I output the pseudovalues and they look fairly normally distributed to me:
https://screenshots.firefox.com/xOvzdhBXow5FiAJ7/localhost

So, barring some issue with the arithmetic (which I hope is the issue, because that's the simplest issue to fix), the other possibilities for the narrow CI bands I can think of are:
- We're using too many blocks (there's some literature on choosing optimum block size that i haven't delved into too deeply yet)
- We're very susceptible to outlier clients (very possible! Saptarshi had floated the idea of normalizing clients so they're all equally weighted, which would alleviate this.)


src/main/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatistics.scala
```diff
@@ -1,27 +1,38 @@
...
-        // currently using the super simple percentile version instead of the BCA algorithm
-        val halfAlpha = Alpha / 2
-        val lowIdx = scala.math.floor(stats.length * halfAlpha).toInt
-        val highIdx = scala.math.ceil(stats.length * (1 - halfAlpha)).toInt
-        (stats(lowIdx), stats(highIdx))
+    jackknifed.map { ds =>
+      import ds.sparkSession.implicits._
+      val (jkMean, jkStdDev, n) = ds
+        .map(m => run(m.histogram))
+        .agg(mean("value"), stddev_samp("value"), count("*"))
```
harterrt: I think we're missing a factor here. Take a look at my math [here](https://www.sharelatex.com/read/rvqkrjmdnttq).

sunahsuh: So, I think the formula in the wikipedia article is mislabeled -- what's labeled as the variance is really the Standard Error ^ 2. Check out reference 4 on that article, p 3 where it shows the formula for jackknife SE: http://people.bu.edu/aimcinto/jackknife.pdf

And that's equivalent to (stddev_samp/sqrt(n)), which is what we're multiplying that Z-score by a few lines later

harterrt: That definition for jackknife SE matches the definition I referenced in my comment. How are you getting to `stddev_samp/sqrt(n)`? My math (linked above) suggests this is should be `sqrt(n-1)*stddev_samp`.

sunahsuh: Oh, okay, looks like SE is labeled STDEV in the latex notes you linked to?

In any case, yes, you're right -- I'm dividing when I should be multiplying! But, stddev_samp uses 1/(n-1) as the scaling factor while stddev_pop uses 1/n so I think it'd be sqrt(n-1) * stddev_pop

See: https://www.sharelatex.com/read/krqpyzzxkjbj
(I'm a little mad that I just wrote LaTeX again :P But I'm so glad this is the problem and not something harder to fix.)

harterrt: Great catches. My labelling was sloppy - probably because I was angry at writing latex. 

Glad this is easy to fix too. I'm interested to see the false positive rate once this is revised. Hopefully we see ~1 false positive in 20.

sunahsuh:
No longer a WIP, ready for actual review! @harterrt to check my corrections, and @fbertsch since he's most familiar with the code. Ran this on a real workload from this week and it finished in under 3 hours!

Note: it's clear from the hoop-jumping I had to do for the CI tests that this should get a refactor but I need to rewrite a bunch of this to make it incremental anyway so I'll spend some time rethinking the whole structure then.

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
```diff
@@ -1,34 +1,111 @@
...
-    val actual = Mean(histo).asStatistic
-    actual.value should equal(4.0)
+    val (asInts, histo, jackknifeAggs) = fixtures(spark)
+    val actual = Mean(histo, Some(jackknifeAggs)).asStatistic
+    val (m, stdDev, n) = asInts.agg(mean("value"), stddev_samp("value"), count("*")).first match {
+      case Row(m: Double, s: Double, n: Long) => (m, s, n)
+      case _ => fail("This should never happen")
+    }
+
+    actual.value should equal(m)
```
harterrt: It may be useful to verify the actual value is within some distance from the population mean (i.e. 10). That could introduce flakiness, but would verify (e.g.) the mean is actually a number. What do you think? Same goes for the confidence intervals.

src/main/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatistics.scala
```diff
@@ -1,27 +1,39 @@
...
+      val (jkMean, jkStdDev, n) = ds
+        .map(m => run(m.histogram))
+        .agg(mean("value"), stddev_pop("value"), count("*"))
+        .first() match {
+        case Row(m: Double, s: Double, n: Long) => (m, s, n)
+        case _ => throw new Exception("Unexpected jackknife statistics results")
+      }
+
+      // 99% confidence interval is the standard z-score formula assuming approx normal distribution for pseudostats
+      val z99 = 2.576 * (jkStdDev * scala.math.sqrt(n.toDouble - 1.0))
```
harterrt: IIRC, the pseudostat stdDev needs to be scaled before it can apply to the actual data. Where are you doing that scaling?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -173,9 +167,9 @@ object ExperimentAnalysisView {
                            errorAggregates: DataFrame, conf: Conf,
                            experimentMetrics: List[String] = List()): List[MetricAnalysis] = {
     val pingCount = experimentsSummary.count()
-    val persisted = repartitionAndPersist(experimentsSummary, pingCount, experiment, experimentMetrics)
-    val bootstrapHistograms = conf.bootstrapHistograms()
-    val bootstrapScalars = shouldBootstrapScalars(pingCount, conf)
+    // Maybe TODO: tune number of blocks to sample size
```
fbertsch: Please move this to a bug with a bit more context.

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
```diff
@@ -1,34 +1,111 @@
...
+
+    // compute the standard 99% CI for the mean for a normal distribution
+    val expectedInterval = 2.576 * (stdDev / scala.math.sqrt(n.toDouble))
```
fbertsch: nit: magic numbers

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
```diff
@@ -1,34 +1,111 @@
...
-    val actual_median = Percentile(histo, 0.5, "Median").asStatistic
-    actual_median.value should equal(3.0)
+    val (asInts, histo, jackknifeAggs) = fixtures(spark)
+    val sortedIntArray = asInts.sort("value").collect()
+    val actualMedian = Percentile(histo, 0.5, "Median", Some(jackknifeAggs)).asStatistic
+    // Change this if we tune NumSamples to an odd number
+    val expected = sortedIntArray((NumSamples.toDouble * 0.5).toInt)
+    actualMedian.value should equal(expected)
+
+    // Using non-parametric interval calc of n(p) +- z * sqrt(n * p * (1 - p)) (where p is the percentile)
```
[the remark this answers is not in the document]
sunahsuh: Sure -- the most accessible citation is here: https://www-users.york.ac.uk/~mb55/intro/cicent.htm
Same formula [here](http://people.stat.sfu.ca/~cschwarz/Stat-650/Notes/PDF/ChapterPercentiles.pdf) on p 13, under "Non-Parametric Intervals"

Both of them source the formula from, I'm pretty sure, Conover's _Practical Nonparametric Statistics_

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
```diff
@@ -1,34 +1,111 @@
...
+    val ordinalInterval = 2.576 * scala.math.sqrt(NumSamples.toDouble * 0.5 * 0.5)
+    val expectedLower = sortedIntArray(scala.math.floor(NumSamples.toDouble * 0.5 - ordinalInterval).toInt)
+    val expectedUpper = sortedIntArray(scala.math.ceil(NumSamples.toDouble * 0.5 + ordinalInterval).toInt)
+
+    actualMedian.confidence_low.get should equal (expectedLower.toDouble +- 1) // +- 1 since these are discrete
```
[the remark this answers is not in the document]
sunahsuh: These tests are deterministic since we're feeding in a seed to the fixture, so we know this particular rendition will always work. And, fwiw, a limitation of the nonparametric formula that I'm using for the expected values is that it tends to produce wider confidence intervals.

src/main/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatistics.scala
sunahsuh: Re: that discussion, the conclusion was that we need to use stddev_pop (instead of stddev_samp), and I made that change on line 27 above

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
sunahsuh: Sure -- the first part definitely makes sense but now that I'm looking at this again, I'm not sure how you'd get to the expected confidence intervals on the mean/median estimators from the expected distribution.

sunahsuh: On second thought I'm going to punt on this since as an additional test it's kinda just testing the scala random library.

src/main/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatistics.scala
```diff
@@ -1,27 +1,39 @@
...
 import scala.annotation.tailrec
 import scala.collection.Map
 
 
 sealed abstract class DescriptiveStatistic(actual: Map[Long, HistogramPoint],
-                                           resampled: Option[List[Map[Long, HistogramPoint]]] = None) {
-  val Alpha = 0.05
+                                           jackknifed: Option[Dataset[MetricAnalysis]] = None)
+  extends java.io.Serializable {
+  val Alpha = 0.01
```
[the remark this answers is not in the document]
sunahsuh: the 6th parameter is "Confidence Level" -- p-value (parameter 7) defaults to None

sunahsuh:
RFAL -- I also just filed https://bugzilla.mozilla.org/show_bug.cgi?id=1454115 re: looking more at the percentile calculations and making sure they're consistent (and, potentially, changing the number of jackknife blocks based on the reading especially when sample sizes are small.)

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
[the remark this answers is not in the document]
harterrt: Just added Conover to my wish list. LGTM

src/test/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatisticsTest.scala
harterrt: :+1:

src/main/scala/com/mozilla/telemetry/experiments/statistics/DescriptiveStatistics.scala
harterrt: Awesome. I was referring to the scaling that you do on line 34, but everything looks right to me.

PR #376, opened 2018-02-13
Use provided jars during assembly

acmiyaguchi:
This uses provided jars for spark and hadoop instead of assembling them in the fat jar.

Any suggestions on other things that could be cut out safely?


acmiyaguchi:
It looks like the hacky modifications made to the hadoop jars are needed for the tests to pass. I'm also not what EMR does with older versions of spark now, it'll probably just run using the provided version.

PR #377, opened 2018-02-13
Use consistent interface for unwrapping nullable JValues

acmiyaguchi:
I've added some methods for extracting nullable types from JValues. I liked the concise notation of `JValue.extractOpt[T].orNull` as opposed to the 2 line match statement. 

I ended up putting these methods into an extension class that makes it easy to add and modify methods. These are apparently treated as static method calls once it's been compiled/jit'ed, so they should be fairly fast. [1] On the other-hand, I'm not entirely sure how fast the json4s `extract` methods are compared to pattern matching since it's a generic method. [2]

I might try out the different cases (original, implicit class + pattern matching, implicit class + extract) and see if it makes a significance difference on runtime.

[1] http://docs.scala-lang.org/sips/completed/value-classes.html
[2] https://blog.scalac.io/2017/05/25/scala-specialization.html

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -412,42 +412,49 @@ object MainSummaryView {
   // to a map containing just the fields we're interested in.
   def messageToRow(message: Message, scalarDefinitions: List[(String, ScalarDefinition)], histogramDefinitions: List[(String, HistogramDefinition)], userPrefs: List[UserPref] = userPrefsList): Option[Row] = {
     try {
-      val fields = message.fieldsAsMap
+      val doc = message.toJValue match {
+        case Some(doc) => doc
+        case None => return None
+      }
+
+      val environment = doc \ "environment"
+      val payload = doc \ "payload"
+      val meta = doc \ "meta"
 
       // Don't compute the expensive stuff until we need it. We may skip a record
       // due to missing required fields.
-      lazy val addons = parse(fields.getOrElse("environment.addons", "{}").asInstanceOf[String])
-      lazy val addonDetails = parse(fields.getOrElse("payload.addonDetails", "{}").asInstanceOf[String])
-      lazy val payload = parse(message.payload.getOrElse(fields.getOrElse("submission", "{}")).asInstanceOf[String])
-      lazy val application = payload \ "application"
-      lazy val build = parse(fields.getOrElse("environment.build", "{}").asInstanceOf[String])
-      lazy val profile = parse(fields.getOrElse("environment.profile", "{}").asInstanceOf[String])
-      lazy val partner = parse(fields.getOrElse("environment.partner", "{}").asInstanceOf[String])
-      lazy val settings = parse(fields.getOrElse("environment.settings", "{}").asInstanceOf[String])
-      lazy val system = parse(fields.getOrElse("environment.system", "{}").asInstanceOf[String])
-      lazy val info = parse(fields.getOrElse("payload.info", "{}").asInstanceOf[String])
-      lazy val simpleMeasures = parse(fields.getOrElse("payload.simpleMeasurements", "{}").asInstanceOf[String])
+      lazy val addons = environment \ "addons"
+      lazy val addonDetails = payload \ "addonDetails"
+      lazy val application = doc \ "application"
+      lazy val build = environment \ "build"
+      lazy val experiments = environment \ "experiments"
+      lazy val profile = environment \ "profile"
+      lazy val partner = environment \ "partner"
+      lazy val settings = environment \ "settings"
+      lazy val system = environment \ "system"
+      lazy val info = payload \ "info"
+      lazy val simpleMeasures = payload \ "simpleMeasurements"
 
       lazy val histograms = MainPing.ProcessTypes.map{
         _ match {
-          case "parent" => "parent" -> parse(fields.getOrElse("payload.histograms", "{}").asInstanceOf[String])
-          case p => p -> payload \ "payload" \ "processes" \ p \ "histograms"
+          case "parent" => "parent" -> payload \ "histograms"
+          case p => p -> payload \ "processes" \ p \ "histograms"
         }
       }.toMap
 
       lazy val keyedHistograms = MainPing.ProcessTypes.map{
         _ match {
-          case "parent" => "parent" -> parse(fields.getOrElse("payload.keyedHistograms", "{}").asInstanceOf[String])
-          case p => p -> payload \ "payload" \ "processes" \ p \ "keyedHistograms"
+          case "parent" => "parent" -> payload \ "keyedHistograms"
+          case p => p -> payload \ "processes" \ p \ "keyedHistograms"
         }
       }.toMap
 
       lazy val scalars = MainPing.ProcessTypes.map{
-        p => p -> payload \ "payload" \ "processes" \ p \ "scalars"
+        p => p -> payload \ "processes" \ p \ "scalars"
       }.toMap
 
       lazy val keyedScalars = MainPing.ProcessTypes.map{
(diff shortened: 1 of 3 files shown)


acmiyaguchi:
I'm piggybacking some changes on this PR which includes some other improvements to the MainSummaryView code.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -381,64 +381,86 @@ object MainSummaryView {
...
+  // A note on json4s JValue extraction:
+  //
+  // JValue provides an `.extract[T]` interface for deconstructing values to
+  // native types. However, fields in the schema are mostly nullable. One
+  // way to extract the value is to use a boxed type e.g. `scala.Long` or
+  // `java.lang.Long`. These are wrappers around the primitive types and
+  // support nullable values. This can increase the the latency of the
+  // program by introducing extra overhead. [1].
+  //
+  // Here, we choose to extract the value into an Option and unwrap into a primitive Any.
```
mreid-moz: Can we get a similar advantage by using `.extractOpt[T]` directly in the construction of the `row` variable using a `Seq` of `Option`s, then mapping to `Seq[Any]` at the end?

Something like
```scala
val row = Row.fromSeq(Seq(
  ...
  (meta \ "clientId").extractOpt[String],
  (meta \ "sampleId").extractOpt[Long],
  ...
).map(_.orNull))

**Comment by acmiyaguchi** [comment]
I had 3 jars above, with an additional commit that changed the number of read partitions to be 4*sc.defaultParallelism. I ran this on a 3 cluster setup.

original - before this PR
before-jvalue - 8179e75
final - c192854 

Here's a very simple script for testing these changes:


```
#!/bin/bash

run_ms () {
        echo "Running ${name}"
        time spark-submit --master yarn \
        --deploy-mode client \
        --class com.mozilla.telemetry.views.MainSummaryView  \
        ${name}.jar \
        --from 20180101 --to 20180101 \
        --bucket net-mozaws-prod-us-west-2-pipeline-analysis/amiyaguchi/ms_test_2/${name} \
        --channel nightly
}

export name="original"
run_ms

export name="before-jvalue"
run_ms

export name="final"
run_ms
```

original - 10 minutes
before-jvalue - 10 minutes
final - 4.7 minutes

Finally, I validated this data using the DatasetComparator (18 minutes).
```
spark-submit --master yarn --class com.mozilla.telemetry.utils.DatasetComparator final.jar --date 20180101 --data-path main_summary/v4 --original-bucket net-mozaws-prod-us-west-2-pipeline-analysis/amiyaguchi/ms_test_2/original --test-bucket net-mozaws-prod-us-west-2-pipeline-analysis/amiyaguchi/ms_test_2/final

Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
=======================================================================================
BEGINNING JOB dataset_comparator FOR 20180101
SUCCESS ON JOB dataset_comparator FOR 20180101
=======================================================================================
```

File: src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala (hunk 1 above)
Comment (acmiyaguchi) [reply src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala]: Yes, the extension class is syntactic sugar but also an easy place to define extraction. I think they both look as noisy either way.

Comment (acmiyaguchi) [reply src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala]: Actually, this isn't as easy as it looks. There's some inconsistency in terms of what the helper functions return (Any vs Option). What could be done is something like

```
val row = Row(
  ...
  (meta \ "clientId").extractOpt[String].orNull,
  (meta \ "sampleId").extractOpt[Long].orNull,
  ...
)
```

**Comment by acmiyaguchi** [comment]
Bumping for review.

File: src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala [hunk 2]
```diff
@@ -383,64 +383,86 @@ object MainSummaryView {
...
-      lazy val profile = parse(fields.getOrElse("environment.profile", "{}").asInstanceOf[String])
-      lazy val partner = parse(fields.getOrElse("environment.partner", "{}").asInstanceOf[String])
-      lazy val settings = parse(fields.getOrElse("environment.settings", "{}").asInstanceOf[String])
-      lazy val system = parse(fields.getOrElse("environment.system", "{}").asInstanceOf[String])
-      lazy val info = parse(fields.getOrElse("payload.info", "{}").asInstanceOf[String])
-      lazy val simpleMeasures = parse(fields.getOrElse("payload.simpleMeasurements", "{}").asInstanceOf[String])
-
-      lazy val histograms = MainPing.ProcessTypes.map{
+      val doc = message.toJValue match {
+        case Some(doc) => doc
```
mreid-moz: Please change the `doc` var here to something else to avoid shadowing the outer variable on the previous line

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
fbertsch: You could still check and see if they are an `Option`:
```
).map(
  _ match {
    case e: Option[Any] => e.orNull
    case o => o
  }
)
```

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -383,64 +383,86 @@ object MainSummaryView {
...
+        case None => return None
```
fbertsch: Please remove early returns. We've found this leads to all kinds of problems from Longitudinal.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
mreid-moz: This still seems worth doing - we get strong types and can [drop a use of `implicit`](https://github.com/databricks/scala-style-guide#implicits) in exchange.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
acmiyaguchi: There are also some early returns for documentId and submissionDate that I'll try to handle as well.

acmiyaguchi:
@fbertsch I've done some modification to avoid the early returns, which includes modifying the messageToRow method to accept JValue instead of message and lazily evaluating the row. There are also a couple of new unit tests for this behavior.

acmiyaguchi:
Wrapping the entire function in a big if-statement performs better than the lazy evaluator by roughly 10% (original 10, lazy evaluator 6.5, if-block 5, in minutes). This isn't an entirely fair comparison because I only made the variables in `Some(Row.merge(row, userPrefsRow, scalarRow, histogramRow))` lazy.

Without early returns, this is probably the best way to handle this case.

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -1701,4 +1703,17 @@ class MainSummaryViewTest extends FlatSpec with Matchers {
     )
     compare(message, expected)
   }
+
+  "Malformed message" should "be ignored" in {
+
+    var message = RichMessage("1234", Map("documentId" -> "foo", "submissionDate" -> "1234"), None)
+    defaultMessageToRow(message).isDefined should be (true)
```
fbertsch: nit: You removed spaces after `be` elsewhere.

acmiyaguchi: Chucked it through the intellij scala reformatting tool for consistency.

acmiyaguchi:
Thanks for the review!

PR #385, opened 2018-03-01
Add three important performance histograms.

mreid-moz:
These have been requested as "core metrics" on the performance front - two of the three are pre-release only, but still worth adding.


mreid-moz:
oops, wrong branch

PR #387, opened 2018-03-01
Remove memory_heap_committed_unused which is, ironically, largely unused

sunahsuh:
None of the nightly pings from 2/27 had data in this histogram, so I think it's safe to say it doesn't belong in main_summary


acmiyaguchi:
It also looks unused in the codebase, so this looks good to me.

PR #388, opened 2018-03-01
Bug 1442326 - Do not filter Android sync event pings anymore

eoger:
Firefox Android sends pings with `appName: "Fennec"`.
Let's remove that filter on the SyncEventView and bring it in line with the other views.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
@@ -72,8 +72,6 @@ object SyncEventView {
           case "4" | "5" => true
         }.where("docType") {
           case "sync" => true
-        }.where("appName") {
-          case "Firefox" => true
         }.where("submissionDate") {
           case date if date == currentDate.toString("yyyyMMdd") => true
         }.records(conf.limit.get)


src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
```diff
@@ -72,8 +72,6 @@ object SyncEventView {
           case "4" | "5" => true
         }.where("docType") {
           case "sync" => true
-        }.where("appName") {
-          case "Firefox" => true
```
acmiyaguchi: We may not want to include pings that get put into the `OTHER` bucket. I suggest using something like `case "Firefox" | "Fennec"`.

thomcc: Worth noting deleting it brings it in line with `Sync{,Flat}View`, see https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/SyncFlatView.scala#L70-L79, https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L69-L78

acmiyaguchi:
It seems like this is consistent with the rest of the sync code, so it looks good to me.

PR #389, opened 2018-03-05
Bug 1431777 - Move Clients Daily out of python_mozetl

relud:
rewrite from python to scala to take advantage of user defined
aggregate functions, which are a java-only construct.

adds experiments column to replace experiments_daily dataset
changes from activity_date to submission_date

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
@@ -163,6 +163,128 @@ class CollectList(inputStruct: StructType, orderCols: List[String], maxLength: O
   }
 }
 
+class AggSearchCounts(validSources: List[String], prefix: String = "search_count_") extends UserDefinedAggregateFunction {
+  /**
+    * Aggregates the search_counts column in main_summary by taking the sum of
+    * valid search counts by source, and creating a struct that maps source to
+    * count with an additional field "all" that is the sum of the others. The
+    * output column should be exploded to avoid issues with schema evolution in
+    * athena. A search count is valid if it has count >= 0 and source is listed
+    * in validSources. prefix is prepended to field names within the struct to
+    * make it safe for star expansion. The input column must have the schema
+    * List[Struct[engine: String, source: String, count: Long] ].
+    **/
+
+  private val numSources = validSources.length
+
+  override def inputSchema: StructType = StructType(
+    Array(
+      StructField(
+        "value",
+        // This is the actual input type
+        ArrayType(
+          StructType(
+            Array(
+              StructField("engine", StringType),
+              StructField("source", StringType),
+              StructField("count", LongType)
+            )
+          )
+        )
+      )
+    )
+  )
+
+  override def bufferSchema: StructType = StructType(
+    List(StructField(prefix + "all", LongType)) ++
+      validSources.map{c => StructField(prefix + c, LongType)}
+  )
+
+  override def dataType: DataType = bufferSchema
+
+  // sum doesn't care about ordering
+  override def deterministic: Boolean = true
+
+  // zeroes out the buffer when the aggregation key changes
+  override def initialize(buffer: MutableAggregationBuffer): Unit = {
+    (0 to numSources).foreach { index =>
+      buffer(index) = 0L
+    }
+  }
+
+  // adds valid sources from Row to the buffer
+  override def update(buffer: MutableAggregationBuffer, input: Row): Unit = {
+    if (!input.isNullAt(0)) {
+      input.getSeq(0).foreach { search_struct: Row =>
+        val source: String = search_struct.getString(1)
+        val count: Long = search_struct.getLong(2)
+        if (count > 0 && validSources.contains(source)) {
+          buffer(0) = buffer.getLong(0) + count
+          // index is offset by one because the first field in the buffer is "all"
+          val index = validSources.indexOf(source) + 1
+          buffer(index) = buffer.getLong(index) + count
+        }
+      }
+    }
+  }
+
+  // merging two buffers results in the sum of values for each key
+  override def merge(buffer1: MutableAggregationBuffer, buffer2: Row): Unit = {
+    (0 to numSources).foreach { index =>
+      buffer1(index) = buffer1.getLong(index) + buffer2.getLong(index)
+    }
+  }
+
+  // the buffer is the output value
+  override def evaluate(buffer: Row): Any = buffer
+}
+
+class AggMapFirst() extends UserDefinedAggregateFunction {
+  /**
+    * Aggregates a Map[String,String] column by choosing the first non-null
+    * value for each key
+    **/
+
+  override def inputSchema: StructType = StructType(
+    Array(
+      StructField(
+        "value",
+        // This is the actual input type
+        MapType(StringType, StringType)
+      )
+    )
+  )
+
+  override def bufferSchema: StructType = StructType(
+    List(StructField("map", MapType(StringType, StringType)))
+  )
+
+  override def dataType: DataType = MapType(StringType, StringType)
+
+  override def deterministic: Boolean = false
+
+  // empty buffer when the aggregation key changes
+  override def initialize(buffer: MutableAggregationBuffer): Unit = {
+    buffer(0) = Map()
+  }
+
+  // add only keys with non-null values from input that are missing in buffer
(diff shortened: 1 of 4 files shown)


src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+    val date = conf.date()
+    val maxRecordsPerFile = conf.maxRecordsPerFile()
+
+    val inputPath = s"s3://${conf.inputBucket()}/main_summary/" +
+      s"${MainSummaryView.schemaVersion}/submission_date_s3=$date"
+    val outputPath = s"s3://${conf.outputBucket()}/clients_daily/" +
+      s"$schemaVersion/submission_date_s3=$date"
+
+    val spark = getOrCreateSparkSession(jobName)
+    // don't write _SUCCESS file so we don't break Spark partition discovery
```
sunahsuh: This is a holdover in some of our older jobs but the _SUCCESS bug has been fixed in Spark for a while now

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+      .write
+      .mode("overwrite")
+      .option("maxRecordsPerFile", maxRecordsPerFile)
+      .parquet(outputPath)
+  }
+
+  def extractDayAggregates(df: DataFrame): DataFrame = {
+    val aggregates = df
+      .groupBy("client_id")
+      .agg(fieldAggregators(0), fieldAggregators.drop(1):_*)
```
sunahsuh: Slightly more idiomatic option is:
`.agg(fieldAggregators.head, fieldAggregators.tail:_*)`

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
+
+  def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
+
+  def fieldAggregators = List(
+    aggSum("aborts_content"),
+    aggSum("aborts_gmplugin"),
+    aggSum("aborts_plugin"),
+    // active_addons
```
sunahsuh: Are these comments carried over from the original job? I'm a bit confused about what they denote, especially since it appears they're not almost consistent, and sometimes seem to refer to something missing

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -163,6 +163,129 @@ class CollectList(inputStruct: StructType, orderCols: List[String], maxLength: O
...
+    )
+  )
+
+  override def bufferSchema: StructType = StructType(
+    List(StructField("map", MapType(StringType, StringType)))
+  )
+
+  override def dataType: DataType = MapType(StringType, StringType)
+
+  // sum doesn't care about ordering
```
sunahsuh: A bit of copy-pasta here :) Also, I think this particular aggregation is actually not deterministic, if we assume a user can switch experiment branches within a day

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+    aggSum("devtools_toolbox_opened_count"),
+    aggFirst("distribution_id"),
+    // userprefs/dom_ipc_process_count
+    aggFirst("e10s_enabled"),
+    aggFirst("env_build_arch"),
+    aggFirst("env_build_id"),
+    aggFirst("env_build_version"),
+    // events
+    // experiments
+    new AggMapFirst()(expr("experiments")).alias("experiments"),
```
sunahsuh: I'd suggest defining the udf as a member function like the other aggregation functions so the usage looks similar

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+    aggMean("first_paint"),
+    // first(
+    //     expr("userprefs.extensions_allow_non_mpc_extensions"
+    //        ).alias("extensions_allow_non_mpc_extensions")
+    // ),
```
sunahsuh: Can we delete this?

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -0,0 +1,235 @@
...
+    aggSum("scalar_parent_storage_sync_api_usage_extensions_using"),
+    // scalar_parent_storage_sync_api_usage_items_stored
+    // scalar_parent_storage_sync_api_usage_storage_consumed
+    aggFirst("scalar_parent_telemetry_os_shutting_down"),
+    aggSum("scalar_parent_webrtc_nicer_stun_retransmits"),
+    aggSum("scalar_parent_webrtc_nicer_turn_401s"),
+    aggSum("scalar_parent_webrtc_nicer_turn_403s"),
+    aggSum("scalar_parent_webrtc_nicer_turn_438s"),
+    aggFirst("search_cohort"),
+    new AggSearchCounts(List(
```
sunahsuh: Same here re: defining the udf as a member function, and I'd consider moving the list of valid search count keys into its own value

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
sunahsuh: Also, I *think* you should be able to use the column name string instead of wrapping it in `expr` -- but if not, `col` is probably the better function to use

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTest.scala
```diff
@@ -0,0 +1,28 @@
+package com.mozilla.telemetry.views
+
+import org.apache.spark.sql.SparkSession
+import org.scalatest.{FlatSpec, Matchers}
+import java.math.BigDecimal
+
+
+
+class ClientsDailyViewTest extends FlatSpec with Matchers {
+  private val spark = SparkSession
```
sunahsuh: You can use `getOrCreateSparkSession` here too!

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTestPayloads.scala
```diff
@@ -0,0 +1,497 @@
...
+    telemetry_enabled: Option[Boolean] = None,
+    timezone_offset: Option[Int] = None,
+    total_time: Option[Int] = None,
+    vendor: Option[String] = None,
+    web_notification_shown: Option[Int] = None,
+    windows_build_number: Option[Long] = None,
+    windows_ubr: Option[Long] = None
+  )
+
+  def getRowAggFirst(sValue: Option[String], bValue: Option[Boolean], iValue: Option[Int]) = MainSummaryRow(
```
sunahsuh: A salute for writing exhaustive tests! I'm not sure I'd have had the patience 😁

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
relud: deleted the extraneous comments copied over from python_mozetl

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
relud: fixed

src/main/scala/com/mozilla/telemetry/utils/udfs.scala
relud: fixed

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
relud: fixed

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTest.scala
relud: i can't figure out how to use `getOrCreateSparkSession` with `.config("spark.ui.showConsoleProgress", "false")`

PR #391, opened 2018-03-07
Bug 1395364 - Use standard utility function for creating SparkSession

akkomar:


src/test/scala/com/mozilla/telemetry/utils/EventsTest.scala
```diff
@@ -63,10 +63,7 @@ class EventsTest extends FlatSpec with Matchers{
     val sparkConf = new SparkConf().setAppName("MainSummary")
     sparkConf.setMaster(sparkConf.get("spark.master", "local[1]"))
     val sc = new SparkContext(sparkConf)
```
sunahsuh: Looks like there's some related cleanup to do here -- can you just remove the sparkConf/sc instatiation here and change the line with the FlatSpec to use `spark.sparkContext` instead?

PR #397, opened 2018-03-13
Add search composition experiment to excluded experiments

sunahsuh:
Waiting for confirmation from Dave Zeber that it's safe to merge this in, but this experiment is causing the downstream experiment analysis job to fail (we're up to 2B+ cumulative pings in this experiment) and the story seems to be that it's safe to ignore it since analysis has concluded on this (pending Zeber's approval)


dzeber:
Yes, this is fine to exclude.

PR #398, opened 2018-03-16
Allow Addon scalar definitions in MainSummary

fbertsch:
- Creates a new Scalars.yaml for Addon probes
- Incorporate DefaultProcess and AllowedProcesses for pings

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/metrics/Scalars.scala
+++ b/src/main/scala/com/mozilla/telemetry/metrics/Scalars.scala
@@ -6,25 +6,67 @@ import org.yaml.snakeyaml.Yaml
 import collection.JavaConversions._
 import scala.io.Source
 import com.mozilla.telemetry.utils.MainPing
+import org.apache.spark.sql.types._
 
-abstract class ScalarDefinition extends MetricDefinition
+import java.io.InputStream
+
+
+trait ScalarType {
+  val name: String
+  val keyed: Boolean
+  val dataType: DataType
+
+  def getName: String = keyed match {
+    case true => List("keyed", name).mkString("_")
+    case false => name
+  }
+
+  def getParquetType = keyed match {
+    case true => MapType(StringType, dataType, valueContainsNull = true)
+    case false => dataType
+  }
+}
+
+case class BooleanScalarType(keyed: Boolean = false) extends ScalarType {
+  val name = "boolean"
+  val dataType = BooleanType
+}
+
+case class UintScalarType(keyed: Boolean = false) extends ScalarType {
+  val name = "uint"
+  val dataType = IntegerType
+}
+
+case class StringScalarType(keyed: Boolean = false) extends ScalarType {
+  val name = "string"
+  val dataType = StringType
+}
+
+
+abstract class ScalarDefinition extends MetricDefinition {
+  val scalarType: ScalarType
+}
 
 case class UintScalar(keyed: Boolean, originalName: String, process: Option[String] = None)
   extends ScalarDefinition {
   val isCategoricalMetric = false
+  val scalarType = UintScalarType(keyed)
 }
 
 case class BooleanScalar(keyed: Boolean, originalName: String, process: Option[String] = None)
   extends ScalarDefinition {
   val isCategoricalMetric = true
+  val scalarType = BooleanScalarType(keyed)
 }
 
 case class StringScalar(keyed: Boolean, originalName: String, process: Option[String] = None)
   extends ScalarDefinition {
   val isCategoricalMetric = true
+  val scalarType = StringScalarType(keyed)
 }
 
 class ScalarsClass extends MetricsClass {
+  val DynamicScalarDefinitionsFile = "/addon/Scalars.yaml"
   val ScalarColumnNamePrefix = "scalar"
 
   // mock[io.Source] wasn't working with scalamock, so now we'll just use the function
@@ -35,7 +77,13 @@ class ScalarsClass extends MetricsClass {
     // want them to get into the Parquet column names, so replace them with
     // underscores ('_'). Additionally, to prevent name clashing, we prefix
     // all scalars.
-    ScalarColumnNamePrefix + '_' + (processName + '_' + scalarName).replace('.', '_')
+    //
+    // Dynamic scalars do not have the scalars_$PROCESS_ prefix,
+    // since they are already separated.
+    (processName match {
+      case MainPing.DynamicProcess => scalarName
+      case p => ScalarColumnNamePrefix + '_' + (p + '_' + scalarName)
+    }).replace(".", "_")
   }
 
   def definitions(includeOptin: Boolean = false, nameJoiner: (String, String) => String = getParquetFriendlyScalarName): Map[String, ScalarDefinition] = {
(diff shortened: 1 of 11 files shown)


sunahsuh:
Tests failed

fbertsch:
@sunahsuh fixed

fbertsch:
@sunahsuh were there any more thoughts from the meeting about this change?

acmiyaguchi:
@fbertsch I think that storing the dynamic scalars in a map-type would be a good idea. Since they are dynamic, we don't have the master `histograms.json` file to reference to when creating separate columns.

A  `scalars.yaml` file increases overhead on our part, since addon developers (or us) will have to manually add new fields. 

I did some hacking on a main_summary-like dataset ([gist](https://gist.github.com/acmiyaguchi/5182749fc28fb57631d4abc970d75381)) that handled transforming the sparse histogram values into a dense column representation by sampling existing data. For example, a histogram with 3 buckets would be transformed into 3 columns `HIST_1`, `HIST_2`, `HIST_3`. 

If we stored dynamic scalars as a map, I could imagine the following workflow:

daily:
1. Explode `payload_dynamic_scalars` into (name, measure) and count instances of each name
2. Generate a master list of (dynamic_scalar, channel, first_seen, last_seen)

We would make able to track the dates that a scalar appears and a rough idea of how often it's being used.

sunahsuh:
@acmiyaguchi that reminds me, one thing we didn't get to cover is that the yaml file is a temporary workaround and there's a plan to extend the probe scraper to include dynamic scalars in other locations. For now, using map columns wouldn't negate the need for a registry here since we'd still need to know the scalar's datatype.

To recap my thoughts on this, I think limiting schema changes and keeping the table from getting (potentially much) wider will make our lives easier. main_summary has already edged into uncharted territory in pushing the number of columns past what's generally supported by our data tooling and I think that should give us some serious pause, because it means we're probably doing something weird. I'd like us to not make that worse.

fbertsch:
There have been good points raised. I'm going to raise a final few.

## Performance

Under the hood, Parquet (and Dremel, so BigQuery as well) stores Map types as two columns: a key and a value. So if you want a specific key, then instead of a efficient columnar access to that key, you first need to read _all_ of the blocks that may contain the key (which, unless we put some efficient data structure in the parquet metadata to prune these blocks {e.g. bloom filter} will be all of them), then find the associated values in the other column (which, again, will be reading every block).

Now imagine that we have enough addon scalars that it would significantly affect the number of columns we have. _If_ that is the case, then it is more than enough to make this operation _very_ slow! *This is, in essence, eliminating any of the benefits of having a column-store at all*.

The alternative, top-level columns, will be efficient access to only the data you're interested in. There is great compression of missing values as well.

## Ease of Querying and Discoverability

The next question arises: how are these queried, and discovered? There are two options here for storage:

- Store every key on every record, and store the value as `NULL` if missing. Terrible for performance, easier for querying, best for discoverability, since any ping will show the addon scalars available.
- Store only the existing keys for the record. Best for performance, find for querying, terrible for discoverability.

Top-level fields suffer none of these issues, an would be immediately available in the schema browser (which, side note, should also be improved for this impending-large table).

## Predictable Types

While having per-datatype maps would indeed be simple to query, I don't think that storing our metadata inside of our data is a good practice. If we're having trouble understanding what types things are, we should improve the schema explorer (or make a new tool), and possibly improve the documentation. Maybe, for example, the auto-complete could tell you what type the column you're writing is.

## Final Notes

I do agree with @sunahsuh that we are reaching uncharted territories, which is exciting! However, I am not of the opinion that using a `MAP` type for these new columns is the appropriate solution. If we are indeed exceeding the limits of our tooling, forcing slow queries and large, difficult-to-understand columns is probably not the solution we want. Instead I propose we split the tables when we reach this point.

We could, for example, split by projects. Certain teams are more likely to use certain probes: engagement scalars are often used in conjunction with the simple measures (for obvious reasons). The Fennec probes are often used together, and not often used with e.g. `BUSY_TAB_ABANDONED`, which is a Desktop-only probe.

fbertsch:
We had a chat about this over IRC. One big point that we had not considered yet was historical data access. When a short-lived addon creates a new scalar, reports it for a period of time, and then goes offline, we want that data to be available. With the current approach, the data will stop being available once we remove the scalar definition (which is something that should happen when they are no longer being reported).

We can allow historical data access in two ways:
1. In the up-and-coming Data Access API, we can allow selection of dates from which the schema will be loaded. Historical parquet files will still have the data, so if we set `mergeSchema=true` then they will be available.
2. The databases can have separate tables per-day, each with their own schema; the primary `main_summary` table would then be a view over all of them. To access historical data that is no longer in the up-to-date schema, one would query those tables directly.

One problem might be date pruning within BigQuery. We would still want the ability to easily filter on the `submission_date` and not have to read all historical files, which may be problematic given it is a view.

For the time being, there is not much downside in just including all scalars forever, since parquet will just not store anything for those columns if they are not reported. Once they are more heavily used we can incorporate some of these ideas.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -470,33 +491,33 @@ object MainSummaryView {
...
-      val keyedHistograms = MainPing.ProcessTypes.map {
+      val keyedHistograms = MainPing.AllowedProcessTypes.map {
         _ match {
           case "parent" => "parent" -> payload \ "keyedHistograms"
           case p => p -> payload \ "processes" \ p \ "keyedHistograms"
         }
       }.toMap
 
-      val scalars = MainPing.ProcessTypes.map {
+      val scalars = MainPing.AllowedProcessTypes.map {
```
sunahsuh: I feel like merging the dynamic scalars with the rest of the processes is creating a lot of extra work (here and in the ScalarsClass) -- it doesn't look like there's a case at the moment where we really need those merged. I think if you creating a separate ScalarsClass object for DynamicScalars and keep the dynamic scalar data separate, you'd end up with a cleaner implementation.

sunahsuh:
Ah, I just noticed that there appears to be a real test failure too :(

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
fbertsch: I'm not sure I'm following. Creating another class specifically for Dynamic scalars seems like it would be very confusing, since `ScalarDefinition` already has a `process` field, and `ScalarType` is independent of `process`. IMO this should remain as-is, minimizing confusion from redundant classes.

Other suggestions are good though! You are correct, this implementation was a bit rushed. changes incoming.

fbertsch:
@sunahsuh working on the test failures (seems not to be reading the resource file for addon scalars), but you can review the code now.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
sunahsuh: I'm okay leaving it as is, but to clarify my position, I think conflating in-tree with dynamic scalars is creating a false and leaky abstraction -- it's not only the process that's different here, but where the list is coming from, and how we want to handle it in code in the places that we use the list. The very fact that we need to special-case dynamic scalars out of the list both in the changes to main summary and to longitudinal speaks to this.

src/main/scala/com/mozilla/telemetry/metrics/Scalars.scala
```diff
@@ -10,45 +10,59 @@ import org.apache.spark.sql.types._
 
 import java.io.InputStream
 
-case class ScalarType(typeStr: String, keyed: Boolean = false, joiner: String = "_") {
+
+trait ScalarType {
+  val name: String
```
sunahsuh: Is this field used anywhere now?

src/main/scala/com/mozilla/telemetry/metrics/Scalars.scala
```diff
@@ -10,45 +10,59 @@ import org.apache.spark.sql.types._
...
     case false => dataType
   }
+}
 
-  def matches(scalar: ScalarDefinition): Boolean = scalar match {
-    case s: BooleanScalar => s.keyed == keyed && dataType == BooleanType
-    case s: StringScalar => s.keyed == keyed && dataType == StringType
-    case s: UintScalar => s.keyed == keyed && dataType == IntegerType
-  }
+case class BooleanScalarType(keyed: Boolean = false) extends ScalarType {
```
sunahsuh: In my earlier comment about the structure of this class I mentioned six objects, and while I don't feel strongly enough to say that this PR should be changed to accommodate that, the reason I suggested that implementation was that it seems like the three types and their three keyed variants are better suited to singleton objects instead of case classes that are instantiated separately for each scalar definition.

PR #399, opened 2018-03-19
Fix handling of null search counts in clients daily

relud:
backfill revealed a missing test case that was breaking the AggSearchCounts udaf

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/udfs.scala
@@ -217,12 +217,14 @@ class AggSearchCounts(validSources: List[String], prefix: String = "search_count
     if (!input.isNullAt(0)) {
       input.getSeq(0).foreach { search_struct: Row =>
         val source: String = search_struct.getString(1)
-        val count: Long = search_struct.getLong(2)
-        if (count > 0 && validSources.contains(source)) {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/utils/udfs.scala
```diff
@@ -217,12 +217,14 @@ class AggSearchCounts(validSources: List[String], prefix: String = "search_count
     if (!input.isNullAt(0)) {
       input.getSeq(0).foreach { search_struct: Row =>
         val source: String = search_struct.getString(1)
-        val count: Long = search_struct.getLong(2)
-        if (count > 0 && validSources.contains(source)) {
-          buffer(0) = buffer.getLong(0) + count
-          // index is offset by one because the first field in the buffer is "all"
-          val index = validSources.indexOf(source) + 1
-          buffer(index) = buffer.getLong(index) + count
+        if (!search_struct.isNullAt(2)) {
```
sunahsuh: I'm fine with giving this a thumbs-up as is, but there's potentially a bit more idiomatic way to do this with `getAs[Option[Long]]`

relud: `getAs[Option[Long]]` fails the test case with: `java.lang.ClassCastException: java.lang.Long cannot be cast to scala.Option`

PR #405, opened 2018-03-28
Bug 1447648 - Add geo_subdivision{1,2} to main_summary_v4 and clients_daily_v6

relud:


src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -120,8 +120,8 @@ object ClientsDailyView {
     aggFirst("app_version"),
     aggFirst("blocklist_enabled"),
     aggFirst("channel"),
-    aggFirst("city"),
-    aggFirst("country"),
+    aggFirst(expr("IF(country IS NOT NULL AND country != '??', IF(city IS NOT NULL, city, '??'), NULL)"), "city"),
```
mreid-moz: It might be good to factor out a common `aggFirstExcept(column, stringToIgnore)` function, though I'm fine with doing that in a follow-up.

PR #406, opened 2018-03-29
Bug 1450079 - Add option to align number of input partitions

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1450079)

The `--alignment` option is an integer representing a multiple of
defaultParallelism. If this is set, then the number of partitions will
be `alignment*sc.defaultParallelism`.

The number of documents in the first-shutdown ping is much smaller than main pings. The number of input partitions is `total document size / 268 MB`. This means the default tuning for moztelemetry and MainSummaryView will lead to poor performance on smaller document sets. This improves the performance with a smaller set of machines in the cluster (10 vs 25) on a large dataset, or a smaller dataset with a larger number of machines (first-shutdown instead of main). 

The option preserves the current behavior, but it would marginally improve MainSummary run time with a value of 4 (3 hours vs 3.4 hours) if added to the airflow config.

## Performance Evaluation
The setup is 3 c3.4xlarge workers from ATMO. This uses a single day of first-shutdown pings from 2018-03-01.

### Unaligned
```
time spark-submit \
--master yarn \
--class com.mozilla.telemetry.views.MainSummaryView \
target/scala-2.11/telemetry-batch-view-1.1.jar \
--to 20180301 \
--from 20180301 \
--bucket telemetry-test-bucket \
--doc-type first_shutdown
```

The number of input partitions is 3.

```
real	39m42.862s
user	1m5.712s
sys	0m9.408s
```

### With alignment
```
time spark-submit \
--master yarn \
--class com.mozilla.telemetry.views.MainSummaryView \
target/scala-2.11/telemetry-batch-view-1.1.jar \
--to 20180301 \
--from 20180301 \
--bucket telemetry-test-bucket \
--doc-type first_shutdown \
--alignment 4
```

The number of input partitions is 256.

```
real    8m43.580s
user    1m3.660s
sys     0m8.820s
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ lazy val root = (project in file(".")).
     libraryDependencies += "org.apache.hadoop" % "hadoop-client" % "2.7.1" excludeAll(ExclusionRule(organization = "javax.servlet")),
     libraryDependencies += "org.apache.hadoop" % "hadoop-aws" % "2.7.1" excludeAll(ExclusionRule(organization = "javax.servlet")),
     libraryDependencies += "com.github.nscala-time" %% "nscala-time" % "2.10.0",
-    libraryDependencies += "org.rogach" %% "scallop" % "1.0.2",
+    libraryDependencies += "org.rogach" %% "scallop" % "3.1.2",
     libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion,
     libraryDependencies += "org.apache.spark" %% "spark-sql" % sparkVersion,
     libraryDependencies += "org.apache.spark" %% "spark-mllib" % sparkVersion,
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -109,6 +109,7 @@ object MainSummaryView {
     val docType = opt[String]("doc-type", descr = "DocType of pings conforming to main ping schema", required=false, default=Some("main"))
     // 500,000 rows yields ~ 200MB files in snappy+parquet
     val maxRecordsPerFile = opt[Int]("max-records-per-file", descr = "Max number of rows to write to output files before splitting", required = false, default=Some(500000))
+    val alignment = opt[Int]("alignment", descr = "Number of input partitions as a multiple `defaultParallelism`", required = false, default=None)
     verify()
   }
 
@@ -161,6 +162,9 @@ object MainSummaryView {
         case _ => "telemetry"
       }
 
+      // if `None`, then data is unaligned at 268MB chunks
+      val numPartitions = conf.alignment.get.map(n => n*sc.defaultParallelism)
+
       val messages = Dataset(telemetrySource)
         .where("sourceName") {
           case "telemetry" => true
@@ -176,7 +180,7 @@ object MainSummaryView {
           case channel => filterChannel.isEmpty || channel == filterChannel.get
         }.where("appVersion") {
           case v => filterVersion.isEmpty || v == filterVersion.get
-        }.records(conf.limit.get)
+        }.records(conf.limit.get, numPartitions)
 
       if(!messages.isEmpty()){
         val rowRDD = messages.flatMap(m => {
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -109,7 +109,8 @@ object MainSummaryView {
     val docType = opt[String]("doc-type", descr = "DocType of pings conforming to main ping schema", required=false, default=Some("main"))
     // 500,000 rows yields ~ 200MB files in snappy+parquet
     val maxRecordsPerFile = opt[Int]("max-records-per-file", descr = "Max number of rows to write to output files before splitting", required = false, default=Some(500000))
-    val alignment = opt[Int]("alignment", descr = "Number of input partitions as a multiple `defaultParallelism`", required = false, default=None)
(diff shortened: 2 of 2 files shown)


acmiyaguchi:
Hmm, so the option does two things. First it sets the the number of partitions to be aligned to the amount of default parallelism, and then it multiplies the number of partitions by the factor specified in the option. I'll try playing around with the wording a bit more, it's not as clear as it should be.

I'm under the strong assumption that it's the number of partitions as opposed to the packing strategy, but I see how the alternative packing strategy could affect runtime. I'm not entirely sure, but I have a feeling that it probably has less of an effect than the number of partitions because each json value has to be read as an object anyways. 

The relevant test that I can run is setting the number of equally sized partitions to something smaller than the defaultParallelism on a subset of data. If I set the number of partitions to 3, then the runtime on a single node will be significantly slower than if the number of partitions were 32 because of RDD immutability.  On main pings, all channels, the moderate improvement likely comes down to maximizing the batch RDD processing pipeline by reducing the number of hanging partitions

acmiyaguchi:
I came up with a test to see the effect of the packing strategy. I will use the fileLimit to fill n=defaultParallelism partitions. I'll then take the same number of partitions and fileLimit and process it using equally sized partitions. This should take less time than processing the entire day and give insight into the partition composition on performance.

fbertsch:
@acmiyaguchi seems like a good test, and probably something we should have done back when that change was first merged.

acmiyaguchi:
@fbertsch Interesting results inbound

First, I modified the test to just run on the nightly pings. Using `limit` only limits the number of gzip files read, which are inconsistent in size. This was also taking far longer than I had originally anticipated too on release.

Strangely enough, it seems that the 6 partitions chosen by the current method and the 6 partitions chosen by equally-sized groups perform very differently. It looks like the former strategy creates very imbalanced partitions that have a high variance in processing time (10.8 +/- 7.9, min: 1.1, max: 27). The latter processes each partition at 10 minutes with little variance (min: 10, max: 11).

Finally, I also ran the job by creating enough partitions for each worker to get a slice of the pie.

### Nightly, default settings
```
time spark-submit \
--master yarn \
--class com.mozilla.telemetry.views.MainSummaryView \
target/scala-2.11/telemetry-batch-view-1.1.jar \
--to 20180301 \
--from 20180301 \
--bucket telemetry-test-bucket \
--doc-type main \
--channel nightly
```

6 partitions
```
real    28m23.627s
user    1m10.940s
sys     0m9.444s
```

### Nighly, `numPartitions=6`
```
time spark-submit \
--master yarn \
--class com.mozilla.telemetry.views.MainSummaryView \
target/scala-2.11/telemetry-batch-view-1.1.jar \
--to 20180301 \
--from 20180301 \
--bucket telemetry-test-bucket \
--doc-type main \
--channel nightly \
--num-partitions 6
```

6 partitions

```
real    12m13.564s
user    1m4.196s
sys     0m9.108s
```

### Nightly, `numPartitions= 1 * sc.defaultParallelism`
```
time spark-submit \
--master yarn \
--class com.mozilla.telemetry.views.MainSummaryView \
target/scala-2.11/telemetry-batch-view-1.1.jar \
--to 20180301 \
--from 20180301 \
--bucket telemetry-test-bucket \
--doc-type main \
--channel nightly \
--alignment 1
```

96 partitions

```
real    4m54.553s
user    1m3.328s
sys     0m8.620s
```


#### Patch

```diff
diff --git a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
index 164fea0..4aa8a35 100644
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -109,6 +109,7 @@ object MainSummaryView {
     val docType = opt[String]("doc-type", descr = "DocType of pings conforming to main ping schema", required=false, default=Some("main"))
     // 500,000 rows yields ~ 200MB files in snappy+parquet
     val maxRecordsPerFile = opt[Int]("max-records-per-file", descr = "Max number of rows to write to output files before splitting", required = false, default=Some(500000))
+    val numPartitions = opt[Int]("num-partitions", required=false, default=None)
     val alignment = opt[Int]("alignment", descr = "Number of input partitions as a multiple `defaultParallelism`", required = false, default=None)
     verify()
   }
@@ -163,7 +164,10 @@ object MainSummaryView {
       }
 
       // if `None`, then data is unaligned at 268MB chunks
-      val numPartitions = conf.alignment.get.map(n => n*sc.defaultParallelism)
+      val numPartitions: Option[Int] = conf.numPartitions.get match {
+        case Some(x: Int) => Some(x)
+        case _ => conf.alignment.get.map(n => n*sc.defaultParallelism)
+      }
 
       val messages = Dataset(telemetrySource)
         .where("sourceName") {
```

fbertsch:
@acmiyaguchi very interesting results, thanks for investigating. I'm most interested in this fact: The min time for the old partition-packing strategy (take until it can't fit anymore, then move to next partition) is less than any other time. Am I reading that right - it took 1.1 minutes, compared to a min of 10 minutes for the partition-packing scheme?

acmiyaguchi:
@fbertsch That's right. I think that it would imply that it's a leftover of sorts.

Also, I modified the options a bit, do they make more sense than the original?

acmiyaguchi:
After taking a day to sit on the wording, I think this is much clearer than it was before. I've also filed a [follow-up bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1451103) for making evenly-spaced partitions aligned to parallelism the default.

PR #408, opened 2018-04-17
Bug 1449674 - Clean up DatasetComparator

relud:
and Bug 1453859 - make printing results optional

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
@@ -6,6 +6,11 @@ import org.rogach.scallop._
 object DatasetComparator {
   def jobName: String = "dataset_comparator"
 
+  case class Columns(originalOnly: Seq[String], testOnly: Seq[String], both: Seq[String])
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/utils/DatasetComparator.scala
```diff
@@ -47,106 +85,76 @@ object DatasetComparator {
...
+  def checkSame(result: Result): Boolean = {
+    result.columns.originalOnly.isEmpty &&
+    result.columns.testOnly.isEmpty &&
+    result.nulls.forall(pair => pair._2.change == 0) &&
+    result.rows.change == 0 &&
+    result.values.originalOnly == 0 &&
+    result.values.testOnly == 0
   }
 
+  case class Columns(originalOnly: Seq[String], testOnly: Seq[String], both: Seq[String])
```
acmiyaguchi: The case classes are useful definitions to have near the top of the class.

relud: fixed

PR #410, opened 2018-04-23
Bug 1444413 - remove cross_sectional

relud:
@harterrt for review as bugzilla component owner


src/main/scala/com/mozilla/telemetry/utils/Aggregation.scala
```diff
@@ -1,32 +0,0 @@
-package com.mozilla.telemetry.utils
-
-package object aggregation {
```
relud: as far as I can tell, this object was only used in CrossSectionalView

harterrt:
This is blocked on the discussion [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1444413#c4)

relud:
@harterrt distribution viewer is decommissioned, so this is ready to go

PR #411, opened 2018-04-23
Add Flash sandbox level userPrefs to Main Summary

hafta:
See Also: bug 1450715 - Add pref dom.ipc.plugins.sandbox-level.flash to telemetry
https://bugzilla.mozilla.org/show_bug.cgi?id=1450715


hafta:
@fbertsch thank you for helping with this

PR #412, opened 2018-04-24
Remove invalid column from clients_daily

relud:
`user_pref_browser_search_region` column does not exist in main_summary


sunahsuh:
Ah, doh --
@acmiyaguchi the change to add this field never made it into master 😿

acmiyaguchi:
@sunahsuh It happens, I did miss the train on the backfill. Why it didn't make it into the dev->master PR is a head scratcher though.

PR #413, opened 2018-04-25
Deploy JARs to s3

fbertsch:
Follow-up to the telemetry-streaming deploy. This is a bit different, it builds the JAR outside of the deploy script and just checks that the file exists. It builds the JAR inside docker.

Next pieces:
- Use deployed JARs in Airflow
- Use this `deploy.sh` file in `telemetry-streaming`.

Note this change was tested on the `deploy` branch, and the jar is in s3.


whd:
I think it would be sensible to keep the same deploy script between this and telemetry-streaming i.e. make this script the canonical version, and move `sbt assembly` out of the deploy script in telemetry-streaming into its `script` section (and also point at this script for its deploy). This is also more similar to lsb-extensions, which builds its artifacts using docker in the script step as well.

run-sbt.sh
```diff
@@ -8,7 +8,7 @@ mkdir -p ~/.ivy2/ ~/.sbt/
 # if we are not inside the docker container, run this command *inside* the
 # docker container
 if [ ! -f /.dockerenv ]; then
-    docker run -t -i -v ~/.ivy2:/root/.ivy2 -v ~/.sbt:/root/.sbt -v $PWD:/telemetry-batch-view telemetry-batch-view ./runtests.sh "$@"
+    docker run -t -i -v ~/.ivy2:/root/.ivy2 -v ~/.sbt:/root/.sbt -v $PWD:/telemetry-batch-view telemetry-batch-view ./run-sbt.sh "$@"
```
whd: Possible candidate for `$0`.

.travis.yml
```diff
@@ -1,13 +1,26 @@
 sudo: required
+
 services:
   - docker
+
 before_install:
+  - shopt -s expand_aliases
```
whd: Using an alias seems kind of silly but I'm not specifically objecting and don't have a better recommendation.

fbertsch:
@whd agreed, I'll be moving the `sbt assembly` out of the build file when I make this the canonical version.

PR #419, opened 2018-04-28
Add ability to auto-deploy

fbertsch:
The use-case for this feature is small changes that do not
necessarily need to be tested on stage, such as adding new
histograms to `main_summary`.

Simply include the string [auto-deploy] to make it so.


fbertsch:
@whd should we redirect this and https://github.com/mozilla/telemetry-batch-view/pull/417 to @haroldwoo?

PR #428, opened 2018-05-22
Add test duration logging and cache CrashAggregateView test fixture

jklukas:
For the following invocation:

./run-sbt.sh "testOnly *CrashAggregateViewTest"

runtime is 18 seconds with the lazy val.

Previously, this test took on order 5 minutes to run,
recreating the fixture for each individual call.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,7 +64,10 @@ lazy val Slow = config("slow").extend(Test)
 configs(Slow)
 inConfig(Slow)(Defaults.testTasks)
 
-testOptions in Test := Seq(Tests.Argument("-l", "org.scalatest.tags.Slow"))
+testOptions in Test := Seq(
+  Tests.Argument("-l", "org.scalatest.tags.Slow"),
+  // -oD add duration reporting; see http://www.scalatest.org/user_guide/using_scalatest_with_sbt
+  Tests.Argument("-oD"))
 testOptions in Slow := Seq()
 
 publishMavenStyle := true
--- a/build.sbt
+++ b/build.sbt
@@ -68,7 +68,9 @@ testOptions in Test := Seq(
   Tests.Argument("-l", "org.scalatest.tags.Slow"),
   // -oD add duration reporting; see http://www.scalatest.org/user_guide/using_scalatest_with_sbt
   Tests.Argument("-oD"))
-testOptions in Slow := Seq()
+testOptions in Slow := Seq(
+  // -oD add duration reporting; see http://www.scalatest.org/user_guide/using_scalatest_with_sbt
+  Tests.Argument("-oD"))
 
 publishMavenStyle := true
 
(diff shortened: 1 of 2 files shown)


jklukas:
It looks like this pattern of `def fixture` exists in many other tests, too, so if this makes sense, we can get some more boost in a follow-up PR.

build.sbt
```diff
@@ -64,8 +64,13 @@ lazy val Slow = config("slow").extend(Test)
 configs(Slow)
 inConfig(Slow)(Defaults.testTasks)
 
-testOptions in Test := Seq(Tests.Argument("-l", "org.scalatest.tags.Slow"))
-testOptions in Slow := Seq()
+testOptions in Test := Seq(
+  Tests.Argument("-l", "org.scalatest.tags.Slow"),
+  // -oD add duration reporting; see http://www.scalatest.org/user_guide/using_scalatest_with_sbt
+  Tests.Argument("-oD"))
```
fbertsch: nit: newline here

jklukas: Do we have code style defined? I know @akkomar was making notes about Scalacheck config. For languages that don't allow trailing commas, it's my preference to leave the closing parenthesis on he same line, since the line will need to change if a new arg is added later anyway.

If we already have a defined style, though, I will definitely defer to that.

akkomar: So far we have this: https://github.com/mozilla/moztelemetry/blob/master/scalastyle-config.xml, which is used in moztelemetry and telemetry-streaming. We'd like to have it hooked also in t-b-v.

fbertsch: Nope, no guide for this, but our scalacheck config lives here: https://github.com/mozilla/moztelemetry/blob/master/scalastyle-config.xml

The trailing `)` isn't really a big deal, I just personally like to read with newlines between more-than-1-line expressions. Feel free to merge as-is if you feel differently :)

akkomar: nit: it's scalastyle, not scalacheck :)

jklukas: scalastyle, yes. Too many `scala.+` names.

build.sbt
```diff
@@ -64,8 +64,14 @@ lazy val Slow = config("slow").extend(Test)
...
-testOptions in Test := Seq(Tests.Argument("-l", "org.scalatest.tags.Slow"))
-testOptions in Slow := Seq()
+testOptions in Test := Seq(
+  Tests.Argument("-l", "org.scalatest.tags.Slow"),
+  // -oD add duration reporting; see http://www.scalatest.org/user_guide/using_scalatest_with_sbt
+  Tests.Argument("-oD")
+)
+testOptions in Slow := Seq(
+  Tests.Argument("-oD")
+)
```
jklukas: Added newlines per @fbertsch. I'll bet scalastyle wouldn't be checking build.sbt even if we did have it set up here.

PR #431, opened 2018-05-23
Bug 1460356 - add attribution to clients_daily

relud:


mreid-moz:
@relud I forgot this needs a "release" step to deploy the change these days. Can you make a tag for this?

PR #432, opened 2018-05-29
Make tests more consistent

jklukas:
In support of [bug 1417571](https://github.com/mozilla/telemetry-batch-view/pull/432), this change puts tests into a more consistent
style before adding tags and splitting out testing on Travis to
multiple builds.

We follow the documentation for the test framework
(https://github.com/holdenk/spark-testing-base)
where the SparkContext and SqlContext are created and torn down
via methods inherited from DataFrameSuiteBase or
DatasetSuiteBase (if Hive support is needed for the tests).

We remove all use of our custom getOrCreateSparkContext function
in tests, and remove all explicit spark.stop() invocations.
This leads to a few other refactorings, including removing try/catch
blocks and turning some test fixtures into lazy vals in order to delay
their instantiation until the spark context is ready.


src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
```diff
@@ -123,8 +125,8 @@ class GenericCountTest extends FlatSpec with Matchers with BeforeAndAfterAll {
     "--output" :: "telemetry-test-bucket/client_count" :: Nil
 
   // make client count dataset
-  val conf = new GenericCountView.Conf(args.toArray)
-  private val aggregates = GenericCountView.aggregate(spark, conf)
+  val viewConf = new GenericCountView.Conf(args.toArray)
```
fbertsch: Why the name change here?

src/test/scala/com/mozilla/telemetry/views/RetentionViewTest.scala
```diff
@@ -96,7 +93,8 @@ class RetentionViewTest extends FlatSpec with Matchers with BeforeAndAfterAll {
     grouped.where(col("channel") === "beta").head.getLong(0) should be(2)
   }
 
-  override def afterAll() {
-    spark.stop()
+  override def beforeAll(): Unit = {
```
fbertsch: Dropped the `BeforeAndAfterAll` trait, but added a `beforeAll`; won't this not be called since it's not actually overriding anything?

src/test/scala/com/mozilla/telemetry/views/GenericCountViewTest.scala
jklukas: These two lines were generating errors because `conf` was already defined. I believe this is due to one of the supertraits pulled in by DataFrameSuiteBase: https://github.com/holdenk/spark-testing-base/blob/561dddb0d0894c79cf3613688f808aa0ab21f39f/src/main/1.3/java/com/holdenkarau/spark/testing/SharedJavaSparkContext.java#L38-L40

src/test/scala/com/mozilla/telemetry/views/RetentionViewTest.scala
jklukas: `DataFrameSuiteBase` extends `BeforeAndAfterAll`. We could consider including `BeforeAndAfterAll` explicitly if you think that enhances readability, but my preference is to keep the list of traits we're implementing shorter.

jklukas:
Added an unrelated commit to get past build failures introduced by https://github.com/mozilla/moztelemetry/pull/20

PR #433, opened 2018-05-30
Bug 1417571 - Improve test times via parallel Travis builds

jklukas:
This PR reduces the runtime for CI from > 30 minutes to ~12 minutes by splitting testing work into parallel jobs and moving deployment to a separate "stage" of the build.

See an example of the layout of jobs and associated runtimes in https://travis-ci.org/mozilla/telemetry-batch-view/builds/385852363

This PR consists of adding Scalatest tags in the Scala source and a significant refactoring of the .travis.yml in order to use those tags to define separate jobs.

We also add a little bit of test fixture cleanup to reduce duplicated work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,6 @@ jobs:
         - PURPOSE='stage the uberjar to S3'
       script:
         - pip install --user awscli
-        - export AWS_ACCESS_KEY_ID=$ARTIFACTS_KEY
-        - export AWS_SECRET_ACCESS_KEY=$ARTIFACTS_SECRET
-        - export AWS_REGION=$ARTIFACTS_REGION
         - docker build -t telemetry-batch-view .
         - docker-run ./run-sbt.sh "set test in assembly := {}" assembly && aws s3 cp $LOCAL_JAR $STAGED_JAR_URL
     # The deploy stage jobs only run once all jobs from the previous phase succeed.
@@ -68,9 +65,6 @@ jobs:
         - PURPOSE='deploy the staged uberjar to S3'
       script:
         - pip install --user awscli
-        - export AWS_ACCESS_KEY_ID=$ARTIFACTS_KEY
-        - export AWS_SECRET_ACCESS_KEY=$ARTIFACTS_SECRET
-        - export AWS_REGION=$ARTIFACTS_REGION
         - mkdir -p $(dirname $LOCAL_JAR)
         - aws s3 cp $STAGED_JAR_URL $LOCAL_JAR
         - JAR=$LOCAL_JAR bash deploy.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,40 +3,75 @@ sudo: required
 services:
   - docker
 
+# Try to keep things quick here since this section runs on every job whether it invokes sbt or not.
 before_install:
+  - export LOCAL_JAR='target/scala-2.11/telemetry-batch-view-1.1.jar'
+  - export STAGED_JAR_URL="s3://$ARTIFACTS_BUCKET/tmp/$TRAVIS_REPO_SLUG/build-$TRAVIS_BUILD_NUMBER.jar"
   - shopt -s expand_aliases
   - env | grep TRAVIS > .travis-env
   - CI_ENV=`bash <(curl -s https://codecov.io/env)`
-  - docker build -t telemetry-batch-view .
   - alias docker-run='docker run -v $PWD:/telemetry-batch-view $CI_ENV --env-file .travis-env telemetry-batch-view'
 
+env:
+  - SCALATEST_ARGS='-n ScalarAnalyzerBuild'
+  - SCALATEST_ARGS='-n ExperimentsBuild'
+  - SCALATEST_ARGS='-n ClientsDailyBuild'
+  # This final build must exclude (-l) all of the tags above and no others
+  # in order to assure we get full test coverage across the parallel builds.
+  - SCALATEST_ARGS='-l "ScalarAnalyzerBuild ExperimentsBuild ClientsDailyBuild"'
+
+# Code to run for the main test jobs.
 script:
   - export TEST_SKIP_REGEX="\[skip-tests\]"
   - if [[ (! $TRAVIS_COMMIT_MESSAGE =~ $TEST_SKIP_REGEX) || -n $TRAVIS_TAG || $TRAVIS_BRANCH == "master" ]]; then
-      docker-run ./run-sbt.sh coverage slow:test coverageReport;
+      docker build -t telemetry-batch-view .;
+      docker-run ./run-sbt.sh coverage "testOnly * -- $SCALATEST_ARGS" coverageReport;
     else
       echo "Skipping tests";
     fi
 
 before_deploy:
-  - export JAR="target/scala-2.11/telemetry-batch-view-1.1.jar"
-  - docker-run ./run-sbt.sh assembly
   - git config --local user.name "Auto Deployer"
   - git config --local user.email "telemetry-alerts@mozilla.com"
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -3,40 +3,75 @@ sudo: required
 services:
   - docker
 
+# Try to keep things quick here since this section runs on every job whether it invokes sbt or not.
 before_install:
+  - export LOCAL_JAR='target/scala-2.11/telemetry-batch-view-1.1.jar'
+  - export STAGED_JAR_URL="s3://$ARTIFACTS_BUCKET/tmp/$TRAVIS_REPO_SLUG/build-$TRAVIS_BUILD_NUMBER.jar"
```
jklukas: This new `tmp/` path was introduced in discussion with @whd in [bug 1465184](https://bugzilla.mozilla.org/show_bug.cgi?id=1465184). We use that area to share data between the test stage and the deploy stage.

.travis.yml
```diff
@@ -3,40 +3,75 @@ sudo: required
...
-      docker-run ./run-sbt.sh coverage slow:test coverageReport;
+      docker build -t telemetry-batch-view .;
+      docker-run ./run-sbt.sh coverage "testOnly * -- $SCALATEST_ARGS" coverageReport;
     else
       echo "Skipping tests";
     fi
 
 before_deploy:
-  - export JAR="target/scala-2.11/telemetry-batch-view-1.1.jar"
-  - docker-run ./run-sbt.sh assembly
```
jklukas: We move the `assembly` invocation here so that we can separate that to its own parallel job.

.travis.yml
```diff
@@ -3,40 +3,75 @@ sudo: required
...
-      repo: mozilla/telemetry-batch-view
-      condition: $TRAVIS_COMMIT_MESSAGE == *"[auto-deploy]"*
-      branch: master
-  - provider: script
-    script: bash deploy.sh
-    skip_cleanup: true
-    on:
-      repo: mozilla/telemetry-batch-view
-      all_branches: true
+jobs:
```
jklukas: This jobs section defines build stages. See https://docs.travis-ci.com/user/build-stages/ for an intro.

.travis.yml
```diff
@@ -3,40 +3,75 @@ sudo: required
...
+  include:
+    # Add an additional job to the 'test' stage for assembling the uberjar and staging in S3.
+    - stage: test
+      env:
+        - PURPOSE='stage the uberjar to S3'
+      script:
+        - pip install --user awscli
+        - export AWS_ACCESS_KEY_ID=$ARTIFACTS_KEY
+        - export AWS_SECRET_ACCESS_KEY=$ARTIFACTS_SECRET
+        - export AWS_REGION=$ARTIFACTS_REGION
```
jklukas: @whd - We could skip this if we changed the Travis repo configuration to set `AWS_ACCESS_KEY_ID` and friends rather than `ARTIFACTS_KEY` and friends. Do you see any danger in using those more general envvars?

whd: I guess I only ever read https://github.com/travis-ci/artifacts/blob/master/USAGE.md and not https://github.com/travis-ci/artifacts/#s3-environment-compatibility. I don't see an issue with using the general env vars and will change the configuration presently.

whd: This is done. I've redundantly added `AWS_DEFAULT_REGION` with the same value as `ARTIFACTS_REGION` because it appears that the artifact tool doesn't support reading from `AWS_DEFAULT_REGION`.

jklukas: Updated to implicitly use the new AWS_* variables.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
```diff
@@ -59,10 +60,15 @@ class ExperimentAnalysisViewTest extends FlatSpec with Matchers with DataFrameSu
     "histogram_content_gc_max_pause_ms_2"
   )
 
-  "Child Scalars" can "be counted" in {
+  lazy val id1Data = {
+    import spark.implicits._
+    predata.toDS().toDF().where(col("experiment_id") === "id1")
+  }
```
jklukas: This is a small performance improvement for this fixture since the same filter is used on `predata` in several tests.

jklukas:
I'm heading out for PTO this afternoon, working again on Wednesday. If you're comfortable with merging after review, feel free to do so while I'm out.

.travis.yml
```diff
@@ -3,40 +3,69 @@ sudo: required
...
+jobs:
+  include:
+    # Add an additional job to the 'test' stage for assembling the uberjar and staging in S3.
+    - stage: test
+      env:
+        - PURPOSE='stage the uberjar to S3'
+      script:
+        - pip install --user awscli
+        - docker build -t telemetry-batch-view .
+        - docker-run ./run-sbt.sh "set test in assembly := {}" assembly && aws s3 cp $LOCAL_JAR $STAGED_JAR_URL
```
fbertsch: We already have `test in assembly := {}` in `build.sbt`, is it required redundantly here?

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
fbertsch: Does this actually improve perf? Seems to me the lazy evaluation would cause this to make the time unchanged. It might improve if we cache the interim dataset.

.travis.yml
jklukas: I missed that bit in build.sbt. I'll take this out.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
jklukas: Making this a `val` means that the processing and filtering only needs to be evaluated once. So I think this is already doing what you propose above ("It might improve if we cache the interim dataset").

The reason it's lazy is to delay evaluation until the spark context is available (after beforeAll runs). So, it only gets evaluated once, but that evaluation happens on first use.

jklukas:
Adding more docs on maintaining tags and how to split tests is a good call. Will add.

jklukas:
@fbertsch - Thanks for your suggestions! I believe I've addressed all outstanding concerns either in code changes or comments here, so ready for another look.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
fbertsch: Ah there is some term overloading here between Spark and Scala. I'm referring to lazy evaluation of Spark DAGs. (I'm assuming you know this - but just so we're on the same page) This function returns a `DataFrame` which isn't evaluated until an action is called on it, e.g. `collect`, `show`.

This isn't a blocker, just a note of interest.

jklukas: I'm still fairly new to Spark, and lazy evaluation of DFs is something I had totally forgotten about. It's not clear to me, then, whether this changes anything in how Spark executes. It does reduce duplication in the code, at least.

fbertsch:
@jklukas one last nit, in the `README`, it might help to put a checklist of things to change when adding a new tag (big fan of checklists here). That isn't a blocker for this PR though.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
fbertsch: Agreed, a good change on that front.

jklukas:
These are good thoughts! Added a checklist and a note on limitations for parallelism.

PR #426, opened 2018-05-18
Versions update

jklukas:
Support for hyphen-separated key names was dropped in sbt 1.0.0;
see https://www.scala-sbt.org/1.x/docs/sbt-1.0-Release-Notes.html#sbt+1.0.0

Also updates to the Spark version used by current default EMR in ATMO.

PR #434, opened 2018-05-31
Spark to 2.3.0 and update other dependencies

jklukas:
This brings back in the Spark version update merged in #426 and subsequently backed out.

It also updates several other libraries to more recent (most recent where possible) versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,9 @@ val localMaven = "s3://net-mozaws-data-us-west-2-ops-mavenrepo/"
 
 resolvers += "S3 local maven snapshots" at localMavenHttps + "snapshots"
 
-val sparkVersion = "2.0.2"
+val sparkVersion = "2.3.0"
+// Should keep hadoop version in sync with the dependency defined by Spark.
+val hadoopVersion = "2.6.5"
 
 lazy val root = (project in file(".")).
   settings(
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -18,28 +20,29 @@ lazy val root = (project in file(".")).
     scalaModuleInfo := scalaModuleInfo.value.map(_.withOverrideScalaVersion(true)),
     libraryDependencies += "com.mozilla.telemetry" %% "moztelemetry" % "1.1-SNAPSHOT",
     libraryDependencies += "com.mozilla.telemetry" %% "spark-hyperloglog" % "2.0.0-SNAPSHOT",
-    libraryDependencies += "org.apache.avro" % "avro" % "1.7.7",
-    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.7.0",
-    libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.4",
```
jklukas: Moved all test dependencies to the bottom of the list and ensured they're only included in Test scope.

build.sbt
```diff
@@ -18,28 +20,29 @@ lazy val root = (project in file(".")).
...
+    libraryDependencies += "org.apache.avro" % "avro" % "1.8.2",
+    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.8.3",
     libraryDependencies += "net.sandrogrzicic" %% "scalabuff-runtime" % "1.4.0",
-    libraryDependencies += "com.holdenkarau" %% "spark-testing-base" % "2.0.0_0.4.7" % "test",
-    libraryDependencies += "org.xerial.snappy" % "snappy-java" % "1.1.2",
-    libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.2.10",
-    libraryDependencies += "joda-time" % "joda-time" % "2.9.2",
-    libraryDependencies += "org.apache.hadoop" % "hadoop-client" % "2.7.1" excludeAll(ExclusionRule(organization = "javax.servlet")),
-    libraryDependencies += "org.apache.hadoop" % "hadoop-aws" % "2.7.1" excludeAll(ExclusionRule(organization = "javax.servlet")),
-    libraryDependencies += "com.github.nscala-time" %% "nscala-time" % "2.10.0",
```
jklukas: It appears we don't use nscala-time anymore, so removed it.

build.sbt
```diff
@@ -18,28 +20,29 @@ lazy val root = (project in file(".")).
...
+    libraryDependencies += "org.apache.hadoop" % "hadoop-aws" % hadoopVersion excludeAll(ExclusionRule(organization = "javax.servlet")),
     libraryDependencies += "org.rogach" %% "scallop" % "3.1.2",
     libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion,
     libraryDependencies += "org.apache.spark" %% "spark-sql" % sparkVersion,
     libraryDependencies += "org.apache.spark" %% "spark-mllib" % sparkVersion,
     libraryDependencies += "org.apache.spark" %% "spark-hive" % sparkVersion,
-    libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.3.0",
+    libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.0",
     libraryDependencies += "org.yaml" % "snakeyaml" % "1.21",
-    libraryDependencies += "com.google.protobuf" % "protobuf-java" % "2.5.0",
```
jklukas: We don't directly use this dependency, so removing it from the list here and instead just relying on the override.

run-sbt.sh
```diff
@@ -12,15 +12,18 @@ if [ ! -f /.dockerenv ]; then
     exit $?
 fi
 
+# Set options that sbt will pass to the JVM
+export SBT_OPTS="-Xss2M -Djava.security.policy=java.policy"
```
jklukas: The security policy is necessary due to a change in derby as discussed in #426

jklukas:
Ran main_summary on nightly data using the master branch and this branch. The DatasetComparator confirmed that the output datasets were the same.

jklukas:
Well, I tested this and then changed just a _few little things_. Tests are now failing, so this will likely need to wait for next week.

run-sbt.sh
fbertsch: It might be good to put a link here for reference.

build.sbt
```diff
@@ -16,30 +18,38 @@ lazy val root = (project in file(".")).
     // Hack to get releases to not fail to upload when the same jar name already exists. Later we will need auto versioning
     isSnapshot := true,
     scalaModuleInfo := scalaModuleInfo.value.map(_.withOverrideScalaVersion(true)),
+
+    // Snapshot dependencies; keep in mind that changes to the master branch of these projects will get pulled in
+    // automatically on next build and could cause unexpected problems while working on unrelated changes.
     libraryDependencies += "com.mozilla.telemetry" %% "moztelemetry" % "1.1-SNAPSHOT",
     libraryDependencies += "com.mozilla.telemetry" %% "spark-hyperloglog" % "2.0.0-SNAPSHOT",
-    libraryDependencies += "org.apache.avro" % "avro" % "1.7.7",
-    libraryDependencies += "org.apache.parquet" % "parquet-avro" % "1.7.0",
```
jklukas: These dependencies have moved further down the list and have updated versions

build.sbt
```diff
@@ -16,30 +18,38 @@ lazy val root = (project in file(".")).
...
-    libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.4",
-    libraryDependencies += "net.sandrogrzicic" %% "scalabuff-runtime" % "1.4.0",
-    libraryDependencies += "com.holdenkarau" %% "spark-testing-base" % "2.0.0_0.4.7" % "test",
-    libraryDependencies += "org.xerial.snappy" % "snappy-java" % "1.1.2",
-    libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.2.10",
```
jklukas: Removed the json4s-jackson so that we get the version spark-core wants transitively.

build.sbt
```diff
@@ -53,7 +63,7 @@ test in assembly := {}
 
 assemblyMergeStrategy in assembly := {
   case PathList("META-INF", xs @ _*) => MergeStrategy.discard
-  case x => MergeStrategy.first
+  case _ => MergeStrategy.first
```
jklukas: Unrelated style fixup. Using underscore to indicate that we don't care about the value in this case.

run-sbt.sh
jklukas: There's a link in the java.policy file itself, and explanation of why it exists. Does that seem sufficient?

jklukas:
This is passing again with `java.policy` taken out, so ready for review again, @fbertsch

PR #435, opened 2018-06-04
Remove heavy users

fbertsch:
Let me know if you see something I missed.


jklukas:
I just hit "update branch". This should be able to pass tests again with #436 merged.

fbertsch:
@mreid-moz they're in the git log, we can always dust 'em off if we want them.

PR #436, opened 2018-06-06
Fix problem that caused CI on forks to fail

jklukas:
We were trying to build an uberjar and stage in S3,
but that requires encrypted env vars that aren't avaialable to
builds on forks, so we need to make the assembly and deploy stages conditional.


jklukas:
https://travis-ci.org/mozilla/telemetry-batch-view/builds/389414524 was successful, so merging. The codecov report is erroneous, I believe, because one of the jobs died and had to be restarted, which confused the merging of reports.

PR #437, opened 2018-06-13
Use new base image that has sbt baked in

jklukas:
This should make startup a bit faster and it will avoid the
problem on TravisCI where parallel jobs all hit bintray at the
same time, leading to throttling and intermittent failed jobs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,4 @@
-FROM java:8
-
-ENV SBT_VERSION 1.1.1
-ENV HBASE_VERSION 1.2.3
-
-# Install sbt
-RUN \
-  curl -L -o sbt-$SBT_VERSION.deb http://dl.bintray.com/sbt/debian/sbt-$SBT_VERSION.deb && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,16 +1,4 @@
...
-
-# Install sbt
-RUN \
-  curl -L -o sbt-$SBT_VERSION.deb http://dl.bintray.com/sbt/debian/sbt-$SBT_VERSION.deb && \
-  dpkg -i sbt-$SBT_VERSION.deb && \
-  rm sbt-$SBT_VERSION.deb && \
-  apt-get update && \
-  apt-get install sbt && \
-  sbt sbtVersion
+FROM hseeberger/scala-sbt:8u151-2.12.4-1.1.1
```
relud: we should make our own image `mozilla/scala-sbt` and use that

jklukas: Updated. See https://github.com/mozilla/scala-sbt/pull/1 for the new mozilla Dockerfile.

PR #440, opened 2018-06-16
Use CircleCI 2.0 in place of TravisCI

jklukas:
This consolidates us down to one CI provider for t-b-v and takes
advantage of CircleCI's more powerful machinery for defining workflows
with parallel jobs and shared files between them.

We eliminate the Scalatest tags we previously used to manually assign
test cases to different build jobs. We now split up tests by source file
and use CircleCI's 'parallelize' config option to reliably assign them
to jobs.

We also eliminate the concept of git tags for controlling deploys,
as discussed during All Hands week. We always upload a successful
build as the current jar for the branch. This means now that whatever
hits `master` will be what's running in production.

The local workflow is simplified somewhat by no longer building a docker
image specific to this project, but rather relying on the mozilla/sbt
image.

This CircleCI configuration runs a bit faster and with better consistency
(9 to 10 minutes) compared to the current TravisCI configuration
(12 minutes to 20 minutes).


jklukas:
Before merging this, we would need to modify configuration for this repo to require circleci passes rather than travis-ci, so that we don't get the "continuous-integration/travis-ci Expected — Waiting for status to be reported" message anymore.

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+#   https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+####################
+
+
+####################
+# Templates: see "anchors" in https://learnxinyminutes.com/docs/yaml/
+####################
+
+# See available image tags at https://hub.docker.com/r/mozilla/sbt/
+sbt_image: &sbt_image mozilla/sbt:8u171_1.1.6
```
jklukas: Replaces this repo's Dockerfile

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+# At time of writing, we use the default container size (2 CPUs and 4GB memory),
+# but may renogotiate with CircleCI and enable larger resource_class
+# options at which point we might want to bump up the Java memory settings
+# and reduce the test parallelism.
+test_settings: &test_settings
+  docker:
+    - image: *sbt_image
+      environment:
+        _JAVA_OPTIONS: "-Xms3G -Xmx3G -Xss4M -XX:MaxMetaspaceSize=512M"
+  working_directory: /telemetry-batch-view
```
jklukas: This section provides what used to be in the Dockerfile

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+
+# At time of writing, we use the default container size (2 CPUs and 4GB memory),
+# but may renogotiate with CircleCI and enable larger resource_class
+# options at which point we might want to bump up the Java memory settings
+# and reduce the test parallelism.
+test_settings: &test_settings
+  docker:
+    - image: *sbt_image
+      environment:
+        _JAVA_OPTIONS: "-Xms3G -Xmx3G -Xss4M -XX:MaxMetaspaceSize=512M"
```
jklukas: We tune memory down to 3G since the container size we have available only has 4GB memory.

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+            cp $JAR "$JAR_DIR/$CIRCLE_BRANCH/$JAR_NAME"
+            cp $TXT_NAME "$JAR_DIR/$CIRCLE_BRANCH/$TXT_NAME"
+
+            # Now upload to S3; see docs for artifact upload tool at https://github.com/travis-ci/artifacts
+            export ARTIFACTS_DEST="$HOME/bin/artifacts"
+            export ARTIFACTS_PERMISSIONS="public-read"
+            export ARTIFACTS_PATHS="$CIRCLE_PROJECT_USERNAME/"
+            export ARTIFACTS_TARGET_PATHS="/$CIRCLE_PROJECT_USERNAME/"
+            curl -sL https://raw.githubusercontent.com/travis-ci/artifacts/master/install | bash
+            $ARTIFACTS_DEST upload
```
jklukas: The content here is mostly copied directly out of deploy.sh, with a little bit of refactoring and updating to use CIRCLE_ variables. We no longer need the full complexity of the code in deploy.sh because we don't have to deal with detecting git tags.

build.sbt
```diff
@@ -1,4 +1,5 @@
 javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint")
+scalacOptions ++= Seq("-Xmax-classfile-name", "242")
```
jklukas: The docker environment on CircleCI ran into some compilation issues due to overlong generated filenames.

.circleci/config.yml
akkomar: Shouldn't we set the same values in `.sbtopts`, just to make it consistent?

jklukas: It's not clear to me why we have both. I'm having trouble finding any documentation about `.sbtopts`. Can we just rely on that and skip the environment variable completely?

jklukas: https://circleci.com/blog/how-to-handle-java-oom-errors/ indicates that `_JAVA_OPTIONS` is Oracle-specific, so this may not even be taking effect in our openjdk-based containers.

deploy.sh
```diff
@@ -1,62 +0,0 @@
-#!/bin/bash
```
jklukas: telemetry-streaming pulls in the master version of this file, so builds there will break if this is merged.

https://github.com/mozilla/telemetry-streaming/pull/144 should be merged first, which pins telemetry-streaming to a particular commit of this file.

.circleci/deploy.sh
```diff
@@ -0,0 +1,48 @@
...
+# when proposing changes.
+
+# Copy $JAR and metadata to a local directory, then upload to S3.
+# This uses the travis-ci/artifacts tool which references environment variables
+# prefixed by AWS_ and ARTIFACTS_, some of which are injected by the CircleCI project.
+
+set -x
+
+# Change PWD to the root of the git repo.
+cd $(git rev-parse --show-toplevel)
```
haroldwoo: Where is circleci's working directory?  Should this be pushed/popped?

run-sbt.sh
```diff
@@ -1,26 +1,20 @@
...
 # create dirs to cache .ivy2 and .sbt if they don't exist already
 mkdir -p ~/.ivy2/ ~/.sbt/
 
-# if we are not inside the docker container, run this command *inside* the
-# docker container
-if [ ! -f /.dockerenv ]; then
-    docker run -t -i -v ~/.ivy2:/root/.ivy2 -v ~/.sbt:/root/.sbt -v $PWD:/telemetry-batch-view telemetry-batch-view ./run-sbt.sh "$@"
-    exit $?
-fi
+IMAGE=$(sed -n 's/.*&sbt_image \(.*\)/\1/p' .circleci/config.yml)
```
haroldwoo: comment what this does e.g. pull image name from conf file

.jvmopts
```diff
@@ -0,0 +1,6 @@
+# See 'sbt -help' for docs on how .jvmopts and .sbtopts are used.
+-Xms3G
+-Xmx3G
```
haroldwoo: 3 is enough now?

run-sbt.sh
jklukas: will do.

.jvmopts
jklukas: 3GB is enough when running only a subset of tests, which is what we now do in CI. It probably slows down running the entire test suite, but that's become untenable at the current size.

I'd like to see us bump the memory up, but that will require us finishing contract discussions with CircleCI to get access to larger containers.

haroldwoo:
I've modified the repo to require circle/assembly, test, and deploy checks to pass prior to merging

PR #442, opened 2018-06-19
Use Scalastyle for enforcing style rules

akkomar:
This introduces Scalastyle to t-b-v. In cases where simple refactorings were not enough for adhering to style rules, checker is turned off (using `// scalastyle:off` comments). These cases will be fixed in separate PRs.

Also:
* sbt version is bumped
* some useful compiler warnings are turned on (next step would be getting rid of them and failing the build on any warnings)

Before merging, Scalastyle-related commits should be squashed.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
@@ -1,14 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.views
 
-import com.mozilla.telemetry.utils.getOrCreateSparkSession
-import com.mozilla.telemetry.utils.{AggSearchCounts, AggMapFirst}
-import org.apache.spark.sql.{Column, DataFrame}
+import com.mozilla.telemetry.utils.{AggMapFirst, AggSearchCounts, getOrCreateSparkSession}
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.{Column, DataFrame}
 import org.rogach.scallop._
 
 object ClientsDailyView {
-  val jobName: String = "clients_daily"
-  val schemaVersion: String = "v6"
+  private val jobName: String = "clients_daily"
+  private val schemaVersion: String = "v6"
 
   class Conf(args: Array[String]) extends ScallopConf(args) {
     val date = opt[String](
@@ -91,18 +93,18 @@ object ClientsDailyView {
       )
   }
 
-  def aggFirst(field: String): Column = first(field, ignoreNulls = true).alias(field)
+  private def aggFirst(field: String): Column = first(field, ignoreNulls = true).alias(field)
 
-  def aggFirst(expression: Column, alias: String): Column = first(expression, ignoreNulls = true).alias(alias)
+  private def aggFirst(expression: Column, alias: String): Column = first(expression, ignoreNulls = true).alias(alias)
 
-  val mapFirst = new AggMapFirst()
-  def aggMapFirst(field: String): Column = mapFirst(col(field)).alias(field)
+  private val mapFirst = new AggMapFirst()
+  private def aggMapFirst(field: String): Column = mapFirst(col(field)).alias(field)
 
-  def aggMax(field: String): Column = max(field).alias(s"${field}_max")
+  private def aggMax(field: String): Column = max(field).alias(s"${field}_max")
 
-  def aggMean(field: String): Column = mean(field).alias(s"${field}_mean")
+  private def aggMean(field: String): Column = mean(field).alias(s"${field}_mean")
 
-  val searchSources = List(
+  private val searchSources = List(
     "abouthome",
     "contextmenu",
     "newtab",
@@ -110,14 +112,14 @@ object ClientsDailyView {
     "system",
     "urlbar"
   )
-  val searchCounts = new AggSearchCounts(searchSources)
-  def aggSearchCounts(field: String): Column = searchCounts(col(field)).alias(field)
+  private val searchCounts = new AggSearchCounts(searchSources)
+  private def aggSearchCounts(field: String): Column = searchCounts(col(field)).alias(field)
 
-  def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
+  private def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
 
-  def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
+  private def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
 
-  def fieldAggregators = List(
+  private val fieldAggregators = List(
     aggSum("aborts_content"),
     aggSum("aborts_gmplugin"),
     aggSum("aborts_plugin"),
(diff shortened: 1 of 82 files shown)


src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -91,33 +94,33 @@ object ClientsDailyView {
...
+  private def aggSearchCounts(field: String): Column = searchCounts(col(field)).alias(field)
 
-  def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
+  private def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
 
-  def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
+private   def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
 
-  def fieldAggregators = List(
+  private val fieldAggregators = List(
```
jklukas: Is the change from def to val intended here?

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -11,13 +14,13 @@ import org.rogach.scallop.ScallopConf
...
-  def schemaVersion = "v1"
+  private val defaultErrorAggregatesBucket = "net-mozaws-prod-us-west-2-pipeline-data"
+  private val errorAggregatesPath = "experiment_error_aggregates/v1"
+  private val jobName = "experiment_analysis"
+  private val schemaVersion = "v1"
   // This gives us ~120MB partitions with the columns we have now. We should tune this as we add more columns.
-  def rowsPerPartition = 25000
-  val defaultJackknifeBlocks = 100
+  private val rowsPerPartition = 25000
+  private val defaultJackknifeBlocks = 100
```
jklukas: def to val here too

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
akkomar: Yeah, it's indistinguishable on the bytecode level and makes these fields consistent with others.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
akkomar: https://github.com/mozilla/telemetry-batch-view/pull/442#discussion_r196412975

akkomar:
Sure, I think running main summary is more than enough. I've been pretty conservative here - any code that required more than simple refactors is left unchanged with style checker turned off for now.

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -91,33 +94,33 @@ object ClientsDailyView {
...
-  val searchCounts = new AggSearchCounts(searchSources)
-  def aggSearchCounts(field: String): Column = searchCounts(col(field)).alias(field)
+  private val searchCounts = new AggSearchCounts(searchSources)
+  private def aggSearchCounts(field: String): Column = searchCounts(col(field)).alias(field)
 
-  def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
+  private def aggSum(field: String): Column = sum(field).alias(s"${field}_sum")
 
-  def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
+private   def aggSum(expression: Column, alias: String): Column = sum(expression).alias(alias)
```
mreid-moz: nit: indentation / spacing

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -206,7 +209,10 @@ object ClientsDailyView {
     aggSum("push_api_notify"),
     aggFirst("sample_id"),
     aggFirst("sandbox_effective_content_process_level"),
-    aggSum(col("scalar_parent_webrtc_nicer_stun_retransmits") + col("scalar_content_webrtc_nicer_stun_retransmits"), "scalar_combined_webrtc_nicer_stun_retransmits_sum"),
+    aggSum(
+      col("scalar_parent_webrtc_nicer_stun_retransmits") +
```
mreid-moz: Personal preference - if it will still pass style checks, I would prefer to see:
```
aggSum(
  col("scalar_parent_webrtc_nicer_stun_retransmits") + col("scalar_content_webrtc_nicer_stun_retransmits"),
  "scalar_combined_webrtc_nicer_stun_retransmits_sum"
)
```
This will keep the expression in the first argument together.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -1,23 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
mreid-moz: I'm hoping to deprecate this view soon - it might be preferable to limit changes to adding the License header and turning scalastyle off. @fbertsch what do you think of this?

src/test/scala/com/mozilla/telemetry/utils/MainPingTest.scala
```diff
@@ -578,31 +581,34 @@ class MainPingTest extends FlatSpec with Matchers {
...
+    (makeExpected(List(null, null, null, null, null, null, 2, 97, null, null)))
 
     // Keyed scalars convert correctly
     val jKeyedScalar = parse(
       """{
         |  "mock.keyed.scalar.uint": {"a": 1, "b": 2}
         |}""".stripMargin)
     MainPing.scalarsToRow(makeScalarsMap(jKeyedScalar), scalarDefs) should be
-      (makeExpected(List(null, null, null, Map("a" -> 1, "b" -> 2), null, null, null, null, null, null)))
+    (makeExpected(List(null, null, null, Map("a" -> 1, "b" -> 2), null, null, null, null, null, null)))
```
mreid-moz: Hm, not indented here? Isn't this a continuation of the statement on the previous line?

akkomar: Yes, it is, I'm using Intellij's automatic formatting here.

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
jklukas: > it's indistinguishable on the bytecode level

I don't understand how that would be the case. I'd expect a `def` to be evaluated on each use, returning a new `List` each time whereas a `val` would be evaluated at definition as part of the constructor. It looks like both these cases of `def -> val` are fine as construction-time values, but there are certainly cases where `lazy val` or `def` is significant because evaluation needs to be delayed until the spark context is available.

akkomar: You're right, there actually are some differences on the bytecode level (I was thinking about constant values). But still, in this case I think we're not winning anything with `def`, and `val` better describes nature of this field.

jklukas: I would agree that `val` appears to better describe the nature of the field, but there's sometimes surprising implicit behavior. As long as this code is exercised by the tests and is passing, I'm fine with the change.

src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
jklukas: This one looks clearly fine as a val.

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
akkomar: Great, it is tested in https://github.com/mozilla/telemetry-batch-view/blob/af564d5611d4ce1f423e66376e0ee5f666fb622c/src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTest.scala#L16

PR #446, opened 2018-06-27
Force clients_daily active_hours_sum to type double

relud:


relud:
relatedly: [ClientsDailyView.scala#L249](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala#L249) and [ClientsDailyView.scala#L241](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala#L241) also do division, but those have been returning `decimal` columns all along.

relud:
The issue is that in `emr-5.13` spark got better at determining the precision needed to store `(3600.0/5)`, so it `active_hours_sum` no longer exceeded the precision limits of the `decimal` type, and so the type stopped falling back to `double`

PR #447, opened 2018-07-02
Update Main Events schema to match event ping events

sunahsuh:
As mentioned in mozilla/telemetry-streaming#149, this is the update to main ping events output to match the new schema from events ping output. The only really interesting part is the new session_start_time field, which matches the format of the session_start_time payload field in the events ping.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainEventsView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainEventsView.scala
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.views
 
-import com.mozilla.telemetry.utils.{Events, S3Store, getOrCreateSparkSession}
+import com.mozilla.telemetry.utils.{S3Store, getOrCreateSparkSession}
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.functions._
 import org.joda.time.{DateTime, Days, format}
@@ -13,6 +13,7 @@ object MainEventsView {
   private val logger = org.apache.log4j.Logger.getLogger(this.getClass.getName)
 
   def schemaVersion: String = "v1"
+
   def jobName: String = "events"
 
   // Configuration for command line arguments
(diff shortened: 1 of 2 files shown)


jklukas:
circleci is showing a compilation error:

```
[error] /telemetry-batch-view/src/test/scala/com/mozilla/telemetry/views/MainEventsViewTest.scala:82:75: integer number too large
[error]     events.select("session_start_time").collect.head.getLong(0) should be(1528394400000)
```

src/main/scala/com/mozilla/telemetry/views/MainEventsView.scala
```diff
@@ -87,12 +88,7 @@ object MainEventsView {
   }
 
   def eventsFromMain(mainSummaryData: DataFrame, sampleId: Option[String]): DataFrame = {
-    val eventsSchema = Events.buildEventSchema
     val partialDataFrame = mainSummaryData
-      .select("document_id", "client_id", "normalized_channel", "country", "locale", "app_name", "app_version", "os",
-        "os_version", "e10s_enabled", "subsession_start_date", "subsession_length", "sync_configured",
-        "sync_count_desktop", "sync_count_mobile", "timestamp", "sample_id", "active_experiment_id",
-        "active_experiment_branch", "experiments", "events")
```
jklukas: Why is this `select` going away? Is it simply redundant with the `exploded.selectExpr` below? Is Spark smart enough to prune the unused columns early if we don't select down the column list here?

src/test/scala/com/mozilla/telemetry/views/MainEventsViewTest.scala
```diff
@@ -84,5 +78,7 @@ class MainEventsViewTest extends FlatSpec with Matchers with DataFrameSuiteBase
     sampledEvents
       .where("document_id = '72062950-3daf-450e-adfd-58eda3151a97'")
       .count should be(1)
+
+    events.select("session_start_time").collect.head.getLong(0) should be(1528394400000)
```
jklukas: Probably need an `L` suffix on the literal here.

PR #450, opened 2018-07-04
Bug 1469388 - Add generic dau view

relud:
Creates datasets that replace queries like `mau_dau_er_*` in [mozilla/stmo_core_product_metrics/health](https://github.com/mozilla/stmo_core_product_metrics/tree/master/health)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/dau/DesktopDauView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/dau/DesktopDauView.scala
@@ -1,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.views.dau
+
+import org.rogach.scallop.{ScallopConf, ScallopOption}
+
+object DesktopDauView extends GenericDauTrait {
+  class CliConf(args: Array[String]) extends ScallopConf(args) {
+    val to: ScallopOption[String] = opt[String](
+      "to",
+      descr = "newest date for with to generate counts (inclusive)",
+      required = true)
+    val from: ScallopOption[String] = opt[String](
+      "from",
+      descr = "oldest date for which to generate counts, default is `--to` value (inclusive)",
+      required = false)
+    val bucket: ScallopOption[String] = opt[String](
+      "bucket",
+      default = Some("s3://telemetry-parquet"),
+      descr = "location where input and output data sets are stored",
+      required = false)
+    verify()
+  }
+
+  val jobName: String = "desktop_dau"
+
+  def getGenericDauConf(args: Array[String]): GenericDauConf = {
+    val conf = new CliConf(args)
+    GenericDauConf(
+      conf.from.getOrElse(conf.to()),
+      conf.to(),
+      inputBasePath = s"${conf.bucket()}/client_count_daily/v2",
+      outputBasePath = s"${conf.bucket()}/dau/v1"
+    )
+  }
+}
--- a/src/main/scala/com/mozilla/telemetry/views/dau/GenericDauTrait.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/dau/GenericDauTrait.scala
@@ -1,0 +1,188 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.views.dau
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+import java.time.temporal.ChronoUnit.DAYS
+
+import com.mozilla.telemetry.utils.getOrCreateSparkSession
+import com.mozilla.telemetry.utils.UDFs.{HllMerge, MozUDFs}
+import org.apache.spark.sql.expressions.Window
+import org.apache.spark.sql.{DataFrame, functions => F}
+
+trait GenericDauTrait {
+  val jobName: String
+  def getGenericDauConf(args: Array[String]): GenericDauConf
+
+  def main(args: Array[String]) {
+    val conf = getGenericDauConf(args)
+    val spark = getOrCreateSparkSession(jobName)
+    // date partition should be read with a string type to minimize string parsing
+    spark.conf.set("spark.sql.sources.partitionColumnTypeInference.enabled", "false")
+
+    val input = spark
+      .read
+      // merge schemas to handle schema evolution
+      .option("mergeSchema", "true")
+      // detect input date column
+      .option("basePath", conf.inputBasePath)
+      // read input dates by path
+      .parquet(conf.inputPaths:_*)
+
+    val result = aggregate(input, conf)
+
+    // write one output date at a time to overwrite at the date level
+    conf.outputDates.values.foreach { date =>
+      result
+        // get one day
+        .where(s"${conf.outputDateColumn}='$date'")
+        // drop old date column format
+        .drop(conf.outputDateColumn)
+        // write with new date column format
+        .write
+        .mode("overwrite")
+        .parquet(s"${conf.outputBasePath}/${conf.outputDateColumn}=$date")
+    }
+  }
+
+  private val SecondsPerDay: Int = 86400
+  def aggregate(input: DataFrame, conf: GenericDauConf): DataFrame = {
+    // register hll functions
+    input.sparkSession.registerUDFs // for hll_create and hll_cardinality
+    input.sparkSession.udf.register("hll_merge", HllMerge)
+
+    // common window config
+    val baseWindow = Window.orderBy(s"int_${conf.inputDateColumn}").partitionBy(conf.columns.map(F.col): _*)
+    // window for calculating mau
+    val mauWindow = baseWindow.rangeBetween(-conf.PrecedingDaysInMonth, Window.currentRow)
+    // window for calculating smoothed_dau
+    val smoothedDauWindow = baseWindow.rangeBetween(-conf.PrecedingDaysInWeek, Window.currentRow)
+    val outputDates = conf.outputDates
+    val outputDateUdf = F.udf((d: String)=>outputDates.getOrElse(d, null))
+
+    val input_with_hll = input
+      // handle conf.where
+      .where(conf.where)
+      // handle conf.columns
+      .groupBy(conf.inputDateColumn, conf.columns: _*)
+      // generate daily hlls
+      .agg(F.expr(s"hll_merge(${conf.hllColumnExpression})").as(conf.hllColumn))
+
+    // handle conf.top
+    conf.top.foldRight(input_with_hll) { case ((column, limit), df) =>
+      // use join to filter input by top "limit" "column" values (eg top 100 country values)
+      df.join(
+        input_with_hll
+          // count by "column"
+          .groupBy(column)
+          .agg(F.expr(s"hll_cardinality(hll_merge(${conf.hllColumn}))").as("count"))
+          // filter to top "limit" counts
+          .orderBy(F.desc("count"))
+          .limit(limit)
+          // collect only "column" values
+          .select(column),
+        column
+      )
+    }
+      // handle conf.minDauThreshold
+      .where(conf.minDauThresholdWhere)
+      // add temporary integer date column as days since epoch
+      .withColumn(
+      s"int_${conf.inputDateColumn}",
+      F.floor(F.unix_timestamp(F.to_date(F.col(conf.inputDateColumn), conf.inputDatePattern)) / SecondsPerDay))
+      // calculate mau hll
+      .withColumn("mau", F.expr(s"hll_merge(${conf.hllColumn})") over mauWindow)
+      // calculate cardinality separately to avoid AnalysisException
+      .withColumn("mau", F.expr("hll_cardinality(mau)"))
+      // calculate dau
+      .withColumn("dau", F.expr(s"hll_cardinality(${conf.hllColumn})"))
+      // don't calculate smoothed_dau for unnecessary
+      .filter(F.col(conf.inputDateColumn).isin(conf.smoothedDauDates: _*))
+      // calculate smoothed_dau
+      .withColumn("smoothed_dau", F.expr("avg(dau)") over smoothedDauWindow)
+      // drop temporary integer date column
+      .drop(s"int_${conf.inputDateColumn}")
+      // calculate er
+      .withColumn("er", F.expr("smoothed_dau/mau"))
+      // rename input date column to output date column
+      .withColumnRenamed(conf.inputDateColumn, conf.outputDateColumn)
+      // reformat input date format to output date format
+      .withColumn(conf.outputDateColumn, outputDateUdf(F.col(conf.outputDateColumn)))
+  }
+}
+
+/** Expanded configuration for GenericDauTrait.aggregate
+  *
+  * @param from oldest date in output date range (inclusive)
+  * @param to newest date in output date range (inclusive)
+  * @param inputBasePath path to an existing parquet table partitioned by "inputDateColumn"
+  * @param outputBasePath path where output will be written, partitioned by "outputDateColumn"
+  * @param additionalOutputColumns additional columns to include in output and break down dau by
+  * @param countColumn optional column generate "hllColumn" from. if absent then "hllColumn" must be in input
+  * @param hllBytes number of bytes to use when generating "hllColumn" from "countColumn"
+  * @param hllColumn hll column used to generate counts
+  * @param inputDateColumn name of date partition in input table
+  * @param inputDatePattern java date pattern to use to interpret inputDateColumn
+  * @param minDauThreshold optional filter for minimum value of dau in output (inclusive)
+  * @param outputDateColumn name to use for date partition in output table
+  * @param outputDatePattern java date pattern to use to interpret "outputDateColumn"
+  * @param top for each (column,limit) pair filter input to top "limit" "column" values by hll count,
+  *            such as ("country",100) for the top 100 countries by hll count,
+  *            each "column" is included in output
+  * @param where where clause by which to filter input, applies before "top"
+  */
+case class GenericDauConf(
+                    // required params
+                    from: String,
+                    to: String,
+                    inputBasePath: String,
+                    outputBasePath: String,
+                    // optional params
+                    additionalOutputColumns: List[String] = Nil,
+                    countColumn: Option[String] = None,
+                    hllBytes: Int = 12,
+                    hllColumn: String = "hll",
+                    inputDateColumn: String = "submission_date",
+                    inputDatePattern: String = "yyyyMMdd",
+                    minDauThreshold: Option[Int] = None,
+                    outputDateColumn: String = "submission_date_s3",
+                    outputDatePattern: String = "yyyyMMdd",
+                    top: List[(String,Int)] = Nil,
+                    where: String = "TRUE"
+                  ) {
+  val PrecedingDaysInWeek: Int = 6
+  val PrecedingDaysInMonth: Int = 27
+  // parse date patterns
+  private val inputDateFmt = DateTimeFormatter.ofPattern(inputDatePattern)
+  private val outputDateFmt = DateTimeFormatter.ofPattern(outputDatePattern)
+  // dates needed from input
+  private val inputDates: List[String] = {
+    val fromDate = LocalDate.parse(from, inputDateFmt)
+    val toDate = LocalDate.parse(to, inputDateFmt)
+    // 27 preceding days in input for calculating mau
+    (0L to DAYS.between(fromDate.minusDays(PrecedingDaysInMonth), toDate))
+      .map(toDate.minusDays(_).format(inputDateFmt)).toList
+  }
+  // convert inputDates to paths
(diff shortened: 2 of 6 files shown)


src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      "top-limit",
+      default = Some(100),
+      descr = "restrict input to top TOP_LIMIT values by count for TOP_COLUMN",
+      required = false)
+    val where = opt[String](
+      "where",
+      descr = "where clause to restrict input",
+      required = false)
+
+    verify()
```
sunahsuh: There are a few argument validations that should be here -- for instance, there should probably be a `dependsOnAll` for top-column/top-limit, and you could do a bit of custom validation to make sure `top-column` appears the `columns` list and avoid the append in the job code (which I think would be cleaner -- seems like if `top-column` doesn't show up in `columns`, that's likely a mistake)

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      descr = "name of submission date partition column in INPUT",
+      required = false)
+    val inputDatePattern = opt[String](
+      "input-date-pattern",
+      default = Some("yyyyMMdd"),
+      descr = "date pattern for BACKFILL_DATE, DATE, and INPUT_DATE_COLUMN",
+      required = false)
+    val outlierThreshold = opt[Int](
+      "outlier-threshold",
+      descr = "drop rows with a count lower than this value",
```
sunahsuh: Might be helpful to explicitly add the word "exclusive" in the descr

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      // handle --columns
+      .groupBy(inputDateColumn, columns:_*)
+      // compute daily hlls
+      .agg(expr(s"hll_merge($hllColumn)").as(hllColumn))
+      // handle --outlier-threshold
+      .where(s"hll_cardinality($hllColumn)>=${conf.outlierThreshold.getOrElse(0)}")
+      // add temporary integer date column as days since epoch
+      .withColumn(
+        s"int_$inputDateColumn",
+        floor(unix_timestamp(to_date(col(inputDateColumn), conf.inputDatePattern()))/SecondsPerDay))
```
sunahsuh: Seems a bit convoluted to turn this into "days since epoch" -- why not keep this as a date and let the sql engine do the date math?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+    // window for calculating smoothed_dau
+    val smoothedDauWindow = baseWindow.rangeBetween(-PrecedingDaysInWeek, Window.currentRow)
+
+    sample
+      // handle --columns
+      .groupBy(inputDateColumn, columns:_*)
+      // compute daily hlls
+      .agg(expr(s"hll_merge($hllColumn)").as(hllColumn))
+      // handle --outlier-threshold
+      .where(s"hll_cardinality($hllColumn)>=${conf.outlierThreshold.getOrElse(0)}")
```
sunahsuh: Oof, if the outlierThreshold is 0, this is pretty expensive for a no-op. Consider only doing this if outlierThreshold is present

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      // add temporary integer date column as days since epoch
+      .withColumn(
+        s"int_$inputDateColumn",
+        floor(unix_timestamp(to_date(col(inputDateColumn), conf.inputDatePattern()))/SecondsPerDay))
+      // calculate mau hll
+      .withColumn("mau", expr(s"hll_merge($hllColumn)") over mauWindow)
```
sunahsuh: did you mean to name this `mau_hll`?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      // calculate cardinality separately to avoid AnalysisException
+      .withColumn("mau",  expr("hll_cardinality(mau)"))
+      // calculate dau
+      .withColumn("dau", expr(s"hll_cardinality($hllColumn)"))
+      // drop hll column
+      .drop(hllColumn)
```
sunahsuh: Is this not something we want to keep? TBH, if we're not keeping `hllColumn` in the output (and therefore need it to be compatible with the Presto UDF) we should strongly prefer the spark hll implementation instead

src/test/scala/com/mozilla/telemetry/views/GenericDauViewTest.scala
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.spark.sql.DataFrame
+import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
+
+class GenericDauViewTest extends FlatSpec with Matchers with DataFrameSuiteBase with BeforeAndAfterAll {
+  private val tempDir = createTempDir().toString
+  private val hllCols = "dau" :: "mau" :: "smoothed_dau" :: "er" :: Nil
+  def test(raw: DataFrame, expect: DataFrame, args: Array[String]): Unit = {
+    val conf = new GenericDauView.Conf(args)
+
+    val input = if (conf.countColumn.isSupplied) {
```
sunahsuh: This looks like logic repeated from the main job -- can we make these bits callable from test too?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      "input",
+      descr = "location of input dataset",
+      required = true)
+    val output = opt[String](
+      "output",
+      descr = "location output dataset",
+      required = true)
+
+    // optional
+    val backfill = opt[String](
```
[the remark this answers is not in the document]
relud: Explanation: I had previously named it `backfill-date`, and I felt this was closer in line with `--date` as observed in `ExperimentAnalysisView` and `ClientsDailyView`. That said, those are the exception here, and I 100% agree that we should stick with the norm. I'll change this to use `--from` and `--to`

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      "backfill",
+      descr = "generate counts from this date through DATE",
+      required = false)
+    val columns = opt[String](
+      "columns",
+      descr = "calculate dau for these comma separated columns",
```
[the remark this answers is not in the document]
relud: perhaps `--additional-output-columns` ?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      default = Some("submission_date"),
+      descr = "name of submission date partition column in INPUT",
+      required = false)
+    val inputDatePattern = opt[String](
+      "input-date-pattern",
+      default = Some("yyyyMMdd"),
+      descr = "date pattern for BACKFILL_DATE, DATE, and INPUT_DATE_COLUMN",
+      required = false)
+    val outlierThreshold = opt[Int](
+      "outlier-threshold",
```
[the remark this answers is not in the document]
relud: how about `min-dau-threshold`

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: i'm going to `s/count/dau/` but I can't think of a good phrasing for this using "exclusive". got any suggestions?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      required = false)
+    val topLimit = opt[Int](
+      "top-limit",
+      default = Some(100),
+      descr = "restrict input to top TOP_LIMIT values by count for TOP_COLUMN",
+      required = false)
+    val where = opt[String](
+      "where",
+      descr = "where clause to restrict input",
+      required = false)
```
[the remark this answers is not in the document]
relud: i'll add an `hll-bits` flag defaulting to 12

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      default = Some("submission_date_s3"),
+      descr = "name of submission date partition column in OUTPUT",
+      required = false)
+    val outputDatePattern = opt[String](
+      "output-date-pattern",
+      default = Some("yyyyMMdd"),
+      descr = "date pattern for OUTPUT_DATE_COLUMN",
+      required = false)
+    val topColumn = opt[String](
+      "top-column",
```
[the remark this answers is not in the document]
relud: this has implementation cost, I will file a follow-up bug.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: I will punt on all of this to the bug for supporting multiple `top-column`s. I'll make the top-columns require inclusion in columns with that.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+          .agg(expr(s"hll_merge(hll_create($column, 12))").as(hllColumn))
+      case _ => df_with_where
+    }
+
+    // handle --top-column and --top-limit
+    val sample = (conf.topColumn.toOption, conf.topLimit.toOption) match {
+      case (Some(column), Some(limit)) => df_with_hll.join(
+        df_with_hll
+          .groupBy(column)
+          .agg(expr(s"hll_cardinality(hll_merge($hllColumn))").as("count"))
```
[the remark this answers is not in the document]
relud: perhaps, but when reading input from `client_count_daily` (the original use case for this job) we can't use `approx_distinct` because we have already thrown out the original column. If this job replaces `client_count_daily` completely, then I'd be willing to switch to that approach.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+    // window for calculating mau
+    val mauWindow = baseWindow.rangeBetween(-PrecedingDaysInMonth, Window.currentRow)
+    // window for calculating smoothed_dau
+    val smoothedDauWindow = baseWindow.rangeBetween(-PrecedingDaysInWeek, Window.currentRow)
+
+    sample
+      // handle --columns
+      .groupBy(inputDateColumn, columns:_*)
+      // compute daily hlls
+      .agg(expr(s"hll_merge($hllColumn)").as(hllColumn))
```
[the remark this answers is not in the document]
relud: if `--count-column` isn't specified, then `--hll-column` is expected to already be present in the dataset (ie `client_count_daily`)

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: will do. does it suffice to specify `.where("TRUE")` in the noop case, or is that also expensive?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: I'll double check, but I was under the impression that `Window.rangeBetween` requires an integer column.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: no, I avoided using column names outside of `"mau"`, `"dau"`, `"smoothed_dau"`, `countColumn`, `hllColumn`, and `inputDateColumn` throughout (and `"int_$inputDateColumn"` because I need both), to avoid corner cases where someone has a column naming collision.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      descr = "calculate dau for these comma separated columns",
+      required = false)
+    val countColumn = opt[String](
+      "count-column",
+      descr = "optional column to count distinct values of, instead of HLL_COLUMN",
+      required = false)
+    val hllColumn = opt[String](
+      "hll-column",
+      default = Some("hll"),
+      descr = "name of the hll type column to use for generating counts",
```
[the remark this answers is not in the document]
relud: will fix

src/test/scala/com/mozilla/telemetry/views/GenericDauViewTest.scala
relud: I'll move it to a separate function and expose it

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: we need to be able to read from `client_count_daily`. if we eliminate `client_count_daily`, then this sounds like a good idea, but I can't find any examples of using spark's hyperloglog that expose the hll value, which is needed so that it can be reused between `dau` and `mau`.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,219 @@
...
+      required = false)
+    val inputDateColumn = opt[String](
+      "input-date-column",
+      default = Some("submission_date"),
+      descr = "name of submission date partition column in INPUT",
+      required = false)
+    val inputDatePattern = opt[String](
+      "input-date-pattern",
+      default = Some("yyyyMMdd"),
+      descr = "date pattern for BACKFILL_DATE, DATE, and INPUT_DATE_COLUMN",
```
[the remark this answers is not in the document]
relud: will fix

relud:
> there's a lot of business logic being passed in as arguments, and permutations of these arguments live in an entirely separate repo (telemetry-airflow). This would probably be a lot cleaner and easier to maintain if we implemented this as interfaces and composed specific jobs out of them. I don't think it's any more overhead than trying to get the arguments correct in the airflow repo, and the number of derived jobs off of this is never going to be huge.

I have 16 jobs coming off of this currently planned. Part of me agrees with you, and part of me thinks it will be a lot of boiler plate. @mreid-moz could your provide a your opinion here. Maybe we could add a separate class where the common arguments are mapped behind a single argument like `--named-query=esr_dau`

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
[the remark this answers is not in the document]
sunahsuh: That should definitely be a validated argument clause then :)

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
sunahsuh: I was thinking something like `"drop rows with a dau lower than this value, exclusive"` -- I'm okay skipping it if it doesn't make sense, since "lower" does imply it's exclusive, I just feel like maybe there's a little room for ambiguity there.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
sunahsuh: Ahh, I see -- but, I think because you're partitioning by all the facet columns anyway, you can use row-based windows

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
sunahsuh: Ah, sorry, I should have been more explicit -- the next clause overwrites this column, and I wanted to make sure that's intentional

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
sunahsuh: Yes, so I have a sneaking suspicion that reuse doesn't really save us that much (although I could very much be wrong)

Anyway, I'm a bit concerned re: dropping the hll column since I've definitely seen queries where we merge hlls between facet columns. For instance, if we have an os and an app_version column, we end up with a row per combination but sometimes you want to know, say, the mau for one OS across all versions and clients switch between app versions so just adding mau wouldn't work, but merging the HLLs would prevent double-counting.

Sounds like client_count_daily is causing multiple complications to this job. What's the benefit of creating this dataset off of that job vs clients_daily?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: if i use a row-based window then it doesn't properly handle missing days, which occasionally occur

mreid-moz:
>> there's a lot of business logic being passed in as arguments, and permutations of these arguments live in an entirely separate repo (telemetry-airflow). This would probably be a lot cleaner and easier to maintain if we implemented this as interfaces and composed specific jobs out of them. I don't think it's any more overhead than trying to get the arguments correct in the airflow repo, and the number of derived jobs off of this is never going to be huge.

> I have 16 jobs coming off of this currently planned. Part of me agrees with you, and part of me thinks it will be a lot of boiler plate. @mreid-moz could your provide a your opinion here. Maybe we could add a separate class where the common arguments are mapped behind a single argument like --named-query=esr_dau

I generally prefer for logic to be in the code rather than in the config, and in my experience it has been somewhat problematic to try to co-ordinate changes across multiple repositories. The other consideration I would make is towards testability - it'll be hard to test the logic if it lives in the airflow job, but should be somewhat easier if we can write unit tests for it in TBV.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: yes, that's intentional

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: okay, updated this to indicate that it's inclusive

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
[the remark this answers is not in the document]
relud: sticking with this method because we are preserving `hllColumn` for later use

relud:
Okay, I went with named permutations I think that meets mark's desire for test-ability and sunah's desire for an interface with composed jobs.

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,211 @@
...
+    val minDauThresholdWhere: String = minDauThreshold
+      .map(t=>s"hll_cardinality($hllColumn)>=$t")
+      .getOrElse("TRUE")
+    // convert countColumn to hll_create expression or hllColumn
+    val hllColumnExpression: String = countColumn
+      .map(c=>s"hll_create($c, $hllBytes)")
+      .getOrElse(hllColumn)
+  }
+
+  class Conf(args: Array[String]) extends ScallopConf(args) {
```
sunahsuh: Can we change the class name here to something like `CommonConf` or somesuch? The `conf` value in the main job later would suggest it's an instance of this class but it's actually the `ExpandedConf` class

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,211 @@
...
+      default = Some("s3://telemetry-parquet"),
+      descr = "location where input and output data sets are stored",
+      required = false)
+
+    verify()
+
+    // default for --from is the value of --to
+    private val fromString = from.getOrElse(to())
+    val expanded: ExpandedConf = tableName() match {
+      case "dau" => ExpandedConf(
```
sunahsuh: This should throw a non-exhaustive match warning. What do you want to see happen in all the other cases?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
```diff
@@ -0,0 +1,211 @@
...
+        fromString,
+        to(),
+        inputBasePath = s"${bucket()}/client_count_daily/v2",
+        outputBasePath = s"${bucket()}/dau/v1"
+      )
+    }
+  }
+
+  def main(args: Array[String]) {
+    aggregate(getOrCreateSparkSession(jobName), new Conf(args).expanded)
```
sunahsuh: My strong preference would be for this job to be a trait with an abstract `main` method, and `ClientCountDailyView` would be an object just a few lines long that implements `main` basically using the code in `Conf.expanded`

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: sure, how about `CliConf`?

src/main/scala/com/mozilla/telemetry/views/GenericDauView.scala
relud: oops, missed handling that. i'll make it respond like scallop does for an unknown flag

relud:
changed it to a trait, but not sure if it's what you had in mind

PR #452, opened 2018-07-12
Bug 1475265 Warn on high rate of experiment branch switching

jklukas:


jklukas:
Log output looks like:

```
18/07/12 11:03:57 INFO ExperimentEngagementAnalyzer$: Pruning 1 out of 5 clients from experiment 'experiment_id' because they appear in more than one branch
18/07/12 11:03:57 WARN ExperimentEngagementAnalyzer$: The fraction of clients appearing in more than one branch of experiment 'experiment_id' is suspiciously high; there might be a problem with the experiment deployment or data collection
```

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -82,22 +82,50 @@ object ExperimentEngagementAnalyzer {
 
 
   private def filterOutliersAndAggregatePerClientDaily(
-      input: DataFrame, outlierPercentile: Double = 0.9999, relativeError: Double = 0.0001): DataFrame = {
+      input: DataFrame,
+      outlierPercentile: Double = 0.9999,
+      relativeError: Double = 0.0001,
+      branchSwitcherFractionThreshold: Double = 0.05): DataFrame = {
```
jklukas: Using 5% as the threshold here. Open to discussion on how to set this value.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -111,8 +139,7 @@ object ExperimentEngagementAnalyzer {
         sum("scalar_parent_browser_engagement_total_uri_count")
           .alias(EngagementMeasures.hourlyUris)
       )
-
-    dailyWithOutliers.persist()
+      .persist()
```
jklukas: No functionality change in this line, just going with the fluent style for calling `persist` used elsewhere in this code.

jklukas:
cc @mreid-moz who suggested this change

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
fbertsch: The basis for this value should probably be empirically derived, can you create a notebook to check it for a few different experiments?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -82,22 +82,50 @@ object ExperimentEngagementAnalyzer {
...
 
   private def filterOutliersAndAggregatePerClientDaily(
-      input: DataFrame, outlierPercentile: Double = 0.9999, relativeError: Double = 0.0001): DataFrame = {
+      input: DataFrame,
+      outlierPercentile: Double = 0.9999,
+      relativeError: Double = 0.0001,
+      branchSwitcherFractionThreshold: Double = 0.05): DataFrame = {
 
-    val clientsSwitchingBranches = input
+    val branchesPerClientPerExperiment = input
```
fbertsch: I would prefer a simpler expression of this statement, something like:
```
input.groupBy("experiment_id", "client_id").agg(approx_count_distinct("experiment_branch"))
```

(or an exact count), unless we've found a significant perf gain from the longer expression.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -82,22 +82,50 @@ object ExperimentEngagementAnalyzer {
...
+        )
+        .foreach { row =>
+          val experimentId = row.getString(0)
+          val nClients = row.getLong(1)
+          val nSwitchers = row.getLong(2)
+          val switcherFraction = nSwitchers.toDouble / nClients.toDouble
+          logger.info(s"Pruning $nSwitchers out of $nClients clients from experiment '$experimentId' " +
+            "because they appear in more than one branch")
+          if (switcherFraction > branchSwitcherFractionThreshold) {
+            logger.warn("The fraction of clients appearing in more than one branch " +
```
fbertsch: Have you confirmed:
1. That this works on a distributed cluster (the logger may need to be `@transient`, since `org.slf4j.Logger` doesn't implement `serializable` directly)
2. That the warning shows up in the driver logs

jklukas:
I rate your review 💯. You raise some great fundamental questions about whether we're doing the right thing here. The proposed changes here increase the scope of work, so I might put this work on hold for a bit.

jklukas:
Closing as this is quite out of date now. There's a bug to track further work on this.

PR #453, opened 2018-07-13
Bug 1474703 Add retention metrics to experiments_aggregates

jklukas:
This adds retention metrics and refactors the analyzer to use fewer bare strings and instead rely on enumerations of column definitions.

I've run this code on recent data via EMR. See retention metric example at the bottom of:

https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/19306

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,8 @@ val localMaven = "s3://net-mozaws-data-us-west-2-ops-mavenrepo/"
 
 resolvers += "S3 local maven snapshots" at localMavenHttps + "snapshots"
 
-val sparkVersion = "2.3.0"
+// Bumped from 2.3.0 to 2.3.1 for bugfix: https://issues.apache.org/jira/browse/SPARK-23986
+val sparkVersion = "2.3.1"
 // Should keep hadoop version in sync with the dependency defined by Spark.
 val hadoopVersion = "2.6.5"
 
--- a/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
+++ b/src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
@@ -3,38 +3,151 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.experiments.analyzers
 
-import breeze.stats.MeanAndVariance
 import breeze.stats.distributions.Poisson
-import com.mozilla.telemetry.experiments.statistics.ZScore99
-import org.apache.spark.sql.DataFrame
+import com.mozilla.telemetry.utils.ColumnEnumeration
+import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{DoubleType, IntegerType}
+import org.apache.spark.sql.{Column, DataFrame}
 
 /**
-  * See https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#definitions
+  * See discussion of confidence levels in
+  * https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#the-shift-plot
   */
-object EngagementMeasures {
-  val dailyHours: String = "engagement_daily_hours"
-  val dailyActiveHours: String = "engagement_daily_active_hours"
-  val hourlyUris: String = "engagement_hourly_uris"
-  val intensity: String = "engagement_intensity"
-  val base: List[String] = List(dailyHours, dailyActiveHours, hourlyUris)
-  val derived: List[String] = List(intensity)
-  val all: List[String] = base ++ derived
+object ConfidenceInterval {
+  val confidenceLevel:  Double = 0.975
+  val confidenceMargin: Double = 0.5 * (1.0 - confidenceLevel)
+  val percentileLow:    Double = 0.0 + confidenceMargin
+  val percentileHigh:   Double = 1.0 - confidenceMargin
 }
 
-object Percentiles {
-  val ints: Seq[Int] = (1 to 9).map(_ * 10)
-  val decimals: Seq[Double] = ints.map(_ * 0.01)
-  val names: Seq[String] = ints.map {
+case class ConfidenceInterval(arr: Array[Double]) {
+  def low:  Double = breeze.stats.DescriptiveStats.percentile(arr, ConfidenceInterval.percentileLow)
+  def high: Double = breeze.stats.DescriptiveStats.percentile(arr, ConfidenceInterval.percentileHigh)
+}
+
+/**
+  * Columns we need to pull from the input DataFrame.
+  */
+object InputCols extends ColumnEnumeration {
+  val experiment_id, experiment_branch, client_id, submission_date_s3 = Val()
+  val total_time, active_ticks, scalar_parent_browser_engagement_total_uri_count = Val()
+}
+
+/**
+  * Derived columns calculated via window functions.
+  */
+object InputWindowCols extends ColumnEnumeration {
+  private val branchSet: Column =
+    collect_set(InputCols.experiment_branch.col).over(
+      Window
+        .partitionBy(InputCols.experiment_id.col, InputCols.client_id.col)
+        .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+    )
+
+  val branch_count = Val(size(branchSet))
+}
+
+/**
+  * Aggregations we apply per experiment/branch/client/day.
+  */
+object DailyAggCols extends ColumnEnumeration {
+  val partitionCols: Array[Column] = Array(
+    InputCols.experiment_id.col,
+    InputCols.experiment_branch.col,
+    InputCols.client_id.col,
+    InputCols.submission_date_s3.col
+  )
+
+  val ticksPerSecond: Double = 5.0
+  val secondsPerHour: Double = 3600.0
+
+  val sum_total_hours = Val(
+    sum(InputCols.total_time.col).cast(DoubleType) / secondsPerHour
+  )
+  val sum_active_hours = Val(
+    sum(InputCols.active_ticks.col.cast(DoubleType) * ticksPerSecond / secondsPerHour)
+  )
+  val sum_total_uris = Val(
+    sum(InputCols.scalar_parent_browser_engagement_total_uri_count.col)
+  )
+}
+
+  * Aggregations per experiment/branch/client, giving a summary per client
+  * of engagement over all completed days of the experiment.
+  */
+object EngagementAggCols extends ColumnEnumeration {
+  val partitionCols: Array[Column] = Array(
+    InputCols.experiment_id.col,
+    InputCols.experiment_branch.col,
+    InputCols.client_id.col
+  )
+
+  val engagement_daily_active_hours = Val(avg(DailyAggCols.sum_active_hours.col))
+  val engagement_hourly_uris = Val(
+    sum(DailyAggCols.sum_total_uris.col) /
+      (sum(DailyAggCols.sum_active_hours.col) + 1.0 / DailyAggCols.secondsPerHour)
+  )
+  val engagement_intensity = Val(
+    sum(DailyAggCols.sum_active_hours.col) /
+      (sum(DailyAggCols.sum_total_hours.col) + 1.0 / DailyAggCols.secondsPerHour)
+  )
+}
+
+/**
+  * Object capturing the logic for naming and computing a particular percentile.
+  *
+  * @param pInt The desired percentile as an integer between 0 and 100
+  */
+case class PercentileComputation(pInt: Int) extends StatisticalComputation {
+  val p: Double = pInt * 0.01
+
+  override def name: String = pInt match {
     case 50 => "Median"
-    case p => s"${p}th Percentile"
+    case _ => s"${pInt}th Percentile"
+  }
+
+  /**
+    * Based on https://github.com/scalanlp/breeze/blob/releases/v1.0-RC2/math/src/main/scala/breeze/stats/DescriptiveStats.scala#L523
+    *
+    * The array must already be sorted.
+    */
+  override def calculate(arr: Array[Double]): Double = {
+      if (p > 1 || p < 0) throw new IllegalArgumentException("p must be in [0,1]")
+      // +1 so that the .5 == mean for even number of elements.
+      val f = (arr.length + 1) * p
+      val i = f.toInt
+      if (i == 0) {
+        arr.head
+      }
+      else if (i >= arr.length) {
+        arr.last
+      }
+      else {
+        arr(i - 1) + (f - i) * (arr(i) - arr(i - 1))
+      }
+    }
+}
+
+object MeanComputation extends StatisticalComputation {
+  override val name: String = "Mean"
+  override def calculate(arr: Array[Double]): Double = {
+    breeze.stats.mean(arr)
   }
 }
 
-object TimeConstants {
-  val ticksPerSecond: Double = 5.0
-  val secondsPerHour: Double = 3600.0
+object StatisticalComputation {
+  val percentileInts: Seq[Int] = 5 +: (10 to 90 by 10) :+ 95
+  val percentiles: Seq[PercentileComputation] = percentileInts.map(PercentileComputation)
+  val values: Seq[StatisticalComputation] = percentiles :+ MeanComputation
+  val names: Seq[String] = values.map(_.name)
 }
+
+trait StatisticalComputation {
+  def name: String
+  def calculate(arr: Array[Double]): Double
+}
+
 
 /**
   * Introduced for bug 1460090, based on a proof-of-concept analysis in
@@ -58,13 +171,12 @@ object ExperimentEngagementAnalyzer {
     */
   def getMetrics(experimentsDF: DataFrame, iterations: Int): Seq[MetricAnalysis] = {
 
-    val clients = filterOutliersAndAggregatePerClientDaily(experimentsDF)
-    clients.persist()
+    val clients = filterOutliersAndAggregatePerClientDaily(experimentsDF).persist()
 
     logger.info(s"Calculating engagement metrics for ${clients.count()} client/experiment/branch rows")
 
     val branches = clients
-      .select("experiment_id", "experiment_branch")
+      .select(InputCols.experiment_id.col, InputCols.experiment_branch.col)
       .distinct()
       .collect()
       .map { row => (row.getString(0), row.getString(1)) }
@@ -72,7 +184,7 @@ object ExperimentEngagementAnalyzer {
     val metricAnalyses =
       for {
         (experimentId, branch) <- branches
-        measure <- EngagementMeasures.all
+        measure <- EngagementAggCols.names
       } yield calculateMetricAnalysis(clients, experimentId, branch, measure, iterations)
 
     clients.unpersist()
@@ -84,59 +196,44 @@ object ExperimentEngagementAnalyzer {
   private def filterOutliersAndAggregatePerClientDaily(
       input: DataFrame, outlierPercentile: Double = 0.9999, relativeError: Double = 0.0001): DataFrame = {
 
-    val clientsSwitchingBranches = input
-      .select("experiment_id", "experiment_branch", "client_id")
-      .distinct()
-      .groupBy("experiment_id", "client_id")
-      .count()
-      .filter(col("count") > 1)
-      .drop(col("count"))
+    val inputWithBranchCount = input
+      .select(col("*"), InputWindowCols.branch_count.expr)
       .persist()
 
-    logger.info(s"Pruning ${clientsSwitchingBranches.count()} " +
+    val numClientsSwitchingBranches = inputWithBranchCount
+      .filter(InputWindowCols.branch_count.col > 1)
+      .select(InputCols.experiment_id.col, InputCols.client_id.col)
+      .distinct()
+      .count()
+
+    logger.info(s"Pruning $numClientsSwitchingBranches " +
       "clients that appear in more than one branch per experiment")
 
-    val inputWithoutSwitchers = input
-      .join(clientsSwitchingBranches.unpersist(), Seq("experiment_id", "client_id"), joinType = "leftanti")
+    val dailyWithOutliers = inputWithBranchCount
+      .filter(InputWindowCols.branch_count.col === 1)
+      .groupBy(DailyAggCols.partitionCols: _*)
+      .agg(DailyAggCols.exprs.head, DailyAggCols.exprs.tail: _*)
+      .select(col("*"), EnrollmentWindowCols.week_number.expr)
+      .persist()
 
-    val dailyWithOutliers = inputWithoutSwitchers
-      .groupBy("experiment_id", "experiment_branch", "client_id", "submission_date_s3")
-      .agg(
-        sum(expr(s"total_time/${TimeConstants.secondsPerHour}"))
-          .cast("double")
-          .alias(EngagementMeasures.dailyHours),
-        sum(expr(s"active_ticks/(${TimeConstants.secondsPerHour}/${TimeConstants.ticksPerSecond})"))
-          .cast("double")
-          .alias(EngagementMeasures.dailyActiveHours),
-        sum("scalar_parent_browser_engagement_total_uri_count")
-          .alias(EngagementMeasures.hourlyUris)
-      )
-
-    dailyWithOutliers.persist()
+    inputWithBranchCount.unpersist()
 
     // Per sguha, we expect every hypothesis test to benefit from removing outliers above the 99.99th percentile;
     // see https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#remove-outliers
     val outlierCutoffs = dailyWithOutliers.stat
-      .approxQuantile(EngagementMeasures.base.toArray, Array(outlierPercentile), relativeError)
+      .approxQuantile(DailyAggCols.names.toArray, Array(outlierPercentile), relativeError)
       .map(_.head)
 
-    val daily = (EngagementMeasures.base, outlierCutoffs)
+    val daily = (DailyAggCols.cols, outlierCutoffs)
       .zipped
       .foldLeft(dailyWithOutliers) { (df, cutInfo) =>
         val (measure, cut) = cutInfo
-        df.where(col(measure) < cut)
+        df.where(measure < cut)
       }
 
     val clients = daily
-      .groupBy("experiment_id", "experiment_branch", "client_id")
-      .agg(
-        avg(EngagementMeasures.dailyHours).alias(EngagementMeasures.dailyHours),
-        avg(EngagementMeasures.dailyActiveHours).alias(EngagementMeasures.dailyActiveHours),
-        expr(s"sum(${EngagementMeasures.hourlyUris})/(1.0 / ${TimeConstants.secondsPerHour} + sum(${EngagementMeasures.dailyActiveHours}))")
-          .cast("double").alias(EngagementMeasures.hourlyUris),
-        expr(s"sum(${EngagementMeasures.dailyActiveHours})/(1.0 / ${TimeConstants.secondsPerHour} + sum(${EngagementMeasures.dailyHours}))")
-          .cast("double").alias(EngagementMeasures.intensity)
-      )
+      .groupBy(EngagementAggCols.partitionCols: _*)
+      .agg(EngagementAggCols.exprs.head, EngagementAggCols.exprs.tail: _*)
       .na.fill(0)
 
     dailyWithOutliers.unpersist()
@@ -190,6 +287,8 @@ object ExperimentEngagementAnalyzer {
     val count = observations.length
     val observationsBroadcast = sc.broadcast(observations)
 
+    val fullSampleComputations = StatisticalComputation.values.map(_.calculate(observations)).toArray
+
     // The "Poisson bootstrap" is a statistical technique where we resample the dataset many times,
     // calculating percentiles on each sample, then use the distribution of each percentile across
     // all samples in order to estimate confidence in each percentile value.
@@ -199,7 +298,7 @@ object ExperimentEngagementAnalyzer {
     // and returns an array giving one value per calculated percentile.
     //
     // See: https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#bootstrapping
-    val percentileArrays: Array[Array[Double]] = sc.parallelize(1 to iterations).flatMap { _ =>
+    val bootstrapComputations: Array[Array[Double]] = sc.parallelize(1 to iterations).flatMap { _ =>
 
       // For each observation, we draw a random poisson value (usually 0 or 1 with a tail of larger integers)
       // and duplicate the observation that many times.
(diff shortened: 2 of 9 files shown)


jklukas:
I've updated https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/19306 to show these new retention metrics.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{DoubleType, IntegerType}
+import org.apache.spark.sql.{Column, DataFrame}
 
 /**
-  * See https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#definitions
+  * See discussion of confidence levels in
+  * https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#the-shift-plot
   */
-object EngagementMeasures {
```
jklukas: As part of the refactor, we eliminate most of the old `object`s that held constants and replace them with `ColumnEnumeration`s.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
-  val dailyHours: String = "engagement_daily_hours"
-  val dailyActiveHours: String = "engagement_daily_active_hours"
-  val hourlyUris: String = "engagement_hourly_uris"
-  val intensity: String = "engagement_intensity"
-  val base: List[String] = List(dailyHours, dailyActiveHours, hourlyUris)
-  val derived: List[String] = List(intensity)
-  val all: List[String] = base ++ derived
+object ConfidenceInterval {
```
jklukas: This code previously borrowed Z score methodology that we used in jackknife method, but we now switch to the confidence interval method discussed by sguha.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+  val confidenceLevel:  Double = 0.975
+  val confidenceMargin: Double = 0.5 * (1.0 - confidenceLevel)
+  val percentileLow:    Double = 0.0 + confidenceMargin
+  val percentileHigh:   Double = 1.0 - confidenceMargin
 }
 
-object Percentiles {
```
jklukas: These Percentiles have been moved into the PercentileComputation case class, where the code for calculating the percentile now lives as well.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+      else if (i >= arr.length) {
+        arr.last
+      }
+      else {
+        arr(i - 1) + (f - i) * (arr(i) - arr(i - 1))
+      }
+    }
+}
+
+object MeanComputation extends StatisticalComputation {
```
jklukas: We add computation of the mean to support a more interesting aggregate for retention metrics where percentiles will always be simply 0 or 1.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -58,21 +171,20 @@ object ExperimentEngagementAnalyzer {
...
 
-    val clients = filterOutliersAndAggregatePerClientDaily(experimentsDF)
-    clients.persist()
+    val clients = filterOutliersAndAggregatePerClientDaily(experimentsDF).persist()
 
     logger.info(s"Calculating engagement metrics for ${clients.count()} client/experiment/branch rows")
 
     val branches = clients
-      .select("experiment_id", "experiment_branch")
+      .select(InputCols.experiment_id.col, InputCols.experiment_branch.col)
```
jklukas: The `ColumnEnumerations` now let us refer to columns in code rather than as strings, so you'll see a number of revisions like this in this PR.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -84,59 +196,44 @@ object ExperimentEngagementAnalyzer {
   private def filterOutliersAndAggregatePerClientDaily(
       input: DataFrame, outlierPercentile: Double = 0.9999, relativeError: Double = 0.0001): DataFrame = {
 
-    val clientsSwitchingBranches = input
-      .select("experiment_id", "experiment_branch", "client_id")
+    val inputWithBranchCount = input
+      .select(col("*"), InputWindowCols.branch_count.expr)
+      .persist()
```
jklukas: We've refactored this branch switching logic to use a window function rather than aggregating and joining back to the dataframe. This is hopefully simpler to understand and maintain.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -84,59 +196,44 @@ object ExperimentEngagementAnalyzer {
...
-      .agg(
-        avg(EngagementMeasures.dailyHours).alias(EngagementMeasures.dailyHours),
-        avg(EngagementMeasures.dailyActiveHours).alias(EngagementMeasures.dailyActiveHours),
-        expr(s"sum(${EngagementMeasures.hourlyUris})/(1.0 / ${TimeConstants.secondsPerHour} + sum(${EngagementMeasures.dailyActiveHours}))")
-          .cast("double").alias(EngagementMeasures.hourlyUris),
-        expr(s"sum(${EngagementMeasures.dailyActiveHours})/(1.0 / ${TimeConstants.secondsPerHour} + sum(${EngagementMeasures.dailyHours}))")
-          .cast("double").alias(EngagementMeasures.intensity)
-      )
+      .groupBy(EngagementAggCols.partitionCols: _*)
+      .agg(EngagementAggCols.exprs.head, EngagementAggCols.exprs.tail: _*)
```
jklukas: The aggregation logic has all been moved into `EngagementAggCols`, but is equivalent to the expressions we're replacing here.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -214,44 +313,23 @@ object ExperimentEngagementAnalyzer {
...
-        Statistic(None, nameForPercentile, stats.mean,
-          Some(stats.mean - z99), Some(stats.mean + z99),
-          None, None)
-      }
+    val statsArray: Array[Statistic] =
+      (fullSampleComputations, bootstrapComputations.transpose, StatisticalComputation.names)
+        .zipped
+        .map { case (fullSampleValue, bootstrapValues, nameForComputation) =>
+          val confidence = ConfidenceInterval(bootstrapValues)
+          Statistic(None, nameForComputation, fullSampleValue, Some(confidence.low), Some(confidence.high), None, None)
```
jklukas: This is entirely new logic to match what was in sguha's proposal.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -214,44 +313,23 @@ object ExperimentEngagementAnalyzer {
...
     MetricAnalysis(experimentId, branch, "", count, measure, "DoubleScalar",
       Map.empty[Long, HistogramPoint],
       Some(statsArray))
   }
 
-
-  /**
-    * Based on https://github.com/scalanlp/breeze/blob/releases/v1.0-RC2/math/src/main/scala/breeze/stats/DescriptiveStats.scala#L523
-    */
-  private def percentileFromSortedArray(arr: Array[Double], p: Double): Double = {
```
jklukas: This logic has been moved into the PercentileComputation case class

build.sbt
```diff
@@ -11,7 +11,8 @@ val localMaven = "s3://net-mozaws-data-us-west-2-ops-mavenrepo/"
 
 resolvers += "S3 local maven snapshots" at localMavenHttps + "snapshots"
 
-val sparkVersion = "2.3.0"
+// Bumped from 2.3.0 to 2.3.1 for bugfix: https://issues.apache.org/jira/browse/SPARK-23986
+val sparkVersion = "2.3.1"
```
jklukas: I've run this on EMR 5.13, and it appears that the version included in the uberjar truly does get picked up over what's installed on the nodes, so upgrading here does what we need to bypass this bug.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
sunahsuh: This matches what I had in the previous bootstrap implementation. The jackknife and bootstrap are related but use slightly different math (and intuition) to calculate the confidence interval.

In the future we may want to look into some of the bias corrected bootstrap versions (the BCa algorithm, for instance)

src/main/scala/com/mozilla/telemetry/utils/ColumnEnumeration.scala
```diff
@@ -0,0 +1,58 @@
...
+  * Enumerations are convenient here because they allow us to define a column
+  * name as a val, and then get access to that Scala-level name as a string to
+  * pass into Spark SQL expressions.
+  *
+  * You can also define a derived column by passing a SQL expression string into Val();
+  * this logical definition is then available as `expr`.
+  */
+abstract class ColumnEnumeration extends Enumeration {
+
+  protected case class Val(private val definition: Column = new Column("")) extends super.Val {
```
sunahsuh: I'm fine if you leave this as-is, but 1. I think if you're really concerned about type safety the typed Dataset option is the happy path, but I do understand, with a bunch of aggregations, why you'd avoid it here, but 2. I think Scala Enumerations are meh, and they don't actually achieve much in the way of type safety (*all* Enumeration are collapsed to a single type 😱), and it doesn't look like you're using most of the advantages besides being able to iterate over all the values (e.g. built-in ordering, etc). I'm eyeing this pattern with wariness, is what I guess I'm saying :)

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+  )
+  val sum_active_hours = Val(
+    sum(InputCols.active_ticks.col.cast(DoubleType) * ticksPerSecond / secondsPerHour)
+  )
+  val sum_total_uris = Val(
+    sum(InputCols.scalar_parent_browser_engagement_total_uri_count.col)
+  )
+}
+
+  * Aggregations per experiment/branch/client, giving a summary per client
```
sunahsuh: Looks like the start of this comment block is missing, which seems like it shouldn't compile? 🤔

sunahsuh: Ah! Looks like an artifact from splitting the commits!

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
-object EngagementMeasures {
-  val dailyHours: String = "engagement_daily_hours"
-  val dailyActiveHours: String = "engagement_daily_active_hours"
-  val hourlyUris: String = "engagement_hourly_uris"
-  val intensity: String = "engagement_intensity"
-  val base: List[String] = List(dailyHours, dailyActiveHours, hourlyUris)
-  val derived: List[String] = List(intensity)
-  val all: List[String] = base ++ derived
+object ConfidenceInterval {
+  val confidenceLevel:  Double = 0.975
```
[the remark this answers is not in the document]
jklukas: 97.5% was specified in the doc. I don't have an opinion one way or the other.

> I choose the 97.5% percentile because the true coverage is likely less than that. Best to be conservative.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+}
+
+/**
+  * Derived columns calculated via window functions.
+  */
+object InputWindowCols extends ColumnEnumeration {
+  private val branchSet: Column =
+    collect_set(InputCols.experiment_branch.col).over(
+      Window
+        .partitionBy(InputCols.experiment_id.col, InputCols.client_id.col)
```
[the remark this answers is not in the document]
jklukas: The intent of `partitionCols` is for the enumerations that represent a set of aggregations, and the partitionCols tells what columns these aggregations will be over.

For InputCols, some of the columns are partitions of the Parquet dataset, but the intent of that enumeration is not about aggregation.

Does that make sense? Is there a way we can make that intent more clear?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+  * Object capturing the logic for naming and computing a particular percentile.
+  *
+  * @param pInt The desired percentile as an integer between 0 and 100
+  */
+case class PercentileComputation(pInt: Int) extends StatisticalComputation {
+  val p: Double = pInt * 0.01
+
+  override def name: String = pInt match {
+    case 50 => "Median"
+    case _ => s"${pInt}th Percentile"
```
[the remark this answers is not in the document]
jklukas: Agree this is not important, but now I want to implement something more robust because it would be fun.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
+  }
+
+  /**
+    * Based on https://github.com/scalanlp/breeze/blob/releases/v1.0-RC2/math/src/main/scala/breeze/stats/DescriptiveStats.scala#L523
```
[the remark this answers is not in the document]
jklukas: Will add a comment. The breeze function always sorts the array. Since we're calculating ~10 percentiles from the same array, I wanted to be able to sort once. I will make that more clear in the docstring.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,39 +3,152 @@
...
 }
 
+object StatisticalComputation {
+  val percentileInts: Seq[Int] = 5 +: (10 to 90 by 10) :+ 95
+  val percentiles: Seq[PercentileComputation] = percentileInts.map(PercentileComputation)
+  val values: Seq[StatisticalComputation] = percentiles :+ MeanComputation
+  val names: Seq[String] = values.map(_.name)
+}
+
+trait StatisticalComputation {
```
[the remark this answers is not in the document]
jklukas: Will do.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -84,59 +196,44 @@ object ExperimentEngagementAnalyzer {
...
+      .persist()
 
-    dailyWithOutliers.persist()
+    inputWithBranchCount.unpersist()
 
     // Per sguha, we expect every hypothesis test to benefit from removing outliers above the 99.99th percentile;
     // see https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#remove-outliers
     val outlierCutoffs = dailyWithOutliers.stat
-      .approxQuantile(EngagementMeasures.base.toArray, Array(outlierPercentile), relativeError)
+      .approxQuantile(DailyAggCols.names.toArray, Array(outlierPercentile), relativeError)
```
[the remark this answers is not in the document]
jklukas: Sounds reasonable. Will do.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -84,59 +196,44 @@ object ExperimentEngagementAnalyzer {
...
       .map(_.head)
 
-    val daily = (EngagementMeasures.base, outlierCutoffs)
+    val daily = (DailyAggCols.cols, outlierCutoffs)
       .zipped
       .foldLeft(dailyWithOutliers) { (df, cutInfo) =>
```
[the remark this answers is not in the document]
jklukas: Does that work?! If so, it makes me so happy.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -199,7 +298,7 @@ object ExperimentEngagementAnalyzer {
     // and returns an array giving one value per calculated percentile.
     //
     // See: https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html#bootstrapping
-    val percentileArrays: Array[Array[Double]] = sc.parallelize(1 to iterations).flatMap { _ =>
+    val bootstrapComputations: Array[Array[Double]] = sc.parallelize(1 to iterations).flatMap { _ =>
```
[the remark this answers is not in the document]
jklukas: I'll move the poisson generator inside and seed with iteration number.

src/main/scala/com/mozilla/telemetry/utils/ColumnEnumeration.scala
jklukas: Yeah, type safety is probably the wrong word here.

What I'm trying to achieve here is being able to define the name of a column once as a code constant and then have that available as a string and as an iterable. Which seems to be the one thing that a Scala enumeration actually does well.

From your perspective, is it easy to understand what's going on, or does the ColumnEnumeration machinery make the code more difficult to understand?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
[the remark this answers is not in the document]
sunahsuh: Two thoughts: I just tried scala Quicksort on an array of already-sorted 200M doubles locally and it took just a second or so (sorting it initially took a little more time) so that's a datapoint re: weighing the cost/benefit of copying the implementation to avoid the sort.

Secondly, if time to sort is a concern consider sorting the initial observations array before broadcast since that would guarantee the pseudosample is already sorted, given the way the samples are drawn.

(I'm sorry, optimization is such a fun nerdsnipe.)

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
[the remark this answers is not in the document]
jklukas: Added logging. Here's the output in tests where we have a few hundred rows:
```
18/07/27 10:53:08 INFO ExperimentEngagementAnalyzer$: Outlier filter: cutting 1 rows with sum_total_hours < 1.0461111111111112
18/07/27 10:53:10 INFO ExperimentEngagementAnalyzer$: Outlier filter: cutting 1 rows with sum_active_hours < 1.4222222222222223
18/07/27 10:53:11 INFO ExperimentEngagementAnalyzer$: Outlier filter: cutting 0 rows with sum_total_uris < 20.0
```

So looks like this generally ends up cutting out the one highest outlier for each metric on small datasets.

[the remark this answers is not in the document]
sunahsuh: Interesting -- I think we'll want to keep an eye on this when the job is running on real data. I do believe you want a `>` in that log line instead of `<`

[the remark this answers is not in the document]
jklukas: Yep. Will fix.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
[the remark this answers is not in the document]
jklukas: > initial observations array before broadcast since that would guarantee the pseudosample is already sorted, given the way the samples are drawn.

Exactly! We're already doing this. The sorting only happens before broadcast, and then we're relying on the generated pseudosample being sorted due to the poisson sampling algorithm.

I'm inclined to keep the copied code and avoid the extra sorting.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -83,6 +99,34 @@ object EngagementAggCols extends ColumnEnumeration {
     InputCols.client_id.col
   )
 
+  val hourlyUrisConsideredActive: Int = 5
+
+  private def retained(weekNumber: Int): Column = {
+    val sumHoursInWeek =
+      sum(
```
sunahsuh: I'm a bit floored Spark doesn't have a boolean-or aggregation method (which I think would make this a bit simpler). Writing a UDF and making it generally available in our utils would be a nice enhancement.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
```diff
@@ -224,5 +224,15 @@ class ExperimentAnalysisViewTest extends FlatSpec with Matchers with DataFrameSu
...
+
+    val filteredRetention1Week = metrics.filter(_.metric_name == "retained_in_week_1")
+    filteredRetention1Week should have length 1
+    val retained = filteredRetention1Week.head
+    val meanStats = retained.statistics.get.filter(_.name == "Mean").head
+    meanStats.value should equal(0.75 +- 0.01)
+    meanStats.confidence_low.get should be < meanStats.value
+    meanStats.confidence_low.get should be >= 0.0
+    meanStats.confidence_high.get should be > meanStats.value
+    meanStats.confidence_high.get should be <= 1.0
```
sunahsuh: Sooo I say this fully aware that I am not exactly known for writing comprehensive tests, but we should probably have more robust tests for the retention numbers in various situations

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -73,6 +73,22 @@ object DailyAggCols extends ColumnEnumeration {
...
+object EnrollmentWindowCols extends ColumnEnumeration {
+  private val intDate: Column = InputCols.submission_date_s3.col.cast(IntegerType)
+  private val enrollmentDate: Column =
+    first(intDate).over(
+      Window
+        .partitionBy(InputCols.experiment_id.col, InputCols.experiment_branch.col, InputCols.client_id.col)
+        .orderBy(InputCols.submission_date_s3.col.asc)
+        .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+    )
+  val week_number = Val(floor((intDate - enrollmentDate) / 7))
```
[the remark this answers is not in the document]
jklukas: That's indeed kind of an embarrassing oversight. Will do.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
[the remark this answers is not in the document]
sunahsuh: @saptarshiguha What do you think?

src/main/scala/com/mozilla/telemetry/utils/ColumnEnumeration.scala
sunahsuh: I do think it's a little magicky/implicit in a way that took me a bit of time to parse – I think particularly the empty Val() invocation was initially confusing (in large part because who seriously decided to name this class a capitalized version of a fundamental language construct? -_-). This might be mitigated by renaming the Val class to something a bit more purpose-built (`ColumnDefinition`, perhaps?).

Another consideration here that I should have mentioned in an earlier review – it'd be more idiomatic to make definition an Option[Column] defaulted to `None` instead of matching on the column as string

jklukas: > it'd be more idiomatic to make definition an Option[Column] defaulted to None instead of matching on the column as string

Yeah, the magic value is smelly. I want to avoid the verbosity of requiring all the definitions to look like `Val(Some(foo))`. I will see if I can provide two separate constructors.

> This might be mitigated by renaming the Val class to something a bit more purpose-built (ColumnDefinition, perhaps?)

I like that idea.

src/test/scala/com/mozilla/telemetry/views/ExperimentAnalysisViewTest.scala
jklukas: Added a new test case to exercise 0 and 1 for retention and active retention in various weeks.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
jklukas: > Writing a UDF and making it generally available in our utils would be a nice enhancement.

Is "our utils" here talking about the utils package in this repo?

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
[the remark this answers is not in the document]
saptarshiguha: okay with me. So jackknife for CI?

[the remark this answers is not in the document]
jklukas: @saptarshiguha - I believe @sunahsuh was proposing we continue with the Poisson bootstrap method here, but use a 99% CI rather than 97.5%. Does that sound reasonable?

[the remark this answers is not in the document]
jklukas: I'm going ahead and changing to 99% CI with the bootstrap method.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
jklukas: Implemented a UDF

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,37 +3,143 @@
...
+  )
+}
+
+/**
+  * Derived enrollment info calculated via window functions.
+  */
+object EnrollmentWindowCols extends ColumnEnumeration {
+  private val date: Column = f.to_date(InputCols.submission_date_s3.col, "yyyyMMdd")
+  private val enrollmentDate: Column =
+    f.first(date).over(
```
sunahsuh: Technically I think `min` would have the same effect without the `orderBy` clause in the window, I'm fairly certain -- I'm not sure what the performance of these retention metrics looks like but if it adds significant time this is one place to start looking re: cutting down time.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,37 +3,143 @@
...
+  private def retainedActive(weekNumber: Int): Column = {
+    val wasActiveAnyDayInWeek =
+      f.max(EnrollmentWindowCols.week_number.col === weekNumber and
+            DailyAggCols.sum_total_uris.col > hourlyUrisConsideredActive)
+    f.when(wasActiveAnyDayInWeek, 1.0) otherwise 0.0
+  }
+
+  val retained_in_week_1 = ColumnDefinition(retained(1))
+  val retained_in_week_2 = ColumnDefinition(retained(2))
+  val retained_in_week_3 = ColumnDefinition(retained(3))
```
sunahsuh: From my reading of Saptarshi's report, week 3 retention actually looks at `where weeksSince=2`" -- to match with his definitions I think we need to shift these numbers by one (and think about how to handle week 0 since we don't want to count the day of first appearance)

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,37 +3,143 @@
...
+
+  val retained_active_in_week_1 = ColumnDefinition(retainedActive(1))
+  val retained_active_in_week_2 = ColumnDefinition(retainedActive(2))
+  val retained_active_in_week_3 = ColumnDefinition(retainedActive(3))
+
+  val engagement_daily_hours = ColumnDefinition(f.avg(DailyAggCols.sum_total_hours.col))
+  val engagement_daily_active_hours = ColumnDefinition(f.avg(DailyAggCols.sum_active_hours.col))
+  val engagement_hourly_uris = ColumnDefinition(
+    f.sum(DailyAggCols.sum_total_uris.col) /
+      (f.sum(DailyAggCols.sum_active_hours.col) + 1.0 / DailyAggCols.secondsPerHour)
```
sunahsuh: I mentioned this in a comment that ended up collapsed, but `1.0 / DailyAggCols.secondsPerHour` seems to be an epsilon to avoid divide-by-zero, and used in a few places in the code. I think it'd be nice (and possibly easier to follow for anyone looking at the code later) to pull that out as a constant called `epsilon`.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
jklukas: It works! I didn't realize that min could be used as a window function. Will change.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
jklukas: Good catch! You're totally right. I think I'll exclude any activity on the `enrollmentDate` for these calculations in order to give 1-week retention some meaning.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
jklukas: I remember that comment now. Sorry for losing track of it.

Thinking about this, it seems safer to explicitly provide an alternative for divide-by-zero. We already have a blanket `na.fill(0.0)` to replace null values with 0.0 before aggregating. I suspect we'd want 0 in this case too where the calculation is poorly defined.

src/main/scala/com/mozilla/telemetry/experiments/analyzers/ExperimentEngagementAnalyzer.scala
```diff
@@ -3,37 +3,148 @@
...
+  val retained_active_in_week_3 = ColumnDefinition(retainedActive(3))
+
+  val engagement_daily_hours = ColumnDefinition(f.avg(DailyAggCols.sum_total_hours.col))
+  val engagement_daily_active_hours = ColumnDefinition(f.avg(DailyAggCols.sum_active_hours.col))
+  val engagement_hourly_uris = ColumnDefinition(
+    f.sum(DailyAggCols.sum_total_uris.col) / f.sum(DailyAggCols.sum_active_hours.col)
+  )
+  val engagement_intensity = ColumnDefinition(
+    f.sum(DailyAggCols.sum_active_hours.col) / f.sum(DailyAggCols.sum_total_hours.col)
+  )
```
jklukas: Removed the "epsilon" 1.0 for these two calculations, so these are allowed to trigger division by zero, which leads to a `null` in the dataframe. We now have a test that exercises this and shows that the nulls are converted to 0.0 via the `na.fill(0.0)` call.

@saptarshiguha - Let me know if this sounds dangerous. It seems more predictable than having this blow up to an arbitrarily large value.

PR #454, opened 2018-07-17
Bug 1476038: Fix experiment summary (again)

sunahsuh:
We were seeing step failures again on this job with what looked like OOMs. It looks like a high-population experiment is the culprit. Two fixes here: ignore experiments tagged as high volume, and don't repartition by experiment_id (which is unbalanced anyway and most of the machines were probably just sitting idle most of the time) -- I'm coalescing down to 10 partitions instead which should keep all the executors busy but keep the number of files from exploding too badly.


sunahsuh:
I forgot to note -- re-ran with these changes for 7/15 and it finished in a little under 2 hours

PR #455, opened 2018-07-20
Bug 1468351

fbertsch:


src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -718,8 +723,14 @@ object LongitudinalView {
   }
   // scalastyle:on return
 
-  def getStandardHistogramSchema(schema: Schema): Schema = {
-    getElemType(schema, "fx_tab_switch_total_ms")
+  // ref: http://fdahms.com/2015/10/14/scala-and-the-transient-lazy-val-pattern/
+  // this val will only be calculated once per partition, and will not be serialized
```
sunahsuh: Oh this is cool!

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -718,8 +723,14 @@ object LongitudinalView {
   }
   // scalastyle:on return
 
-  def getStandardHistogramSchema(schema: Schema): Schema = {
-    getElemType(schema, "fx_tab_switch_total_ms")
```
sunahsuh: I think I need a bit more context on the fix -- on first glance, getStandardHistogramSchema and  standardHistogramSchema don't seem equivalent but it looks like it's being used as a drop-in replacement

sunahsuh: Wow, that's convoluted :D I see that histogram stopped recording in 57, so interesting that this didn't fail earlier.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -1036,7 +1047,8 @@ object LongitudinalView {
       Option(root.build)
     } catch {
       case e: Throwable =>
-        // Ignore buggy clients
+        // Log buggy clients errors and continue
+        logger.warn(org.apache.commons.lang.exception.ExceptionUtils.getStackTrace(e))
```
sunahsuh: I have a tiny concern about logging the entire trace for all the bad clients blowing up our logs, but we can probably ignore that until it actually happens

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -730,14 +741,14 @@ object LongitudinalView {
                                    histogramDefinitions: Map[String, HistogramDefinition]) {
     val histogramsList = payloads.map(Histograms.stripKeyedHistograms)
 
-    val uniqueKeys = histogramsList.flatMap(x => x.keys).distinct.toSet
+    val uniqueKeys = histogramsList.flatMap(x => x.keys).distinct
```
sunahsuh: Amazing what innocuous things cause this job to fail

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
fbertsch: Good point - let's keep an eye on this.

PR #457, opened 2018-07-25
Bug 1468351 - Pre-filter client pings before partitioning in Longitudinal

fbertsch:
This is currently running, I expect we'll know tomorrow whether setting the value at 5000 is small enough.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -220,13 +220,9 @@ object LongitudinalView {
   private def getOrderKey(message: Message): Option[((String, String, Int), Int)] = {
     // This replicate the existing code, except returns a small data structure rather
     // than the entire ping, and doesn't filter in the for loop
+    // Replicated code is in the `flatMap` stage of `clientMessages` processing
 
     val fields = message.fieldsAsMap
-
-    val payload = message.payload.getOrElse(fields.getOrElse("submission", "{}")) match {
-      case p: String => parse(p) \ "payload"
-      case _ => JObject()
-    }
 
     for {
       clientId <- fields.get("clientId").asInstanceOf[Option[String]]
@@ -266,19 +262,17 @@ object LongitudinalView {
             if(clientId != currentClientId) {
               currentClientKeys.clear()
               currentClientKeysCount = 0
-              currentClientId = clientId.asInstanceOf[String]
+              currentClientId = clientId
             }
 
             if(currentClientKeysCount < DefaultMaxHistorySize) {
-              val key: (String, Int) = (startDate.asInstanceOf[String], counter.asInstanceOf[Int])
-              currentClientKeys += key
+              currentClientKeys += ((startDate, counter))
             }
 
             currentClientKeysCount += 1
 
-            if(currentClientKeysCount == filterLimit){
-              val clientKeys: (String, Set[(String, Int)]) = (clientId.asInstanceOf[String], currentClientKeys.toSet)
-              allowedClientKeys += clientKeys
+            if(currentClientKeysCount == filterLimit) {
+              allowedClientKeys += ((clientId, currentClientKeys.toSet))
             }
           }
 
@@ -350,13 +344,12 @@ object LongitudinalView {
       logger.info(s"Clients seen: $clientsSeen")
       logger.info(s"Clients ignored: $clientsIgnored")
 
-      if((1.0 * clientsIgnored) / clientsSeen > MaxFractionIgnoredClients){
+      if((1.0 * clientsIgnored) / clientsSeen > MaxFractionIgnoredClients) {
         throw new Exception(s"More clients ignored than are allowed. Ignored $clientsIgnored out of $clientsSeen clients.")
       }
     } catch {
       // Delete incomplete data
       case e: Exception =>
-        e.printStackTrace()
         s3Handler.deletePrefix(outputBucket, prefix)
         throw e
     }
--- a/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
@@ -8,7 +8,7 @@ import com.mozilla.telemetry.avro
 import com.mozilla.telemetry.heka.{Dataset, Message}
 import com.mozilla.telemetry.metrics._
 import com.mozilla.telemetry.parquet.ParquetFile
-import com.mozilla.telemetry.utils.{deletePrefix, _}
+import com.mozilla.telemetry.utils.{deletePrefix => utilsDeletePrefix, _}
 import org.apache.avro.generic.{GenericData, GenericRecord, GenericRecordBuilder}
 import org.apache.avro.{Schema, SchemaBuilder}
 import org.apache.spark.Partitioner
@@ -22,7 +22,14 @@ import scala.collection.mutable.ListBuffer
 import scala.math.abs
 import scala.reflect.ClassTag
 
-protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int = 1000) extends Iterator[List[Map[String, Any]]] {
+import scala.collection.mutable.ListBuffer
+
+class S3Handler extends Serializable {
+  def uploadFile(localFile: java.io.File, outputBucket: String, prefix: String): Unit = S3Store.uploadFile(localFile, outputBucket, prefix)
+  def deletePrefix(bucket: String, prefix: String): Unit = utilsDeletePrefix(bucket, prefix)
+}
+
+protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int) extends Iterator[List[Map[String, Any]]] {
   // Less than 1% of clients in a sampled dataset over a 3 months period has more than 1000 fragments.
   var buffer = ListBuffer[Map[String, Any]]()
   var currentKey =
@@ -84,6 +91,18 @@ object LongitudinalView {
   // Clients are ignored when they have any bad data of some sort
   val MaxFractionIgnoredClients = .005
 
+  // This value is the total number of pings that a client must have
+  // for their pingset to be early filtered (e.g. before partitioning)
+  //
+  // There is a tradeoff here:
+  //  - Increasing the size means that we filter fewer clients
+  //  - Decreasing the size means we increase the size of the data structure,
+  //    which is in memory on each executor (and the driver)
+  //
+  // From 2018/01/22 to 2018/06/22, 2158 clients had 5000 or more pings.
+  val DefaultFilterLimit = 5000
+  val DefaultMaxHistorySize = 1000
+
   // Column name prefix used to prevent name clashing between scalars and other
   // columns.
   private val ScalarColumnNamePrefix = "scalar_"
@@ -116,7 +135,7 @@ object LongitudinalView {
     }
   }
 
-  private class Opts(args: Array[String]) extends ScallopConf(args) {
+  class Opts(args: Array[String]) extends ScallopConf(args) {
     val from = opt[String]("from", descr = "From submission date. Defaults to 6 months before `to`.", required = false)
     val to = opt[String]("to", descr = "To submission date", required = true)
     val outputBucket = opt[String]("bucket", descr = "bucket", required = true)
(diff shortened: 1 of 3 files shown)


fbertsch:
@sunahsuh the idea here is to gather a list of allowed `(client_id, submission_date, subsession_counter)` tuples, distribute those to the executors (before we repartition the messages), and filter there.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +186,95 @@ object LongitudinalView {
...
+
+            currentClientKeysCount += 1
+
+            if(currentClientKeysCount == filterLimit){
+              val clientKeys: (String, Set[(String, Int)]) = (clientId.asInstanceOf[String], currentClientKeys.toSet)
+              allowedClientKeys += clientKeys
+            }
+          }
+
+          allowedClientKeys.iterator
```
fbertsch: Each partition will return a list of clients and their set of allowed `(submission_date, subsession_counter)` pairs. We can then use those to pre-filter the data in line 295.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +186,95 @@ object LongitudinalView {
...
-    deletePrefix(outputBucket, prefix)
+    s3Handler.deletePrefix(outputBucket, prefix)
 
     try {
       // Sort submissions in descending order
       implicit val ordering = Ordering[(String, String, Int)].reverse
+
+      val allowedLargeClientKeys = messages
+        .flatMap(getOrderKey)
+        .repartitionAndSortWithinPartitions(new ClientIdPartitioner(numPartitions))
```
fbertsch: I used `ClientIdPartitioner` and `mapPartitions` to mirror the partitioning and sorting already done, which should avoid unintended consequences. (I had considered using Dataframes, but I can't verify the behavior would be the same).

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +186,95 @@ object LongitudinalView {
...
+              allowedClientKeys += clientKeys
+            }
+          }
+
+          allowedClientKeys.iterator
+        }
+        .collect()
+        .toMap
+
+      val clientLookupMap = messages.sparkContext.broadcast(allowedLargeClientKeys)
```
fbertsch: Clients are only included in the `clientLookupMap` if they have more than `DefaultFilterLimit` pings. Clients who are not in the `clientLookupMap` will have all of their pings aggregated.

fbertsch:
Good news, the run completed. Some stats:
- Took about 30 minutes longer (time spent creating the pre-filter client map)
- Only 52 clients ignored
- 1248 parquet files from a 39-node cluster (so every partition was successfully written)

This definitely fixed the issue, and 5000 is a reasonable number to set the initial value.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -22,7 +22,14 @@ import scala.collection.mutable.ListBuffer
...
 
-protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int = 1000) extends Iterator[List[Map[String, Any]]] {
+import scala.collection.mutable.ListBuffer
+
+class S3Handler extends Serializable {
+  def uploadFile(localFile: java.io.File, outputBucket: String, prefix: String): Unit = S3Store.uploadFile(localFile, outputBucket, prefix)
+  def deletePrefix(bucket: String, prefix: String): Unit = utilsDeletePrefix(bucket, prefix)
+}
+
+protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int) extends Iterator[List[Map[String, Any]]] {
```
mreid-moz: If we're removing the default of 1000 records, we should probably update the comment on the next line.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -84,6 +91,18 @@ object LongitudinalView {
   // Clients are ignored when they have any bad data of some sort
   val MaxFractionIgnoredClients = .005
 
+  // This value is the total number of pings that a client must have
+  // for their pingset to be early filtered (e.g. before partitioning)
```
mreid-moz: Please expand on "early filtered" - specifically, mentioning that clients with more pings will be excluded from the final dataset.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -22,7 +22,14 @@ import scala.collection.mutable.ListBuffer
 import scala.math.abs
 import scala.reflect.ClassTag
 
-protected class ClientIterator(it: Iterator[(String, Map[String, Any])], maxHistorySize: Int = 1000) extends Iterator[List[Map[String, Any]]] {
+import scala.collection.mutable.ListBuffer
+
+class S3Handler extends Serializable {
```
mreid-moz: Please add a comment what this class is for.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
fbertsch: Default is still 1000 records.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+    // Only show parent scalars
+    val scalarDefinitions = Scalars.definitions(includeOptin).filter(_._2.process == Some("parent"))
+
+    run(opts.to(), opts.outputBucket(), messages, sc.defaultParallelism, handler, histogramDefinitions, scalarDefinitions, DefaultFilterLimit)
+
+    sc.stop()
+  }
+
+  private def getOrderKey(message: Message): Option[((String, String, Int), Int)] = {
+    // This replicate the existing code, except returns a small data structure rather
```
mreid-moz: nit: please specify which existing code (the `flatMap` stage of `clientMessages` processing)

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+    // than the entire ping, and doesn't filter in the for loop
+
+    val fields = message.fieldsAsMap
+
+    val payload = message.payload.getOrElse(fields.getOrElse("submission", "{}")) match {
```
mreid-moz: `payload` is unused, this parsing can be removed. (Aside: why didn't scalastyle notice that?)

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+
+    // Only show parent scalars
+    val scalarDefinitions = Scalars.definitions(includeOptin).filter(_._2.process == Some("parent"))
+
+    run(opts.to(), opts.outputBucket(), messages, sc.defaultParallelism, handler, histogramDefinitions, scalarDefinitions, DefaultFilterLimit)
+
+    sc.stop()
+  }
+
+  private def getOrderKey(message: Message): Option[((String, String, Int), Int)] = {
```
mreid-moz: Optional clean-up: consider introducing a separate type for this `(String, String, Int)` tuple to make it clearer that it is `clientId`, `startDate`, `profileSubsessionCounter`.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+          var currentClientKeys: ListBuffer[(String, Int)] = new ListBuffer()
+          var currentClientKeysCount: Int = 0
+
+          while(it.hasNext) {
+            val ((clientId, startDate, counter), _) = it.next()
+
+            if(clientId != currentClientId) {
+              currentClientKeys.clear()
+              currentClientKeysCount = 0
+              currentClientId = clientId.asInstanceOf[String]
```
mreid-moz: `clientId` must already be a `String` here, the type coercion should not be needed.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+            }
+
+            if(currentClientKeysCount < DefaultMaxHistorySize) {
+              val key: (String, Int) = (startDate.asInstanceOf[String], counter.asInstanceOf[Int])
```
mreid-moz: Same re: type coercions - the iterator is already typed.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -167,25 +200,95 @@ object LongitudinalView {
...
+              currentClientKeys += key
+            }
+
+            currentClientKeysCount += 1
+
+            if(currentClientKeysCount == filterLimit){
```
mreid-moz: style nit: add a space between `)` and `{`

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -255,7 +356,8 @@ object LongitudinalView {
     } catch {
       // Delete incomplete data
       case e: Exception =>
-        deletePrefix(outputBucket, prefix)
+        e.printStackTrace()
```
mreid-moz: Please use `logger` to print this (and ideally a message about where it blew up we can grep for)

src/test/scala/com/mozilla/telemetry/views/LongitudinalViewTest.scala
```diff
@@ -4,28 +4,51 @@
...
+
 import scala.collection.JavaConversions._
 import scala.collection.mutable.WrappedArray
 import scala.language.reflectiveCalls
 import scala.io.Source
 
+// ScalaMock objects and functions are not serializable
+// so we'll need to mock the class ourselves
+// ref: https://groups.google.com/forum/#!topic/scalatest-users/7t7MVNzGUTQ
+class MockHandler(outputPath: String) extends S3Handler with Serializable {
```
mreid-moz: Sorry for your luck :-/

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
```diff
@@ -201,20 +304,18 @@ object LongitudinalView {
               info = parse(json)
               JString(startDate) <- (info \ "subsessionStartDate").toOption
               JInt(counter) <- (info \ "profileSubsessionCounter").toOption
-            } yield ((clientId, startDate, counter.toInt),
+              subsessionCounter = counter.toInt
+
+              if(!clientLookupMap.value.contains(clientId) || clientLookupMap.value(clientId).contains((startDate, subsessionCounter)))
```
mreid-moz: TIL about this pattern in a `for` block. Cool!

src/test/scala/com/mozilla/telemetry/views/LongitudinalViewTest.scala
```diff
@@ -592,15 +653,15 @@ class LongitudinalViewTest extends FlatSpec with Matchers with PrivateMethodTest
   "ClientIterator" should "not trim histories of size < 1000" in {
     val template = ("foo", Map("client" -> "foo"))
     val history = List.fill(ParentConstant)(template)
-    val split_history = new ClientIterator(history.iterator).toList
+    val split_history = new ClientIterator(history.iterator, LongitudinalView.DefaultMaxHistorySize).toList
```
mreid-moz: Should these use a literal `1000` to match the test case names and list-filling and assert behaviour?

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
fbertsch: I also wanted make this a case class, but ordering of those isn't completely clear. For example, see the complications here: https://meta.plasm.us/posts/2013/10/13/ordering-case-classes/

My goal was to leave the existing code unchanged (except for adding that `if` statement) and mirror it exactly in partitioning and ordering.

src/test/scala/com/mozilla/telemetry/views/LongitudinalViewTest.scala
fbertsch: I'd rather keep these as-is, that way if we change that default for some reason, we have to change the tests too in order to verify the new value will work as expected.

src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala
mreid-moz: Ok, it makes sense to minimize the change that's being made here.

PR #461, opened 2018-08-08
Bug 1477719 - Fail if too many records are ignored in Main Summary

akkomar:
This makes MainSummary job fail if more than 0.5% records read were ignored (not deserialized properly).

Since we can get the ignore ration only after output is written, in case of failure output location will contain data. I'm not deleting it since the assumption is it will be overwritten anyway and it might come in handy for developers.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
@@ -7,18 +7,15 @@ import java.time._
 import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
 
-import org.apache.spark.sql.types._
+import com.mozilla.telemetry.heka.Dataset
+import com.mozilla.telemetry.metrics._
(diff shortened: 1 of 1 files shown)


akkomar:
Oh, didn't realize there are other jobs depending on this one not run from Airflow. Switching to deleting data then...

Re `.005` - we're not ignoring anything recently. How about going down to `.00005` - that would correspond to ~9k ignores on recent days' data.

We could also make a metric out of it.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -28,6 +25,9 @@ object MainSummaryView {
   def schemaVersion: String = "v4"
   def jobName: String = "main_summary"
 
+  // Allow at most .5% of records to be ignored
+  // Records are ignored when we can't properly deserialize them
+  val MaxFractionIgnoredClients = .005
```
fbertsch: Just noticed that this should be `MaxFractionIgnoredPings`!

PR #463, opened 2018-08-08
Add font editor histograms to main_summary

tdsmith:
I've been fielding some questions from product about these and we'll want to see what happens in the bigger release channels. Right now, these histograms are set to expire in 66; does that pose any concerns for y'all?


tdsmith:
If you feel okay about this as-is, that's good enough for me. A comment might get stale; we'll see if these provide useful feedback before trying to extend them.

tdsmith:
This will be conflicty with #458 so I'll just do it there.

PR #464, opened 2018-08-08
Add --bucket-protocol option to DesktopDau

relud:
as per discussion on [mozilla/telemetry-airflow#334](https://github.com/mozilla/telemetry-airflow/pull/334#discussion_r208314015)


relud:
yes. this allows for testing against the local filesystem, instead of the `s3://` one that isn't the same outside emr and doesn't support setting an endpoint url that can be pointed to moto or the like

PR #465, opened 2018-08-08
In-content search variants

wlach:


wlach:
@fbertsch See https://bugzilla.mozilla.org/show_bug.cgi?id=1481671 for context

wlach:
@fbertsch As discussed on slack, I think we want the .replace, as your expanded regex with cause search histograms keys like `sap-follow-on:urlbar:SomeCodeHere` to be incorrectly assigned a search engine corresponding to their search type (`sap-follow-on`). Another related side benefit is that the replace only solves a specific problem, making it less likely that this change will cause unanticipated breakage.

PR #466, opened 2018-08-09
Handle missing days in GenericDauTrait

relud:


src/main/scala/com/mozilla/telemetry/views/dau/GenericDauTrait.scala
```diff
@@ -29,7 +29,7 @@ trait GenericDauTrait {
       // detect input date column
       .option("basePath", conf.inputBasePath)
       // read input dates by path
-      .parquet(conf.inputPaths:_*)
+      .parquet(conf.inputPaths.filter(hadoopExists):_*)
```
sunahsuh: Hmm, I'm inclined to just let spark deal with this -- if you're concerned about the schema merge I *believe* Spark will only attempt to merge the schema for the partitions that are requested instead of for all time.

relud: We just determined that Spark won't do this for us. see [bug 1482287](https://bugzilla.mozilla.org/show_bug.cgi?id=1482287) for test and results.

PR #467, opened 2018-08-13
Add view for active DAU

relud:
and clean up other dau issues

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/dau/DesktopActiveDauView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/dau/DesktopActiveDauView.scala
@@ -3,8 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.views.dau
 
-object DesktopDauView extends GenericDauTrait {
-  val jobName: String = "desktop_dau"
+object DesktopActiveDauView extends GenericDauTrait {
+  val jobName: String = "desktop_active_dau"
 
   def getGenericDauConf(args: Array[String]): GenericDauConf = {
     val conf = new BaseCliConf(args)
@@ -12,8 +12,11 @@ object DesktopDauView extends GenericDauTrait {
     GenericDauConf(
       conf.from.getOrElse(conf.to()),
       conf.to(),
-      inputBasePath = s"${conf.bucketProto()}${conf.bucket()}/client_count_daily/v2",
-      outputBasePath = s"${conf.bucketProto()}${conf.bucket()}/desktop_dau/v1"
+      countColumn = Some("client_id"),
+      inputBasePath = s"${conf.bucketProto()}${conf.bucket()}/clients_daily/v6",
+      inputDateColumn = "submission_date_s3",
+      outputBasePath = s"${conf.bucketProto()}${conf.bucket()}/desktop_active_dau/v1",
+      where = "scalar_parent_browser_engagement_total_uri_count_sum >= 5"
     )
   }
 }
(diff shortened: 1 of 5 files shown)


src/test/scala/com/mozilla/telemetry/views/dau/GenericDauConfTest.scala
```diff
@@ -47,9 +47,7 @@ class GenericDauConfTest extends GenericDauTraitTest {
         (1 to 3).map{_=>("2018-01-31", randomUUID.toString, randomUUID.toString, "C", "c")}.toList :::
         (1 to 2).map{_=>("2018-01-31", randomUUID.toString, randomUUID.toString, "C", "d")}.toList :::
         (1 to 2).map{_=>("2018-02-01", randomUUID.toString, randomUUID.toString, "C", "c")}.toList :::
-        (1 to 3).map{_=>("2018-02-01", randomUUID.toString, randomUUID.toString, "C", "d")}.toList :::
-        // pad the list so that all needed dates exist
-        (4 to 30).map { d => (f"2018-01-$d%02d", null, null, null, null) }.toList
+        (1 to 3).map{_=>("2018-02-01", randomUUID.toString, randomUUID.toString, "C", "d")}.toList
```
relud: removed unnecessary padding because as of 8d6d5e9 we handle missing dates

src/main/scala/com/mozilla/telemetry/views/dau/DesktopActiveDauView.scala
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.views.dau
+
+import org.rogach.scallop.{ScallopConf, ScallopOption}
+
+object DesktopActiveDauView extends GenericDauTrait {
+  class CliConf(args: Array[String]) extends ScallopConf(args) {
```
sunahsuh: Seems like we're re-using these arguments a few times -- perhaps we should have a base ScallopConf alongside the trait?

relud: Good point. I moved it to the trait, and it can be overridden as necessary.

src/main/scala/com/mozilla/telemetry/views/dau/GenericDauTrait.scala
```diff
@@ -11,11 +11,34 @@ import com.mozilla.telemetry.utils.{getOrCreateSparkSession, hadoopExists}
...
+      "bucket",
+      default = Some("telemetry-parquet"),
+      descr = "bucket where input and output data sets are stored",
+      required = false)
+    val bucketProto: ScallopOption[String] = opt[String](
+      "bucket-protocol",
+      default = Some("s3://"),
+      descr = "hadoop compatible filesystem protocol to be used with --bucket",
+      required = false)
+    verify()
```
sunahsuh: I'd leave off the `verify()` and let subclasses do that instead -- this way, every job can start with this conf and add more fields as they'd like (see arkadiusz's the streaming base class in telemetry-streaming for this pattern)

src/main/scala/com/mozilla/telemetry/views/dau/DesktopDauView.scala
```diff
@@ -3,35 +3,12 @@
...
-      descr = "hadoop compatible filesystem protocol to be used with --bucket",
-      required = false)
-    verify()
-  }
-
   val jobName: String = "desktop_dau"
 
   def getGenericDauConf(args: Array[String]): GenericDauConf = {
     val conf = new CliConf(args)
+    conf.verify()
```
relud: @sunahsuh is this what you had in mind for moving `verify()`?

PR #468, opened 2018-08-13
Bug 1478474 - Add ammonite to sbt

relud:
I don't recall if anyone objected to this in the team meeting when it was brought up, but I needed it today.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,7 +53,8 @@ lazy val root = (project in file(".")).
     // Test dependencies
     libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.5" % Test,
     libraryDependencies += "com.github.tomakehurst" % "wiremock-standalone" % "2.18.0" % Test,
-    libraryDependencies += "com.holdenkarau" %% "spark-testing-base" % "2.3.0_0.9.0" % Test
+    libraryDependencies += "com.holdenkarau" %% "spark-testing-base" % "2.3.0_0.9.0" % Test,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -94,3 +95,9 @@ updateOptions := updateOptions.value.withLatestSnapshots(false)
 val scalaStyleConfigUrl = Some(url("https://raw.githubusercontent.com/mozilla/moztelemetry/master/scalastyle-config.xml"))
 (scalastyleConfigUrl in Compile) := scalaStyleConfigUrl
 (scalastyleConfigUrl in Test) := scalaStyleConfigUrl
+
+// Configure ammonite to replace `sbt console`
+addCommandAlias(name = "ammonite", value = "test:runMain ammonite.Main")
+addCommandAlias(name = "amm", value = "ammonite")
+addCommandAlias(name = "console", value = "ammonite")
+trapExit in (Test, run) := false
```
relud: This clears up the error when exiting the REPL, but it also makes sbt exit when ammonite exits, rather than allowing further sbt commands.

I think that writing an alternative main method could fix this better, but I don't know how to do that correctly, and just adding an object somewhere under test seems bad.

sunahsuh: Paging @akkomar

akkomar: This seems somewhat similar to https://github.com/lihaoyi/Ammonite/issues/738.

Does bumping sbt to v1.2.1 help?

relud: no effect

PR #469, opened 2018-08-17
TAAR: Removed the hardcoded blacklist 

crankycoder:
This replaces the hardcoded blacklist and replaces it with an S3 loaded whitelist for TAAR addons.

The logic in getAddonData has been inverted to assert that an addonId is contained in the whitelist instead of not within a blacklist.

cc/ @mlopatka

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
+++ b/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
@@ -100,7 +100,7 @@ object AddonRecommender {
   /**
     * Get a new dataset from the Longitudinal dataset containing only the relevant addon data.
     * @param sparkSession The SparkSession used for loading the Longitudinal dataset.
-    * @param addonBlacklist The a list of addon GUIDs to exclude.
+    * @param addonWhitelist The a list of addon GUIDs that are valid
     * @param amoDB The AMO database fetched by AMODatabase.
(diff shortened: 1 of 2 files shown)


crankycoder:
I've manually tested the dataload of the S3 data using the ./run-sbt.sh and executing the scala REPL using `console`, but haven't had a chance to test that this job runs locally.

crankycoder:
cc/ @dexterp37 can you take a quick look at this?  I've inverted the use of the blacklist and turned it into a whitelist.

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -105,7 +105,7 @@ object AddonRecommender {
     * @return A 4 columns Dataset with each row having the client id, the addon GUID
     *         and the hashed client id and GUID.
     */
-  private def getAddonData(sparkSession: SparkSession, addonBlacklist: List[String], amoDB: Map[String, Any]) = {
+  private def getAddonData(sparkSession: SparkSession, addonWhitelist: List[String], amoDB: Map[String, Any]) = {
```
Dexterp37: Since you've changed the variable name, please also update the docstring which is still mentioning the old variable name a few lines above.

crankycoder: Thanks!  I've just pushed the docstring change.

PR #470, opened 2018-08-21
WIP Add option to override longitudinal dataset location to addon recomme…

sunahsuh:
…nder

@crankycoder I'd be tagging you for review but I can't add you as a reviewer


sunahsuh:
marking as WIP since I need to fix one of the tests

sunahsuh:
The backstory is that I generated a manual run of the longitudinal dataset so we could re-run all the longitudinal-based the TAAR jobs with fresh data before the v3 experiment launch. I added similar pieces to the other jobs in python_mozetl

PR #476, opened 2018-08-31
Bug 1487902 - Fix bucket options in CrashSummary and Sync views

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1487902)

The crash summary and sync views are currently failing due to changes in the base options. This PR sets the bucket value as not required and removes the conflicting options from the sync view. Having an optional bucket means that current jobs may fail with run-time errors if poorly specified.

I will post confirmation from manual runs of this patch on ATMO.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/BatchJobBase.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/BatchJobBase.scala
@@ -50,7 +50,7 @@ abstract class BatchJobBase extends Serializable {
     val outputBucket = opt[String](
       "bucket",
       descr = "Destination bucket for parquet data",
-      required = true)
+      required = false)
   }
 
   protected def shouldStopContextAtEnd(spark: SparkSession): Boolean = {
--- a/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/SyncEventView.scala
@@ -20,7 +20,6 @@ object SyncEventView extends BatchJobBase {
 
   // Configuration for command line arguments
   private class Conf(args: Array[String]) extends BaseOpts(args) {
-    override val outputBucket = opt[String]("bucket", descr = "Destination bucket for parquet data", required = false)
     val outputFilename = opt[String]("outputFilename", descr = "Destination local filename for parquet data", required = false)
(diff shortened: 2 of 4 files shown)


acmiyaguchi:
`spark-submit --master yarn --class com.mozilla.telemetry.views.CrashSummaryView telemetry-batch-view-1.1.jar `
before: 
```
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
[scallop] Error: Required option 'bucket' not found
```

after:

image

`spark-submit --master yarn --class com.mozilla.telemetry.views.SyncView telemetry-batch-view-1.1.jar --bucket telemetry-test-bucket`

before:
```
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
Picked up _JAVA_OPTIONS: -Djava.io.tmpdir=/mnt1/ -Xmx15000M -Xms1000M
[scallop] Error: Option identifier 'bucket' is not unique
```

after:
image

acmiyaguchi:
There's an option to write to local disk if the bucket doesn't exist in SyncView:

https://github.com/mozilla/telemetry-batch-view/blob/cf5be4aff11b7f6ee7d3ea5db236b64c0ffe732c/src/main/scala/com/mozilla/telemetry/views/SyncView.scala#L109-L112

I don't think this is a useful option to have around, so it could probably be removed eventually. The override keyword doesn't look like it was working the way it should have been; I should have added this to the set of jobs that I tested. 

I also have a bug filed for the [CrashViewSummary](https://bugzilla.mozilla.org/show_bug.cgi?id=1487905) on it's inconsistent bucket option, but its a two part change that needs to be done with airflow. 

I think I should have run all of the jobs for ~5 minutes a piece to avoid this kind of issue. I added a `scripts/test_run.sh` script where I intended to add testing configurations for all of the jobs. It ended up choosing a subset (that didn't include sync view) to save time. Here, I ended up running the job and `ctrl-c`'ed out after the job started to run as confirmation. This doesn't require any extra testing options, so I think it should be possible to script this as a sanity check.

PR #477, opened 2018-09-05
Bug 1458356 - Migrate jodatime usage in telemetry-batch-view to java.time API

acmiyaguchi:
[bug link](Bug 1458356)

This removes `joda-time` and the scala wrapper `nscala_time` from this repo. I tested the changes against `scripts/test_run.sh` which runs jobs with defaults and times out after a minute. If our jobs are going to fail at the entrypoint, they usually do it around the 40 second mark.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
+++ b/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
@@ -5,6 +5,8 @@ package com.mozilla.telemetry.ml
 
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Paths}
+import java.time.{Instant, ZoneOffset}
+import java.time.format.DateTimeFormatter
 
 import breeze.linalg.{DenseMatrix, DenseVector}
 import com.github.fommil.netlib.BLAS.{getInstance => blas}
@@ -13,7 +15,6 @@ import org.apache.spark.ml.evaluation.NaNRegressionEvaluator
 import org.apache.spark.ml.recommendation.{ALS, ALSModel}
 import org.apache.spark.ml.tuning.{CrossValidator, ParamGridBuilder}
 import org.apache.spark.sql.SparkSession
-import org.joda.time.{DateTime, format}
 import org.json4s.JsonDSL._
 import org.json4s._
 import org.json4s.jackson.JsonMethods._
@@ -305,10 +306,10 @@ object AddonRecommender {
         val cwd = new java.io.File(".").getCanonicalPath
         val outputDir = new java.io.File(cwd, output)
 
-        val fmt = format.DateTimeFormat.forPattern("yyyyMMdd")
+        val fmt = DateTimeFormatter.ofPattern("yyyyMMdd")
         val date = conf.train.runDate.get match {
           case Some(f) => f
-          case _ => fmt.print(DateTime.now)
+          case _ => Instant.now().atOffset(ZoneOffset.UTC).format(fmt)
         }
 
         train(outputDir.getCanonicalPath, date, conf.train.privateBucket(), conf.train.publicBucket(), conf.train.longitudinalOverride.toOption)
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
@@ -3,8 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.views
 
-import java.time.{LocalDate, ZoneId}
-import java.util.Date
+import java.time.{LocalDate, OffsetDateTime, ZoneOffset}
+import java.time.format.DateTimeFormatter
 
 import com.mozilla.telemetry.utils.{deletePrefix, getOrCreateSparkSession}
 import org.apache.spark.sql.SparkSession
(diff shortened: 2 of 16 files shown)


src/test/scala/com/mozilla/telemetry/utils/UtilsTest.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  it should "wrap the timezone" in {
+    val cases = Seq(
+      ("2018-09-01T08:00:00.0-08:00", ZoneOffset.ofHours(-8)),
+      ("2018-09-01T08:00:00.0-00:00", ZoneOffset.ofHours(0)),
+      ("2018-09-01T08:00:00.0-12:00", ZoneOffset.ofHours(-12)),
+      ("2018-09-01T08:00:00.0-13:00", ZoneOffset.ofHours(-1)),
+      ("2018-09-01T08:00:00.0+14:00", ZoneOffset.ofHours(14)),
+      ("2018-09-01T08:00:00.0+15:00", ZoneOffset.ofHours(3)))
```
acmiyaguchi: I'm not sure if `+15:00` is actually equivalent to `+03:00`, I feel like this might be off by a day.

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -305,10 +306,10 @@ object AddonRecommender {
         val cwd = new java.io.File(".").getCanonicalPath
         val outputDir = new java.io.File(cwd, output)
 
-        val fmt = format.DateTimeFormat.forPattern("yyyyMMdd")
+        val fmt = DateTimeFormatter.ofPattern("yyyyMMdd")
         val date = conf.train.runDate.get match {
           case Some(f) => f
-          case _ => fmt.print(DateTime.now)
+          case _ => Instant.now().atOffset(ZoneOffset.UTC).format(fmt)
```
acmiyaguchi: `Instant.now()` should be equivalent to `LocalTime.now(Clock.systemUTC())`, since `Instance` is relative to epoch by default. The zone offset transforms the instant into a `OffsetDateTime` which can be formatted with `DateTimeFormatter`.

src/main/scala/com/mozilla/telemetry/views/QuantumRCView.scala
```diff
@@ -40,23 +42,22 @@ object QuantumRCView extends BatchJobBase {
...
-  private val fmt = DateTimeFormat.forPattern("yyyyMMdd")
+  private val fmt = DateTimeFormatter.ofPattern("yyyyMMdd")
 
-  // Joda time does not have a nice way of getting the first Sunday of the
-  // current week, so this is a workaround.
-  private def getFirstDayOfWeek(other: DateTime) = other.dayOfWeek.get match {
-    case 7 => other
-    case _ => other.minusWeeks(1).withDayOfWeek(7)
+  // Get Sunday as the first day of the week
+  private def getFirstDayOfWeek(other: LocalDate) = other.getDayOfWeek() match {
```
acmiyaguchi: Apparently when the zone is set to US, the first day of the week is actually Sunday. Date-times and calendaring is hard.

sunahsuh:
Doh, looks like a real test failure in longitudinal here

src/test/scala/com/mozilla/telemetry/views/LongitudinalViewTest.scala
```diff
@@ -345,14 +345,14 @@ class LongitudinalViewTest extends FlatSpec with Matchers with PrivateMethodTest
 
   "Scalar fields" must "be converted correctly" in {
     val stringFields = Array(
-      "submission_date"       -> "2016-01-28T00:00:00.000Z",
```
acmiyaguchi: Apparently, millisecond precision is not part of the iso8601 format.

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
acmiyaguchi: I'm not too sure when we should be using Instants over LocalDate. Here, I think it was easier to work with the epoch time directly, as opposed to worrying about setting the LocalDate properly with the UTC clock. In the other jobs, the clock was available directly through the base class.

src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala
```diff
@@ -152,19 +152,27 @@ object ExperimentSummaryView extends BatchJobBase {
...
-  def shouldProcessExperiment(r: NormandyRecipe, date: Date): Boolean = {
+  def shouldProcessExperiment(r: NormandyRecipe, date: OffsetDateTime): Boolean = {
     experimentsQualifyingAction.contains(r.action) &&
     r.arguments.isHighVolume != Some(true) &&
     !excludedExperiments.contains(r.arguments.slug.get) &&
-    ((r.enabled == true) || r.last_updated.after(date)) // is this experiment enabled for this date?
+    (r.enabled || r.last_updated.isAfter(date)) // is this experiment enabled for this date?
   }
 
+  // Custom json4s serializer for java time
```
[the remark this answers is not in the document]
acmiyaguchi: Having the full set of java.time dates would be nice, otherwise it'll just be a one-off.

acmiyaguchi:
Thanks @sunahsuh, I added some documentation to the testing script. 

In general, the java time library is pretty nice to use. I definitely ran into some crufty code around longitudinal, but overall the changes were straightforward. Also if there's use for java serializers for json4s in the future in other jobs, it might be nice to have around.

PR #486, opened 2018-10-11
Bug 1497857 Reduce memory usage in ExperimentEngagementAnalyzer

jklukas:
Ran this successfully as a one-off for the experiment that has been failing the past few days, so I think this accomplishes what we need.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
@@ -206,6 +206,8 @@ object ExperimentAnalysisView extends BatchJobBase {
 
       val metadata = ExperimentAnalyzer.getExperimentMetadata(persisted).collect()
 
+      persisted.unpersist()
+
       metadata ++ baseMetrics
     }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/views/ExperimentAnalysisView.scala
```diff
@@ -206,6 +206,8 @@ object ExperimentAnalysisView extends BatchJobBase {
 
       val metadata = ExperimentAnalyzer.getExperimentMetadata(persisted).collect()
 
+      persisted.unpersist()
```
sunahsuh: Technically because the [spark context should be stopped](https://github.com/mozilla/telemetry-batch-view/pull/486/files#diff-f0a72dc1efa84cdd5d8ec4e13ed89fc1R136) on runs via EMR this shouldn't make a difference in memory usage (definitely still makes sense to add this here in case we run this on databricks though)

jklukas:
I was totally unaware that unpersist was cascading, so I agree that there's likely more code in ExperimentEngagementAnalyzer that isn't reusing cache where expected. I may do a more comprehensive review of that later.

PR #487, opened 2018-10-11
Bug 1498309 Refactor CircleCI config and use cache for assembly

jklukas:


.circleci/config.yml
```diff
@@ -6,6 +6,12 @@
 #   https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
 ####################
 
+####################
+# Version: determines what features are available.
+####################
+
+version: 2.1
```
jklukas: CircleCI 2.1 gives us the ability to create custom "commands" and "executors" which we take advantage of here to clean things up a bit.

.circleci/config.yml
```diff
@@ -14,74 +20,96 @@
...
-prep_cache_settings: &prep_cache_settings
-  name: Clean directories to restore
-  command: |
-    rm -r ~/.ivy2 ~/.sbt
-
-restore_cache_settings: &restore_cache_settings
-  keys:
-    - telemetry-batch-view-{{ .Branch }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
-    - telemetry-batch-view-{{ .Branch }}
-    - telemetry-batch-view-
```
jklukas: These three templates have been refactored into two commands below.

.circleci/config.yml
```diff
@@ -14,74 +20,96 @@
...
-
-# At time of writing, we use the default container size (2 CPUs and 4GB memory),
-# but may renogotiate with CircleCI and enable larger resource_class
-# options at which point we might want to bump up the Java memory settings
-# and reduce the test parallelism.
-common_settings: &common_settings
-  docker:
-    - image: *sbt_image
-  working_directory: /telemetry-batch-view
-  # resource_class: xlarge
```
jklukas: This common_settings template has been refactored to be an executor

.circleci/config.yml
```diff
@@ -14,74 +20,96 @@
...
-    fi
-
-early_return_for_skip_tests: &early_return_for_skip_tests
-  name: Early return if the latest non-merge commit message contains "[skip-tests]"
-  command: |
-    COMMIT_MESSAGE=$(git log --format=%B --no-merges -n 1)
-    if [[ "$COMMIT_MESSAGE" =~ "[skip-tests]" ]]; then
-      echo "Skipping tests due to [skip-tests] flag, so marking this step successful"
-      circleci step halt
-    fi
```
jklukas: These two early return templates are now defined as commands.

PR #488, opened 2018-10-16
Bug 1499365 - Handle missing columns in ClientsDailyView

relud:


src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -68,21 +69,29 @@ object ClientsDailyView {
...
+    val agg = Some(fieldAggregators
+      // unresolvedAttributeNames gets the names of columns used as inputs to an expression
+      .flatMap(unresolvedAttributeNames)
+      // filter out the column names that are present in df
+      .toSet -- df.columns)
+      // filter Some(missingColumns) to None unless nonEmpty
+      .filter(_.nonEmpty)
+      // filter expressions using missingColumns out of fieldAggregators
+      .map(missingColumns =>
+        // TODO throw warning
```
relud: I don't know how to throw a warning, and the internet is *really* unhelpful. I think I should just print a warning to the logger, and move on?

sunahsuh: That sounds like a fine plan to me!

PR #490, opened 2018-10-18
Bug 1485152: Add full active_addons list to ClientsDailyView

sunahsuh:
- Removes deprecated active_experiments columns
- Refactors individual udfs into their own files in a new `com.mozilla.telemetry.utils.udfs` package
- Adds a udf for aggregating lists of structs using a column in the struct as a unique index
- Adds the active_addons column to ClientsDailyView

I considered using collect_set but I realized on days where an addon changed (e.g. a version update) we'd end up with duplicate entries for the addon, and given the initial use case for this column is TAAR I think having one entry per addon_id per day makes more sense.

Apologies for the rearrangement in ClientsDailyView, which makes it harder to see the changes -- I let IntelliJ clean up the code so I could clean up imports :/

cc @crankycoder and @benmiroglio as folks who might want to know when this PR gets merged


src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTest.scala
```diff
@@ -119,6 +123,21 @@ class ClientsDailyViewTest extends FlatSpec with Matchers with DataFrameSuiteBas
     )
   }
 
+  it must "handle 'active_addons' properly" in {
+    test(
+      List(
+        MainSummaryRow(),
+        MainSummaryRow(active_addons = Some(Seq(Addon("a")))),
+        MainSummaryRow(active_addons = Some(Seq(Addon("a"), Addon("b", true), Addon("c")))),
+        MainSummaryRow(active_addons = Some(Seq(Addon("b", false))))
```
relud: Are addons nullable within the list in main_summary? e.g. `Some(Seq(None)))`
Is the addon id nullable within the list in main_summary? e.g. `Some(Seq(Addon(null)))`

If so, can we add tests for those cases?

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTestHelpers.scala
```diff
@@ -21,9 +23,8 @@ object ClientsDailyViewTestHelpers {
     aborts_content: Option[Int] = None,
     aborts_gmplugin: Option[Int] = None,
     aborts_plugin: Option[Int] = None,
+    active_addons: Option[Seq[Addon]] = None,
```
relud: on line 165 I had to use `Option[scala.collection.Seq[SearchCount]]`. I don't recall why, but it definitely broke tests not to.

Could you remove `scala.collection.` there, or add it here, or comment why they are different?

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTestHelpers.scala
```diff
@@ -21,9 +23,8 @@ object ClientsDailyViewTestHelpers {
...
     active_addons_count: Option[Long] = None,
-    active_experiment_branch: Option[String] = None,
-    active_experiment_id: Option[String] = None,
```
relud: `active_experiment_id` and `active_experiment_branch` need to be removed at each of their two later occurrences in this file for tests to pass

sunahsuh: Blargh there was a bad last-minute merge when I collided with your changes from yesterday T_T

src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala
```diff
@@ -311,4 +188,143 @@ object ClientsDailyView {
...
+      .toSet -- df.columns)
+      // filter Some(missingColumns) to None unless nonEmpty
+      .filter(_.nonEmpty)
+      // filter expressions using missingColumns out of fieldAggregators
+      .map { missingColumns =>
+      logger.warn(s"JOB $jobName $schemaVersion MISSING INPUT COLUMNS: ${missingColumns.mkString(", ")}")
+      fieldAggregators
+        .filter(unresolvedAttributeNames(_)
+          .forall(!missingColumns.contains(_)))
+    }
```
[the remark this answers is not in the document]
sunahsuh: Do you mean just moving that closing brace back up inline?

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTest.scala
sunahsuh: I just looked at the schema definition again and 1. addon_id is required and 2. containsNull is false in the active_addons array definition

src/test/scala/com/mozilla/telemetry/views/ClientsDailyViewTestHelpers.scala
sunahsuh: I just took it out and tests passed.. not sure what's up

relud: wfm :D

PR #515, opened 2018-11-16
Bug 1504247 - Add page load time metric to main_summary

tdsmith:
We'd like to be able to track this for WebRender.


tdsmith:
Thanks!

PR #517, opened 2018-12-14
De-whitelist removed histogram

jklukas:
It appears that "DEVTOOLS_DEVELOPERTOOLBAR_OPENED_COUNT"
no longer exists in mozilla-central.


jklukas:
This histogram was added in  #458, so cc @tdsmith on the removal. We're trying to track down the mozilla-central change that introduced this.

tdsmith:
This makes sense; the developer toolbar feature was removed and we don't need to track it anymore. Thanks for the heads up!

PR #518, opened 2019-01-17
 Switch AddonRecommender to clients_daily 

akkomar:
Some notes:
* switching to clients_daily increases training dataset ~100x
* we should be able to run this daily (test runs were around 40 mins)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
+++ b/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
@@ -113,7 +113,7 @@ object AddonRecommender extends DatabricksSupport {
       .where("channel = 'release'")
       .where("app_name = 'Firefox'")
       .where(s"submission_date_s3 >= '$dateFrom'")
-      .where(s"sample_id <= '$sampling'")
+      .where(s"sample_id < '$sampling'")
       .selectExpr(
         "client_id",
         "active_addons",
--- a/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
+++ b/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
@@ -5,8 +5,8 @@ package com.mozilla.telemetry.ml
 
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Paths}
-import java.time.{Instant, ZoneOffset}
 import java.time.format.DateTimeFormatter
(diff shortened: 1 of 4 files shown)


akkomar:
Hey @Dexterp37 do you recommend any specific way for testing change to `clients_daily` dataset here? I.e. some comparison of outputs between old and new job?

akkomar:
Just noticed it fails on Databricks with
```
19/01/17 14:28:55 ERROR AddonRecommender$: Failed to write the model: Cannot run program "hdfs": error=2, No such file or directory
```
@acmiyaguchi I'll tag you again after fxing this.

akkomar:
Hey @Dexterp37 I'm wondering what's the purpose of https://github.com/mozilla/telemetry-batch-view/pull/518/files#diff-ebfbc196fc5b4a7f5a210a2363aa5ae1R281
How was the model fetched to "local" machine used? (What was "local" here? ATMO driver node or developer's machine?)

Dexterp37:
> Hey @Dexterp37 do you recommend any specific way for testing change to `clients_daily` dataset here? I.e. some comparison of outputs between old and new job?

Yes, I would look for the same addons being recommended: ideally same order, same addons. From there, I would check the differences with @mlopatka.

> Hey @Dexterp37 I'm wondering what's the purpose of https://github.com/mozilla/telemetry-batch-view/pull/518/files#diff-ebfbc196fc5b4a7f5a210a2363aa5ae1R281
> How was the model fetched to "local" machine used? (What was "local" here? ATMO driver node or developer's machine?)

Due to some permission issue, it was impossible back then to copy the file from the ATMO driver node to the desired S3 location. So it was going from HDFS, to "local" then to S3. Saving directly to S3 would be better.

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -223,6 +225,7 @@ object AddonRecommender {
       .setEvaluator(evaluator)
       .setEstimatorParamMaps(paramGrid)
       .setNumFolds(10)
+      .setParallelism(20)
```
acmiyaguchi: Why did you add this setting?

akkomar: We're training with cross-validation here, this setting allows for all models to be trained and evaluated at the same time (by default it's serial run).

crankycoder:
@acmiyaguchi can we rebase and get this merged in?

PR #521, opened 2019-02-21
Bug 1526278 - Fix PLUGINS_NOTIFICATION_SHOWN

mreid-moz:
Add a new field to count the zero bucket, since accumulating false
in this histogram has a particular meaning. From the description:

The number of times the click-to-activate notification was shown:
false: shown by in-content activation
true: shown by location bar activation


mreid-moz:
I think we're [ok for the other histograms](https://bugzilla.mozilla.org/show_bug.cgi?id=1526278#c1), this one is just weird that accumulating "false" has a special meaning.

PR #523, opened 2019-03-13
Bug 1534977: Fix Addon recommender StackOverflowError

sunahsuh:
Examining the spark logs for the particular node that failed in the last run, it looks like we're getting a StackOverflowError in one of the recursive training steps. This PR:

- Adds the ability to pass extra configs to the spark context and 
- Increases the jvm stack size on the spark drivers and executors to 8m


sunahsuh:
I looked for the instance ID for the internal IP ip-172-31-27-158 and found that node's spark logs :)

If you're on a mac I use this to grab the spark master log: https://gist.github.com/sunahsuh/c051e33545f35ef6fc630256af5d7be1

PR #526, opened 2019-03-30
Set decimal precision in clients_daily for bigquery compatibility

relud:
bigquery gets mad when loading decimal fields with any precision other than `38,9`


relud:
> is there a place we can document this

I was just going to fix datasets on an as-needed basis, since last i heard we are basically only copying select datasets to bq anyway.

PR #529, opened 2019-04-09
Bug 1536662: Handle cc tlds correctly for pioneer Online News Dwell Time 

sunahsuh:
Previously, the simple TLD + 1 domain logic in the dwell time dataset job meant that `co.uk` was being counted as a single domain. For country level TLDs, we're now using TLD + 2 (if applicable). This new implementation has the downside of adding specificity in some instances when it's potentially undesirable (e.g. the popular `.io` domain.) We can change the handling if that's a dealbreaker but this seemed like a good first stab.

(The second commit will have the changes you're probably interested in)

Additionally, adding tests revealed a significant bug where clients with only a single dwell time record were being dropped from the dataset, so that may also alter analysis.

A generated dataset is currently available in the pioneer analysis bucket under `/online_news_v2/dwell_time_cctlds/`


tdsmith:
Thanks for looking into this! Would it make sense to use the [public suffix list](https://publicsuffix.org/) for stemming?

Thanks for calling out the single-record bug.

tdsmith:
Oh, maybe not, that format is more expressive than I'd realized.

tdsmith:
Heh, Crossref has a partial Scala implementation of a "PSL parser" on Github but they just ignore wildcards and exceptions on the basis that they're rare, which is probably realistic. It doesn't look like it's FOSS licensed or I'd link it.

PR #530, opened 2019-04-11
Bug 1388025 - Remove crash aggregates code + docs

wlach:
It has been turned off in airflow, superseded by error_aggregates.


wlach:
> This looks good, but can you also add a note to the [`GRAVEYARD` doc](https://github.com/mozilla/telemetry-batch-view/blob/ade23bd69334bf89a857745cce5c6607ba0df8ad/GRAVEYARD.md)?

Done! I will do the same for #531.

PR #535, opened 2019-04-29
Bug 1543408 Add user_pref_browser_search_region to MainSummaryView

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1543408


jklukas:
This ran overnight, and the results look reasonable from a quick query.

PR #536, opened 2019-04-30
Bug 1536175 - Add "Headless" field

mreid-moz:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -798,6 +798,9 @@ object MainSummaryView extends BatchJobBase {
         (settings \ "intl" \ "requestedLocales").extractOpt[List[String]],
         (settings \ "intl" \ "systemLocales").extractOpt[List[String]],
 
+        // bug 1536175
+        (system \ "gfx" \ "Headless").extractOpt[Boolean],
```
sunahsuh: Wait, actually -- is the key capitalized? (if so, that differs from convention)

mreid-moz: Yup! It's documented [here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/environment.html) and I checked in `about:telemetry` locally.

PR #538, opened 2019-05-20
Bug 1552940 Add disk type to MainSummaryView

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1552940


jklukas:
Merging for now. Will post another PR if we find we need additional fields, since I haven't heard back yet.

PR #539, opened 2019-05-24
Bug 1552213 - Add UPDATE histograms to main_summary

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1552213


wlach:
Only comment is that it's better to have a descriptive commit message / pr name.

PR #541, opened 2019-06-10
Bug 1557525 - Add crash time to crash summary

wlach:


src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
```diff
@@ -98,6 +98,7 @@ case class Meta(
 
 case class Payload(
     crashDate: String,
+    crashTime: Option[String],
```
wlach: I am uncertain whether this should be a string or a date type of some kind?

wlach:
@akkomar does this seem sane to you? See https://bugzilla.mozilla.org/show_bug.cgi?id=1557525 for context.

I've successfully run this locally up to the point that it tries to upload to s3 (which failed because I guess sending data to s3 requires a different set of credentials than pulling it down)

src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
akkomar: I'd leave this as is because we don't have this field in the schema (so it's not validated).

src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
```diff
@@ -119,6 +120,7 @@ case class CrashSummary (
     build_version: String,
     build_id: String,
     channel: String,
+    crash_time: String,
```
akkomar: How about:
```suggestion
    crash_time: Option[String],
```
to keep this consistent with `client_id`?

src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
```diff
@@ -135,6 +137,7 @@ case class CrashSummary (
   def this(ping: CrashPing) = {
     this(
       client_id = ping.clientId,
+      crash_time = ping.payload.crashTime.getOrElse(""),
```
akkomar: Same as above:
```suggestion
      crash_time = ping.payload.crashTime,
```

src/test/resources/crash_ping_indented.json
```diff
@@ -46,6 +46,7 @@
       },
       "version": 1,
       "crashDate": "2017-01-05",
+      "crashTime": "2017-01-05T11:00:00.000Z",
```
akkomar: Since we have this I'd also add a check in the test: https://github.com/mozilla/telemetry-batch-view/blob/master/src/test/scala/com/mozilla/telemetry/views/CrashSummaryViewTest.scala#L93-L111

PR #544, opened 2019-06-29
Add Main Summary options to read ndjson instead of heka protobuf

relud:
this is currently reading from the stage decoded ndjson, because the prod ndjson doesn't exist yet.

I still need to test this in dataproc.


src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,110 @@
...
+    // TODO add docType to telemetry-decoded output path
+    val inputPath = s"$bucket/$prefix/${isoSubmissionDate.getOrElse("*")}/*/*.ndjson.gz"
+
+    new Dataset(inputPath, Map())
+  }
+
+  implicit def datasetToRDD(dataset: Dataset)(implicit sc: SparkContext): RDD[Option[JValue]] = {
+    dataset.records()
+  }
+}
```
relud: @mreid-moz you were right, creating a shim to read from ndjson instead of protobuf for this was p easy.

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,110 @@
...
+    ).map {
+      case Some(base64) => Try(Base64.getDecoder.decode(base64))
+      case _ => Failure
+    }.map {
+      case Success(payload) => parseOpt(payload.toString)
+      case _ => None
+    // provide doc \ "meta" to better match com.mozilla.telemetry.heka.Message.toJValue
+    }.map {
+      case Some(doc) =>
+        val submissionTimestamp = ZonedDateTime.parse(
```
akkomar: I think it wouldn't hurt to explicitly normalize this to UTC, either here or below where it's used.
This might help avoid potential headaches while testing locally.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -277,7 +281,9 @@ object MainSummaryView extends BatchJobBase {
         implicit val sc = spark.sparkContext
         val sqlContext = spark.sqlContext
         val hadoopConf = sc.hadoopConfiguration
-        hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
+        if (conf.outputFilesystem() == "s3") {
+          hadoopConf.set("fs.s3n.impl", "org.apache.hadoop.fs.s3native.NativeS3FileSystem")
```
akkomar: Don't we need similar config for `gs`?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -312,22 +318,56 @@ object MainSummaryView extends BatchJobBase {
...
+            // apply filterChannel as needed
+            val datasetWithFilteredChannel = filterChannel match {
+              case Some(expect) => dataset.where("app_update_channel") {
+                case channel => channel == expect
+              }
+              case _ => dataset
+            }
+            // apply filterVersion as needed
+            val datasetWithFilteredVersion = filterVersion match {
+              case Some(expect) => dataset.where("app_version") {
```
akkomar:
```suggestion
              case Some(expect) => datasetWithFilteredChannel.where("app_version") {
```
Typo?

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -312,22 +318,56 @@ object MainSummaryView extends BatchJobBase {
...
+                case v => v == expect
+              }
+              case _ => datasetWithFilteredChannel
+            }
+            // convert to rdd
+            datasetWithFilteredVersion.records()
+          case Some("heka") =>
+            // return
```
akkomar: Nit about comments here and above - I'm not sure they bring much value since variable and method naming is pretty good here.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -339,13 +379,13 @@ object MainSummaryView extends BatchJobBase {
         //  - "error" (the default) causes the job to fail after any data is
         //    loaded, so we can't do single day incremental updates.
         //  - "ignore" causes new data not to be saved.
-        // So we manually add the "submission_date_s3" parameter to the s3path.
-        val s3prefix = s"${filterDocType}_summary/$schemaVersion/submission_date_s3=$submissionDate"
-        val s3path = s"s3://${conf.outputBucket()}/$s3prefix"
+        // So we manually add the "submission_date" parameter to the outputPath.
+        val keyPrefix = s"${filterDocType}_summary/$schemaVersion/submission_date=$submissionDate"
```
akkomar: Omitting `_s3` would break AWS runs.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: I think this only applies to the `s3n://` filesystem, which we don't use? I just didn't want to mess with the existing thing. I'm pretty sure dataproc already has `gs://` correctly configured. I'm sure I will find out in testing.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: good point, forgot to put this back when i was refactoring a change

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: will remove, this is an artifact of before I refactored the variable and method naming to be easy to read >_>

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
relud: yeah, copypasta

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
akkomar: You're right, it probably has.

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
relud: this parses `submission_timestamp`, which is generated by the edge server, and [must end with timezone `Z`](https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#edge-server-pubsub-message-schema), so it's already UTC.

relud: I will add a comment pointing this out

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,112 @@
...
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.json4s.{DefaultFormats, Formats, JValue}
+import org.json4s.JsonAST.{JObject, JString, JInt}
+import org.json4s.jackson.JsonMethods.parseOpt
+
+import scala.util.{Failure, Success, Try}
+
+class Dataset private (inputPath: String, clauses: Map[String, PartialFunction[String, Boolean]]) {
```
sunahsuh: 🙌 
Should this go in https://github.com/mozilla/moztelemetry though? I think we have a few places where we'll want to use this

relud: maybe, depends where else we need to use it.

I don't think we should allow use of the ndjson beyond temporary migration measures which will eventually be updated to read from bigquery or pubsub.

I feel like putting this in moztelemetry implies it's less half-baked than it is. I really only implemented the bits that support main summary, in the easiest way possible, rather than implementing this as a solid drop in place replacement for `heka.Dataset`.

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
acmiyaguchi: I ran into an issue with gcs with pyspark that required setting `spark.conf.set("fs.gs.implicit.dir.repair.enable", False)`. Access to decoded data like `moz-fx-data-stage-data`  is read-only, but the gcs connector want to run an implicit directory repair with write access. If the job runs into a similar issue, this should work.

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -2325,4 +2331,125 @@ class MainSummaryViewTest extends FlatSpec with Matchers with DataFrameSuiteBase
 
     compare(message, expected)
   }
+
+  private val tempDir = createTempDir().toString
+
+  "ndjson.Dataset source" should "succeed" in {
```
akkomar: Can we be more precise here and describe what's behind "succeed"? :)

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -2325,4 +2331,125 @@ class MainSummaryViewTest extends FlatSpec with Matchers with DataFrameSuiteBase
...
+         |}""".replaceAll("\\s+\\|", ""),
+      s"""{"document_id":"$docId2","client_id":"$clientId1","sample_id":42,"channel":"release",
+         |"normalized_channel":"release","normalized_os_version":"10.0","country":"US","city":
+         |"Hermiston","geo_subdivision1":"OR","geo_subdivision2":"Umatilla",
+         |"submission_date":"20190701","timestamp":1561939200000000000,
+         |"client_submission_date":"Mon, 1 Jul 2019 00:00:00 GMT","client_clock_skew":0
+         |}""".replaceAll("\\s+\\|", ""),
+      s"""{"document_id":"$docId3","submission_date":"20190701","timestamp":1561939200000000000}"""
+    )
+    assert(result == expect)
```
akkomar:
```suggestion
    result should contain theSameElementsAs expect
```
this will be more robust.

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+}
+
+object Dataset {
+  val DATE_NO_DASH: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd")
+  private var PROD_BUCKET: String = "gs://moz-fx-data-prod-data"
+  private var STAGE_BUCKET: String = "gs://moz-fx-data-stage-data"
+
+  private def useTestingPath(path: String): Unit = {
```
akkomar: I suggest to:
* define `PROD_BUCKET` and `STAGE_BUCKET` as `val`s
* remove `useTestingPath` method
* add a way to override buckets to `apply` method, for example like:
  `def apply(dataset: String, submissionDate: Option[String] = None, prod_bucket_override: Option[String] = None, stage-bucket_override: Option[String] = None): Dataset`
* use these new parameters in test


This way we'll:
* get rid of mutability here
* avoid introduction of private method used only in test and called via reflection

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -2325,4 +2331,125 @@ class MainSummaryViewTest extends FlatSpec with Matchers with DataFrameSuiteBase
...
+        "attributeMap" -> baseAttributes,
+        "payload" -> (basePayload + ("document_id" -> docId3))
+      )
+    ).map(
+      msg => asJson(msg + ("payload" -> base64(asJson(msg("payload")))))
+    ).zipWithIndex.foreach {
+      case (line, index) => writeTextFile(s"$prefix/$index.ndjson.gz", s"$line\n", compression)
+    }
+
+    ndjson.Dataset invokePrivate PrivateMethod[ndjson.Dataset]('useTestingPath)(datasetTestingPath)
```
akkomar: See a comment about private method in `Dataset`.

src/main/scala/com/mozilla/telemetry/views/BatchJobBase.scala
```diff
@@ -64,6 +64,6 @@ object BatchJobBase {
 
 trait DatabricksSupport {
   protected def shouldStopContextAtEnd(spark: SparkSession): Boolean = {
-    !spark.conf.get("spark.home", "").startsWith("/databricks")
+    !spark.conf.get("spark.home", "").startsWith("/databricks") && spark.conf.get("mozilla.doNotStopSparkContext", "").isEmpty
```
akkomar: `"mozilla.doNotStopSparkContext"` is not related to Databricks (this is `DatabricksSupport` trait).

Do we really need this? What if the core logic of `MainSummaryView` was extracted to a method and we called it in the `MainSummaryViewTest`, like in other tests? Then we wouldn't need this flag.

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
relud: i did not know how to do that, TYSM

src/main/scala/com/mozilla/telemetry/views/BatchJobBase.scala
relud: I moved this from the `DatabricksSupport` trait, into the option `--disable-stop-context-at-end`

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -268,20 +270,31 @@ object MainSummaryView extends BatchJobBase {
       default=Some("fixed"))
     val inputPartitionMultiplier = opt[Int]("input-partition-multiplier", descr="Partition multiplier for aligned read-mode", default=Some(4))
     val schemaReportLocation = opt[String]("schema-report-location", descr="Write schema.treeString to this file")
+    val inputSource = choice(Seq("heka", "ndjson"), name="input-source",
+      descr="Read raw pings from heka protobuf in S3 (heka) or newline delimited json in Google Cloud Storage (ndjson)", default=Some("heka"))
+    val inputBucket = opt[String]("input-bucket", descr="Bucket override for --input-source=ndjson", default=Some(Dataset.BUCKET))
+    val disableStopContextAtEnd = opt[Boolean]("disable-stop-context-at-end", descr = "Flag to leave spark context running", required = false)
```
relud: I moved the config option from `trait DatabricksSupport` to here.

> Do we really need this?

I think calling `main` directly is the most accurate way to test our jobs, and that better testing constitutes something we really need.

>  What if the core logic of MainSummaryView was extracted to a method and we called it in the MainSummaryViewTest, like in other tests? Then we wouldn't need this flag.

We have logic (here and elsewhere) that does not fit between `getOrCreateSparkSession` and `spark.stop`, which I think we really should be testing.

Even if I could test all of the changes in this PR with that approach, I think this is a flag worth having so that non-databricks use cases can leave the spark context running, because it has so many side-effects.

akkomar:
I'm holding on with review until we have the "colon" problem fixed (@relud  ping me if I'm blocking you)

relud:
@akkomar https://github.com/mozilla/gcp-ingestion/pull/699 fixes the colon issue, I think we're good to move forward with review

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,138 @@
...
+    // add dashes to submissionDate for backwards compatibility
+    val isoSubmissionDate = submissionDate.map(LocalDate.parse(_, Dataset.DATE_NO_DASH).format(ISO_DATE))
+
+    // look up relevant sink output
+    val prefix = dataset match {
+      case "telemetry" => s"$bucket/telemetry-decoded_gcs-sink/output"
+    }
+
+    // TODO move document_namespace and document_type to path
+    // TODO use hive partition format for filterable path elements and allow submissionDate as a clause
```
akkomar: We could add an issue link here if we have one for this.

src/test/scala/com/mozilla/telemetry/views/MainSummaryViewTest.scala
```diff
@@ -2325,4 +2328,124 @@ class MainSummaryViewTest extends FlatSpec with Matchers with DataFrameSuiteBase
 
     compare(message, expected)
   }
+
+  private val tempDir = createTempDir().toString
+
+  "ndjson source" should "supply doc.meta" in {
```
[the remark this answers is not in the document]
relud:
```suggestion
  "main method" should "correctly process ndjson-serialized input" in {
```

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
relud: I'm actually going to remove these `TODO`s because we decided they aren't presently feasible.
```suggestion
```

src/main/scala/com/mozilla/telemetry/ndjson/Dataset.scala
```diff
@@ -0,0 +1,138 @@
...
+  def records(limit: Option[Int] = None)(implicit spark: SparkSession): RDD[Option[JValue]] = {
+    // add dashes to submissionDate for backwards compatibility
+    val isoSubmissionDate = submissionDate.map(LocalDate.parse(_, Dataset.DATE_NO_DASH).format(ISO_DATE))
+
+    // look up relevant sink output
+    val prefix = dataset match {
+      case "telemetry" => s"$bucket/telemetry-decoded_gcs-sink/output"
+    }
+
+    // TODO move document_namespace and document_type to path
```
relud:
```suggestion
```

PR #546, opened 2019-07-04
Optimize crash summary view

wlach:
* Integrate accumulation of processed/discarded counts into map
* Use cached accumulator values to calculate total (as opposed to
  getting the size all over again)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
+++ b/src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
@@ -229,14 +229,15 @@ object CrashSummaryView extends BatchJobBase {
       val processedPings = spark.sparkContext.longAccumulator("processedPings")
       val discardedPings = spark.sparkContext.longAccumulator("discardedPings")
       val crashPings = messages.records()
-        .map(x => this.transformPayload(x.fieldsAsMap, x.payload))
-      crashPings.foreach(x => {
-        if (!x.isDefined) {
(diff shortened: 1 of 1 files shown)


wlach:
When I was profiling the crash summary job, I noticed some dumb things:

image

1. Spark spent thirty minutes on a foreach loop to count the number of processed and discarded pings. I don't see why we can just incorporate this into the map statement.
2. Spark spent thirty minutes at the end doing (another) count of the total number of pings, even though this was just the sum of what was calculated in (1)

It's possible (1) won't be a giant improvement in practice (if later statements use the cached dataframe which was used to generate the counts), but (2) seems like a no-brainer.

I think I've forgotten much of the scala I've learned a couple years ago, so feel free to let me know if I'm doing something wrong.

src/main/scala/com/mozilla/telemetry/views/CrashSummaryView.scala
```diff
@@ -229,14 +229,15 @@ object CrashSummaryView extends BatchJobBase {
...
-      crashPings.foreach(x => {
-        if (!x.isDefined) {
-          discardedPings.add(1)
-        } else {
-          processedPings.add(1)
-        }
-      })
+        .map(x => {
+          val payload = this.transformPayload(x.fieldsAsMap, x.payload);
+          if (!payload.isDefined) {
```
akkomar:
```suggestion
          if (payload.isEmpty) {
```
^would be more idiomatic.

PR #547, opened 2019-07-23
Bug 1567339 - Add parameters to attribution in environment data

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1567339)

This adds `experiment` and `variation` stub attribution fields to main_summary. Unfortunately, attribution was originally added as a nested field, so it can't be updated to accommodate these new fields. They've been added as `attribution_experiment` and `attribution_variation` at the top level instead.

The BigQuery tables are a better choice in this case (see this [schema diff](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/03466d5532a89c107889ebf3badf333d7da89534)).


acmiyaguchi:
Doubling up the columns might be confusing since one set of columns will have more data than the other. Maybe a view would solve this problem downstream, if there's a demand for consistent column naming w.r.t. attribution fields in main_summary.

PR #550, opened 2019-09-10
Support reading from payload_bytes_decoded in MainSummary

relud:


src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
```diff
@@ -407,15 +408,15 @@ object MainSummaryView extends BatchJobBase {
...
+            outputBucket, keyPrefix)
         }
 
         if (shouldStopContextAtEnd(spark) && !conf.disableStopContextAtEnd()) { spark.stop() }
       }
     } catch {
       // Delete incomplete data
       case e@TooManyRecordsIgnoredException(_, bucket, prefix) =>
-        deletePrefix(bucket, prefix)
+        hadoopDelete(s"$bucket/$prefix", recursive = true)
```
akkomar: This won't work on S3 in `HEKA` mode.

src/main/scala/com/mozilla/telemetry/utils/package.scala
```diff
@@ -179,4 +179,13 @@ package object utils{
     val path = new Path(pathString)
     FileSystem.get(path.toUri, new Configuration()).exists(path)
   }
+
+  /** Detect existence of a directory or file using hadoop
```
akkomar: Docstring doesn't match method signature, did you mean:
```suggestion
  /** Delete file or directory in Hadoop filesystems
```

src/main/scala/com/mozilla/telemetry/views/MainSummaryView.scala
akkomar: Oh yes, my bad and you are right. This will work because S3Filesystem is configured on AWS clusters.

src/test/scala/com/mozilla/telemetry/spark/FakeBigQueryDataSource.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.sources.{DataSourceRegister, RelationProvider}
+
+/** Register a custom Spark Data Source for faking BigQuery Storage API with Spark temporary views.
+  *
+  * <p>Requires src/test/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
+  * to contain com.mozilla.telemetry.spark.FakeBigQueryRelation
+  */
+class FakeBigQueryDataSource extends DataSourceRegister with RelationProvider {
```
akkomar: Clever!

relud:
> have you managed to run this on a full day

it still hasn't succeeded yet. most recent run didn't *fail*, but I killed it after letting it run for 18 hours.

relud:
I couldn't get this to work via the storage API, but exporting a day in avro format takes less than 2 minutes, and is ~2TB (with a project-wide limit of exporting 10TB), and reading that appears to be working much better. I plan to handle the export and cleanup of gcs objects as a separate task in airflow.

PR #551, opened 2019-09-19
Add extractors for Sync engine steps

linacambridge:
The `steps` field is a new field added to the Sync ping in https://bugzilla.mozilla.org/show_bug.cgi?id=1552621. Each step has a name, millisecond time, and optional named counts, using the same format as validation problems. We use these in new bookmark sync, to report timings and counts for each stage.

This extracts the steps into the summary view (but not the flat engines view), so that we can query them.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
+++ b/src/main/scala/com/mozilla/telemetry/utils/SyncPingConversion.scala
@@ -48,10 +48,16 @@ object SyncPingConversion {
     StructField("os", StringType, nullable = false)
   ))
 
-  // Data about a single validation problem found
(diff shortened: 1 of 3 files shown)


linacambridge:
@acmiyaguchi r?

acmiyaguchi:
@linacambridge As for the flattened view, it makes sense to take advantage of the ping tables and [BigQuery destination tables](https://cloud.google.com/bigquery/docs/tables#creating_a_table_from_a_query_result). It may be a good time to rewrite the query for [mozilla/bigquery-etl](https://github.com/mozilla/bigquery-etl/) to do something similar to `python_mozetl` for `sync_summary`.

PR #566, opened 2021-02-05
TAAR colab job GCS migration

eu9ene:
- switched from S3 to GCS
- added compression


src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
```diff
@@ -204,9 +205,12 @@ object AddonRecommender extends DatabricksSupport {
         logger.info(s"Checkpoint directory is not specified")
     }
 
+    val compressionCodec = new BZip2Codec()
+    compressionCodec.setConf(new Configuration())
+
     import spark.implicits._
     implicit val formats = DefaultFormats
-    val json_str = hadoopRead(s"$privateBucket/telemetry-ml/addon_recommender/only_guids_top_200.json")
+    val json_str = hadoopRead(s"$privateBucket/addon_recommender/only_guids_top_200.json.bz2", Some(compressionCodec))
```
haroldwoo, line 213: nit: other variables are named with camelCase instead of underscores, but it looks like you didnt name this var anwyay

haroldwoo (approved):

src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala
eu9ene: renamed

